PR #12, opened 2020-02-10
Basic Form components

suzubara:
This PR adds a number of basic form components, using the markup and styles [demonstrated by USWDS](https://designsystem.digital.gov/components/form-controls/).

This includes form inputs:
- Checkbox
- Dropdown (Select)
- Radio
- TextInput
- Textarea

as well as other UI components related to forms:
- ErrorMessage
- Fieldset
- Form
- FormGroup
- Label

Because the form inputs are _[uncontrolled](https://reactjs.org/docs/uncontrolled-components.html)_ they are handling little more than rendering the input HTML with USWDS classes. They also all accept their native HTML attributes as props, and accept an `inputRef` prop that is returned the ref to the given input element. This should allow use of these components in applications, agnostic of any other form libraries being used. So far, they have only been tested for usage with [React Hook Form](https://react-hook-form.com/), but as others start using this library it would be good to test how they work with other common React form libraries and make any modifications that might be needed.

I recommend reviewing by either looking at the components in Storybook locally, or try to include on a React project you're working on.

image

image

image


src/components/forms/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  const classes = classnames(
+    'usa-input',
+    {
+      'usa-input--error': error,
+      'usa-input--success': success,
+      'usa-input--small': small,
+      'usa-input--medium': medium,
+    },
+    className
```
gidjin: similar to the dropdown there doesn't appear to be a visual difference between disabled and enabled. I wanted to double check if that was intended or not. Also same for read only, but that may not need a visual difference either, and I don't see any details on this on the uswds site.

src/components/forms/Textarea/Textarea.tsx
```diff
@@ -0,0 +1,55 @@
...
+    name,
+    className,
+    error,
+    success,
+    children,
+    inputRef,
+    ...inputProps
+  } = props
+
+  const classes = classnames(
```
gidjin: same question here on disabled and readonly as for text input.

src/components/forms/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,37 @@
...
+    | null
+    | undefined
+}
+
+export const Dropdown = (
+  props: DropdownProps & React.SelectHTMLAttributes<HTMLSelectElement>
+): React.ReactElement => {
+  const { id, name, className, inputRef, children, ...inputProps } = props
+
+  const classes = classnames('usa-select', className)
```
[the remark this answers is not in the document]
suzubara: yes it should :( however in this PR I was just using the pre-existing USWDS classes/styles and they don't seem to have any style for disabled dropdown. It is probably a good thing to add but I defer to @joshmfranklin on this

src/components/forms/TextInput/TextInput.tsx
suzubara: same as above

src/components/forms/Textarea/Textarea.tsx
suzubara: same as above

src/components/forms/TextInput/TextInput.stories.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+export const disabled = (): React.ReactElement => (
+  <TextInput id="input-disabled" name="input-disabled" type="text" disabled />
+)
+
+export const readonly = (): React.ReactElement => (
+  <TextInput id="input-readonly" name="input-readonly" type="text" readOnly />
+)
+
+/*
```
sojeri: use or remove commented out code?

src/components/forms/Dropdown/Dropdown.tsx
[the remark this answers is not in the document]
joshmfranklin: Generally I think that disabled form elements aren't the greatest design decision. If an element is disabled should it even be on the page? could it be rendered as static text instead? If we do need to include a disabled form element then I think we need to visually make it different or add a cursor style of `not allowed` like @gidjin noted. I think we could come back to this once we know of a situation in which it's needed.

suzubara:
@sojeri I think having a built-in example app would be a great idea so it's easier to test out changes within this library, I added it as an item to the roadmap!

PR #13, opened 2020-02-17
Adds dependabot for dependency management

esacteksab:


dynamike:
@suzubara This PR will add dependabot to the repo, which submit PRs with updates the various JS modules. Right now it will run weekly.

.dependabot/config.yml
```diff
@@ -0,0 +1,9 @@
+version: 1
+update_configs:
+  # Keep go modules up to date, batching pull requests weekly
```
suzubara: seems like this should be `js modules` :)

dynamike: doh :)

esacteksab: Ooops! My copy and paste is showing!

PR #21, opened 2020-03-11
Add support for additional custom class names for Button

gidjin:
In our use of `Button` it would be helpful to add custom class names so we could make the custom theming we need for MilMove. This PR adds support for that.


src/components/Button/Button.tsx
```diff
@@ -13,6 +13,7 @@ interface ButtonProps {
   big?: boolean
   unstyled?: boolean
   onClick?: (event: React.MouseEvent<HTMLButtonElement>) => void
+  className?: string
```
suzubara: so this is just a suggestion but I think it will be useful for us to adopt throughout this lib... I recently discovered you can do this when defining prop types for components that implement HTML elements:
```
export const Button = (props: ButtonProps & React.HTMLAttributes<HTMLButtonElement>): React.ReactElement => {...
```

that will automatically extend any native HTML or React props that a `<button>` element should be able to receive, and then we don't have to explicitly include props such as `onClick` and `className` in our `ButtonProps`! super useful :D there is a full list of React types here https://www.saltycrane.com/cheat-sheets/typescript/react/latest/

gidjin: This sounds like a great idea, I'll update this PR accordingly.

PR #32, opened 2020-03-19
Release candidate: 1.1.0

suzubara:
## [1.1.0] - 2020-03-19

- Add initial set of USWDS form components (Checkbox, Dropdown, ErrorMessage, Fieldset, Form, FormGroup, Label, Radio, TextInput, Textarea)
- Add native HTML attributes to existing Tag & Button components in order to support custom CSS classes
- Added new Modal component (designed by Truss) and related Overlay and ModalContainer components
- Configured use of SASS/SCSS and SASS/SCSS modules so that we can use USWDS SCSS
- Configure Dependabot for automated dependency updates


CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @suzubara
```
suzubara: Anyone who wants to add themselves here should feel free to do so!

PR #48, opened 2020-04-05
resolve 2/3 security advisories

sojeri:
This doesn't fix the `kind-of` issue but does fix the other two alerts attached to this repo. Kind of is a lot messier so I think we can hold out a few more days to see if big maintainers like storybook and FB will fix their deps. It cleans up yarn audit output *a ton*.

### minimist

security advisory: [Prototype pollution in minimist](https://github.com/advisories/GHSA-vh95-rmgr-6w4m)

came from gonzales-pe pkg via postcss-sass
- fixed in tonyganch/gonzales-pe#304 (=> `4.3.0`)
- incorporated in AleshaOleg/postcss-sass#143 (=> `0.4.4`)

I could have fixed by patching yarn.lock alone, but used the force upgrade method for explictness here. See my work in b8e6805d, c0f0a829, and 791b1e27.

### acorn

security advisory: [Regular Expression Denial of Service in Acorn](https://github.com/advisories/GHSA-6chw-6frg-f759)

TBH, came from bad yarn logic. This is a long-time bug in yarn which happens from time to time: our dependency tree had no reason to have both 6.4.0 and 6.4.1, but yarn isn't great at cleaning these dupes. So I manually patched yarn.lock in fac6225b.

## reviewer notes

opened this based on yarn audit alone, so please check to ensure I didn't break the build? :)

### yarn audit output-- before my changes (394 vulnerabilities)

Screen Shot 2020-04-05 at 6 31 40 PM

### yarn audit output-- after my changes (1 vulnerability)

this last vulnerability isn't from kind of. a closed issue on storybook claims this is already fixed, but upgrading to 5.3.18 did *not* resolve this. I am still bored so may open a PR to upgrade marksy & then can follow up to upgrade storybook if nobody else catches it.

Screen Shot 2020-04-05 at 6 32 50 PM

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -5975,6 +5975,13 @@ gonzales-pe@^4.2.4:
   dependencies:
     minimist "1.1.x"
 
+gonzales-pe@^4.3.0:
+  version "4.3.0"
+  resolved "https://registry.yarnpkg.com/gonzales-pe/-/gonzales-pe-4.3.0.tgz#fe9dec5f3c557eead09ff868c65826be54d067b3"
+  integrity sha512-otgSPpUmdWJ43VXyiNgEYE4luzHCL2pz4wQ0OnDluC6Eg4Ko3Vexy/SrSynglw/eR+OhkzmqFCZa/OFa/RgAOQ==
+  dependencies:
+    minimist "^1.2.5"
+
 good-listener@^1.2.2:
   version "1.2.2"
   resolved "https://registry.yarnpkg.com/good-listener/-/good-listener-1.2.2.tgz#d53b30cdf9313dffb7dc9a0d477096aa6d145c50"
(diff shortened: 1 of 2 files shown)


suzubara:
I don't feel comfortable committing manual fixes directly to the yarn.lock file -- IMO they will be too easily lost and it's not clear enough what versions we're requiring and why. I found this yarn feature that sounds like a good enough alternative: https://classic.yarnpkg.com/en/docs/selective-version-resolutions/

sojeri:
@suzubara TIL! I will add those to `package.json` 👍

yarn.lock
```diff
@@ -11179,7 +11164,7 @@ stylelint@^13.2.1:
     lodash "^4.17.15"
     log-symbols "^3.0.0"
     mathml-tag-names "^2.1.3"
-    meow "^6.0.1"
+    meow "^6.1.0"
```
sojeri, line 11167: meow package image: a cat spitting rainbow droplets

( https://github.com/sindresorhus/meow )

PR #105, opened 2020-04-24
Add title component

eamahanna:
## PR Description
This PR adds the Title component. This component may be used through out an application, and is included in the USWDS Header Component: https://designsystem.digital.gov/components/header/

Screen Shot 2020-04-24 at 2 01 30 PM

## For Reviewers
Should this component be composed in an alternate manner to make it more reuse-able?  

To test this component execute the following commands in the react-uswds repo:
```
yarn install
yarn storybook
```
Navigate to the Title component.

### DIFF
--- a/src/components/Title/Title.test.tsx
+++ b/src/components/Title/Title.test.tsx
@@ -1,4 +1,4 @@
-import React, { Children } from 'react'
+import React from 'react'
 import { render } from '@testing-library/react'
 
 import { Title } from './Title'
--- a/src/components/Title/Title.test.tsx
+++ b/src/components/Title/Title.test.tsx
@@ -1,13 +1,11 @@
-import React from 'react'
+import React, { Children } from 'react'
 import { render } from '@testing-library/react'
 
 import { Title } from './Title'
 
-const testTitle = 'Test Title'
-
 describe('Title component', () => {
   it('renders without errors', () => {
-    const { queryByText } = render(<Title title={testTitle}></Title>)
+    const { queryByText } = render(<Title>Test Title</Title>)
     expect(queryByText('Test Title')).toBeInTheDocument()
   })
 })
--- a/src/components/Title/Title.test.tsx
+++ b/src/components/Title/Title.test.tsx
@@ -1,0 +1,16 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+
+import { Title } from './Title'
+
+const testTitle = 'Test Title'
+const testLink = '#testLinkOne'
+
+describe('Title component', () => {
+  it('renders without errors', () => {
+    const { queryByText } = render(
+      <Title title={testTitle} link={testLink}></Title>
+    )
+    expect(queryByText('Test Title')).toBeInTheDocument()
+  })
+})
--- a/src/components/Title/Title.test.tsx
+++ b/src/components/Title/Title.test.tsx
@@ -4,13 +4,10 @@ import { render } from '@testing-library/react'
 import { Title } from './Title'
 
 const testTitle = 'Test Title'
-const testLink = '#testLinkOne'
 
 describe('Title component', () => {
   it('renders without errors', () => {
-    const { queryByText } = render(
-      <Title title={testTitle} link={testLink}></Title>
-    )
+    const { queryByText } = render(<Title title={testTitle}></Title>)
     expect(queryByText('Test Title')).toBeInTheDocument()
   })
 })
--- a/src/components/Title/Title.tsx
+++ b/src/components/Title/Title.tsx
@@ -3,13 +3,18 @@ import classnames from 'classnames'
 
 interface TitleProps {
   children: React.ReactNode
+  className?: string
 }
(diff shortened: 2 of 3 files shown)


src/components/Title/Title.tsx
```diff
@@ -0,0 +1,21 @@
...
+  link?: string
+}
+
+export const Title = ({ title, link }: TitleProps): React.ReactElement => {
+  const classes = classnames('usa-logo')
+
+  return (
+    <div className={classes}>
+      <em className="usa-logo__text">
+        <a href={link}>{title}</a>
```
haworku, line 15: A couple things. 
- 🤔 This very much feels like it should be a `<h1>`.  It is literally our site title- for accessibility reasons too.  I wonder why `em` is being used by the uswds design team...
-  As written, `link` prop could be undefined - we should probably not wrap the text in an anchor tag in the first place for those cases.

trussworks-infra:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against fc8312aec7723757f078f5612c4a538bb25fc048
</p>

src/components/Title/Title.tsx
```diff
@@ -0,0 +1,18 @@
...
+interface TitleProps {
+  title: React.ReactNode
+}
+
+export const Title = ({ title }: TitleProps): React.ReactElement => {
+  const classes = classnames('usa-logo')
+
+  return (
+    <div className={classes}>
+      <h1 className="usa-logo__text">{title}</h1>
```
haworku, line 13:
```suggestion
      link?
      <h1 className="usa-logo__text">
        <a href={link}>{title}</a>
      </h1>
      : <h1 className="usa-logo__text">{title}</h1>
```
Oops! My comment was unclear.  What you had before was good, just conditionally display the link. Only include it when it exists.  If it doesn't exist,  leave it out .

suzubara: I think we should avoid rendering `<a>` in general since we don't know what kind of routing a user is using. Also if we _do_ want to render `<a>` we need to allow for element attributes to be passed in

eamahanna:
> Sorry it took me awhile to get to this 🤦‍♀️ I have some suggestions:
> 
> * Should we call this component `HeaderTitle` or maybe `HeaderLogo`? I'm thinking for clarity purposes, because this component does not exist outside of the context of the header on the USWDS site, and so it may not work as expected if it's not rendered inside of the header. Or it could be located inside `components/Header`.
> * Here are my 2 cents on `h1` vs `em`: by default we should render `<em class="usa-logo__text">` in this component since that follows the examples from https://designsystem.digital.gov/components/header/ . However the name of that CSS class implies there could be another type of logo, such as an image. So we might want to think about how to allow other kinds of titles/logos to be passed in (such as an h1 or an image) in the future. Call me crazy but we could go as far as creating a `Logo` component which renders the `<div class="usa-logo">` for positioning, and a `TextLogo` component which renders `<em class="usa-logo__text">` for default text styling.
> * Since the purpose of this component is to render its contents, IMO the title should be passed in as `children` instead of a prop. So rendering this component would look like
> 
> ```
> <Title>Test Title</Title>
> ```

When I merge this branch into my main header branch, I will move the title into the header folder.

src/components/Title/Title.test.tsx
```diff
@@ -0,0 +1,11 @@
+import React, { Children } from 'react'
```
suzubara, line 1: `{ Children }` looks like an extra import, maybe it was auto added by accident?

src/components/Title/Title.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import classnames from 'classnames'
+
+interface TitleProps {
+  children: React.ReactNode
+}
+
+export const Title = ({ children }: TitleProps): React.ReactElement => {
```
suzubara, line 8: for the sake of flexibility, I think we should try to extend the React HTML attribute props wherever possible, and that way someone can pass in standard React props such as `className` & `children`, as well as native HTML attributes (like aria attributes) without us having to explicitly handle them. I need to add documentation to this repo about taking that approach because I don't think it is well known, but it would look something like this:

```
export const Title = ({ className, children, ...props } : TitleProps & React.HTMLAttributes<HTMLDivElement>): React.ReactElement => {

...

return (
  <div className={classes} {...props}>
  ...
```

src/components/Title/Title.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+import classnames from 'classnames'
+
+interface TitleProps {
+  className?: string
```
suzubara, line 5: this is _extremely_ nitpicky but `className` is part of `React.HTMLAttributes` and you can therefore omit it here

Issue #78, opened 2020-04-15
USWDS Component: Footer


suzubara:
Add Footer component: https://designsystem.digital.gov/components/footer/

Should support all options: Big, Medium, Slim
Should support mobile view (seen particularly in Big Footer)

Render content:
- Return to top link
- Navigation tree (might be sub components, and/or children)
- Sign up form or similar form
- Address, logo, and social media links(sub components rendered as children)

PR #107, opened 2020-04-27
Add footer, slim footer story example

haworku:
# `<Footer />`

The footer displays primary content (upper section) and secondary content (lower section).  This PR adds the initial component and a story example for displaying the slim footer. 

PR (1 of 3) for #78


trussworks-infra:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against fefd91eaba3b115f86b680113c2386a4765f0d86
</p>

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,44 @@
...
+   * Content in upper footer section, e.g. navigation, searchbar, signup form
+   */
+  primary: React.ReactNode
+  /**
+   * Content in lower footer section, e.g. contact information
+   */
+  secondary: React.ReactNode
+}
+
+// TODO: Add in "Return to Top" handling
```
haworku, line 19: `returnToTop` could be an option prop -- thoughts?  Or should this component just render `children`?  It seems like it is supposed to also encapsulate styles so that is why I made `primary` and `secondary` in the first place.

src/components/Footer/Footer.stories.tsx
```diff
@@ -0,0 +1,104 @@
...
+const Logo = ({ big, medium, slim }: SizeProps): React.ReactElement => {
+  const containerClasses = classnames('usa-footer__logo grid-row', {
+    'mobile-lg:grid-col-6 mobile-lg:grid-gap-2': big || medium,
+    'grid-gap-2': slim,
+  })
+
+  return (
+    <div className={containerClasses}>
+      <div className="grid-col-auto">
+        <img className="usa-footer__logo-img" src="#" alt="img alt text" />
```
haworku, line 80: 🤔 Should we add a mock logo image file to our library? It might be nicer for use in stories.

src/components/Footer/Footer.stories.tsx
```diff
@@ -0,0 +1,104 @@
...
+const MockNav = ({ medium, slim }: SizeProps): React.ReactElement => {
+  const itemClasses = classnames(
+    'desktop:grid-col-auto usa-footer__primary-content',
+    {
+      'mobile-lg:grid-col-4': medium,
+      'mobile-lg:grid-col-6': slim,
+    }
+  )
+
+  const items = Array(4).fill({
```
suzubara, line 51: 🤯 TIL about `Array.fill`

src/components/Footer/Footer.stories.tsx
suzubara: Yeah! we could use the empty circle used on USWDS for something easy? https://designsystem.digital.gov/components/footer/

as long as it's only imported to the `.stories.tsx` file it won't be included in the packaged library.

src/components/Footer/Footer.tsx
suzubara: I agree that `primary` and `secondary` is the way to go, since all of the USWDS footers divide up the content that way pretty explicitly. What if we had an optional `onReturnToTop` fn prop that, if it exists renders the Return to top link and hooks up the event handler?

haworku:
@suzubara Yes about refactoring out components. Added `Address` and `Logo` for now so everything is together. 🤔 What is `ContactInfo`- the styled items in `Address`?

FYI planning to add `FooterNav`, `FooterNavExtended` and `SocialLinks` in later PRs when the medium and big story examples go up.

src/components/Footer/Footer.stories.tsx
haworku: Just pushed a couple more commits to round out this PR. Ended up adjusting the `npm run storybook` script to get the static logo image loading properly.  Followed [this ](https://storybook.js.org/docs/configurations/serving-static-files/#2-via-a-directory)approach.

src/components/Footer/Footer.tsx
haworku: Made `returnToTop` a component when I realized that if I only made the handler a prop I am hardcoding English, and that specific phrase, into our repo. I'll make an issue about lang and i18n so we can discuss.

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,56 @@
...
+    {
+      'usa-footer--big': big,
+      'usa-footer--medium': medium,
+      'usa-footer--slim': slim,
+    },
+    footerAttributes.className
+  )
+
+  return (
+    <footer className={classes}>
```
suzubara, line 47: do we also want to include `{...footerAttributes}` here to pass along anything else?

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,56 @@
...
+      {returnToTop && returnToTop}
+      <div className="usa-footer__primary-section">{primary}</div>
+
+      <div className="usa-footer__secondary-section">
+        <div className="grid-container">{secondary}</div>
```
suzubara, line 52: why not leave `<div className="grid-container">` to be part of the `secondary` prop as well?

src/components/Footer/Logo/Logo.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  return (
+    <div className={containerClasses} data-testid="footerLogo">
+      {heading ? (
+        <>
+          <div className="grid-col-auto">{image}</div>
+          <div className="grid-col-auto">{heading}</div>
+        </>
+      ) : (
+        <>{image}</>
```
suzubara, line 37: I think since the container is a `grid-row` regardless of whether or not there's a heading, the `image` should still be inside of a `<div className="grid-col-auto">`?

src/components/Footer/Footer.stories.tsx
```diff
@@ -0,0 +1,102 @@
...
+              {item.heading}
+            </a>
+          </li>
+        ))}
+      </ul>
+    </nav>
+  )
+}
+
+const returnToTop = (
```
suzubara, line 57: Do you think this should be made a subcomponent as well? it could take an onClick handler (and i18n text)

src/components/Footer/Footer.tsx
haworku: Because a `grid-container` is not always used to wrap the secondary content.

PR #115, opened 2020-04-27
Remove css modules typescript loader, add generic css type declarations instead

### LINKED ISSUE #41: Generated files changing even if underlying code was not touched
Running a `yarn storybook` or `yarn build` or `yarn install` on any of the branches of this project seems to modify some generated files even if nothing was changed. Not sure what is causing this to be different than before.

```
❯ git status
On branch release-1.1.0
Your branch is up to date with 'origin/release-1.1.0'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   src/components/Alert/Alert.module.css.d.ts
        modified:   src/components/Modal/Modal.module.scss.d.ts
        modified:   src/components/Table/Table.module.css.d.ts
```

suzubara:
**Fixes #41** 

This stems from the underlying issue that TypeScript doesn't recognize `.css, .scss` files by default, and so you have to configure it if you are going to import styles into TS. The original solution was to use a webpack loader (https://github.com/seek-oss/css-modules-typescript-loader) which auto-generates `.d.ts` files for all (S)CSS files that are imported to TS, but it seemed to be buggy because it kept generating new versions with slightly different formatting (ie, with single quotes or without) even when there were no code changes. Because newer solutions have popped up, I decided to remove the loader instead of trying to dig into the issue.

I looked at a few alternative solutions but there doesn't seem to be a lot of consensus on the "right" way to handle this issue, so I ended up going with the easiest which is to add generic type declarations for (S)CSS module files.

Alternatives include:
- Other/newer webpack loaders or CLI tools that autogenerate the .d.ts flies such as:
  - https://github.com/Jimdo/typings-for-css-modules-loader
  - https://github.com/babakness/sass-module-types
  - https://github.com/dropbox/typed-css-modules-webpack-plugin
- A TypeScript service plugin that enhances IDE when using TS & (S)CSS modules: https://github.com/mrmckeb/typescript-plugin-css-modules

It is worth exploring these and deciding ultimately whether we want to generate & export .d.ts files for CSS modules in this library, but for now this PR should fix the issue where the generated files keep changing without actual code changes.


sojeri:
question unrelated to your changes but which I ran into when playing w/ this bug myself: why do we hook yarn build into any yarn call w/ the prepare script? I saw the change where you implemented this ( https://github.com/trussworks/react-uswds/commit/7266143734d4f4bbe282ad58e39128c1bff4965f ) but still don't quite understand the rationale.

suzubara:
> question unrelated to your changes but which I ran into when playing w/ this bug myself: why do we hook yarn build into any yarn call w/ the prepare script? I saw the change where you implemented this ( [7266143](https://github.com/trussworks/react-uswds/commit/7266143734d4f4bbe282ad58e39128c1bff4965f) ) but still don't quite understand the rationale.

ohh yes, we should remove this (and I _think_ replace with `prepublishOnly` ...). it's an artifact from when we were using this on SABER before it was a public package published on npm, and let us get around committing built files.

suzubara:
@sojeri I was going to include that in the upcoming release flow updates since it's related to releasing

PR #144, opened 2020-05-07
DateInput Component for Forms

eamahanna:
## PR Description
This PR creates the form date input component described [here](https://designsystem.digital.gov/components/form-controls/#date-input). 

## For Reviewers 
I am not sure if the hard coded Month, Day, Year strings should be props....probably because "hard coded", but I would like to hear your thoughts. Let me know if there is anything else that stands out to you.


trussworks-infra:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 2b2ac8ea0bfcc01bac82d6d67c0d9cf21089c80a
</p>

src/components/forms/DateInput/DateInput.tsx
```diff
@@ -0,0 +1,102 @@
...
+    hint,
+    className,
+    ...divAttributes
+  } = props
+
+  const classes = classnames('usa-memorable-date', className)
+
+  return (
+    <Fieldset legend={legend}>
+      <span className="usa-hint" id={`${id}_hint`}>
```
suzubara, line 68: We probably want to omit this `<span>` entirely if there is no `hint`. Also, I looked around the USWDS source code and it _looks_ like a `<span className="usa-hint">` can be present in any kind of fieldset element. So I might suggest creating a `forms/Hint` component. That way users can put hint text in any fieldset, and they can also assign whatever `id` (and other span attributes) they need to. What do you think?

suzubara: In fact, looking at the USWDS markup again I think maybe this component should _just_ render the `<div class="usa-memorable-date">` and leave the fieldset/hint for the user to add. IMO unlike the header/footer, the form components should be more composable to leave them open to a variety of use cases

src/components/forms/DateInput/DateInput.tsx
```diff
@@ -0,0 +1,102 @@
...
+  className?: string
+}
+
+const DateInputElement = (
+  props: DateInputElementProps & React.HTMLAttributes<HTMLElement>
+): React.ReactElement => {
+  const { id, name, label, maxLength, minLength } = props
+
+  return (
+    <div className="usa-form-group usa-form-group--month">
```
suzubara, line 33: This class needs to change based on which input it is (`usa-form-group--day`, etc.). Also there is a `forms/FormGroup` component I think you could use here.

src/components/forms/DateInput/DateInput.tsx
suzubara: Okay, I'm making one last suggestion about the structure of this component. The thing that gives me pause is that as-is, there's no way for a user to pass in custom props to the `<input>` components themselves, and this is often needed for a variety of reasons, such as hooking into a forms library or just for custom handling of events. So, at the risk of going back down the "what even are sub-components" rabbit hole, what if these components were exported separately & meant to be composed together by the user, such as...

```
<DateInputGroup> // renders <div class="usa-memorable-date"> with children
  <DateInput unit="month" /> // renders basically the DateInputElement you already have
  <DateInput unit="day" />
  <DateInput unit="year" />
</DateInputGroup>
```

... then the user can pass in custom label text, name, id, input attributes, etc. as they need, and they can even control which parts of the date are included and what order/structure they're rendered in. WDYT? I don't know if that `month/day/year` prop has to be called `unit` either, that is just what came to mind right away, but something else might be better.

eamahanna: I think your last comment makes a lot of sense. I think having the DateInputGroup and DateInput components is the way to go.

haworku: Nice. Like the idea of a new type of form group `DateInputGroup`.  I do think that the forms/hint component might also be useful generally. Maybe that's a new component to create an issue for, beyond the scope of this work.

src/components/forms/DateInputGroup/DateInputGroup.stories.tsx
```diff
@@ -0,0 +1,18 @@
...
+    info: `
+USWDS 2.0 DateInput component
+
+Source: https://designsystem.digital.gov/components/form-controls/#date-input
+`,
+  },
+}
+
+// TODO Determine if this is the way to display this component
+export const defaultDateInputGroup = (): React.ReactElement => (
```
suzubara, line 16: Since `DateInputGroup` on its own doesn't do much I wouldn't say we _need_ to have stories for it

src/components/forms/DateInputGroup/DateInputGroup.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+
+export const DateInputGroup = (
+  props: React.HTMLAttributes<HTMLElement>
+): React.ReactElement => {
+  const { children, ...divAttributes } = props
+
+  return (
+    <div
+      className="usa-memorable-date"
```
suzubara, line 10: Since `divAttributes` can include a custom `className`, we should use `classnames` here to append `props.className` to `usa-memorable-date` so it doesn't get overwritten

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## UNRELEASED
 
+- Added form dateInput and dateInputGroup components
```
haworku, line 10: Can remove this. No need for adjustments in the changelog anymore 🎉

src/components/forms/DateInput/DateInput.stories.tsx
```diff
@@ -0,0 +1,60 @@
...
+    name="testName"
+    label="Month"
+    unit="month"
+    maxLength={2}
+    minLength={2}
+  />
+)
+
+export const dateOfBirthExample = (): React.ReactElement => (
+  <Fieldset legend={'Date of birth'}>
```
haworku, line 29: 🎨 fyi can just be `legend="Date of Birth"` a `string` is a `ReactNode`. So is a number, null or several other types of values.

src/components/forms/DateInput/DateInput.stories.tsx
```diff
@@ -0,0 +1,60 @@
...
+  parameters: {
+    info: `
+USWDS 2.0 DateInput component
+
+Source: https://designsystem.digital.gov/components/form-controls/#date-input
+`,
+  },
+}
+
+export const defaultDateInput = (): React.ReactElement => (
```
suzubara, line 17: nitpick: technically I'd call this a `monthDateInput` instead of "default", and we could in fact have a story for each of `month, day, year`, but this is an enhancement not a blocker.

src/components/forms/DateInput/DateInput.stories.tsx
```diff
@@ -0,0 +1,60 @@
+import React from 'react'
+import { DateInput } from './DateInput'
+import { DateInputGroup } from '../DateInputGroup/DateInputGroup'
+import { Fieldset } from '../Fieldset/Fieldset'
+
+export default {
+  title: 'Form/DateInput',
```
suzubara, line 7: I just noticed because this is `Form` and not `Forms` this is resulting in a separate Storybook folder. not a blocker but we should clean it up

src/components/forms/DateInput/DateInput.stories.tsx
eamahanna: I added additional stories and renamed them :)

Issue #40, opened 2020-03-26
Investigate loki and other visual regression testing tools for storybook


gidjin:
As part of automated testing, I recommend we set up https://loki.js.org/ to run against storybook in this project, and the corresponding CircleCI job to run it so that contributions and improvements do not break the visual appearance of components unexpected.

haworku:
From my understanding to scope of this ticket has expanded.  The goal is now to review options visual regression testing (from [this list](https://storybook.js.org/docs/testing/automated-visual-testing/)) and then implement one. 

Capturing a few more points from Slack discussions:

- We want to specifically test our storybook stories, checking for visual regressions in components as rendered in stories. This is the use case.
- There are concerns about having to use docker to run `loki` tests effectively. This could be a barrier to entry for devs using this tool.
- As possible, we want automation for this type of testing as opposed to putting the onus on the developer to make sure they run a tool and check in new screenshots when there changes
- Whatever decisions are made there is a desire to document it thoroughly (maybe via ADR).

eamahanna:
@haworku your summary above is spot on. I am currently evaluating alternatives to loki for visual regression testing of story book. There are numerous tools that can do this, so I have narrowed down the research to the [tools listed within the storybook docs](https://storybook.js.org/docs/testing/automated-visual-testing/). 

To further eliminate options, I am only reviewing libraries/services that are open source or free to open source libraries. Additionally, I an only looking into libraries/services that have existing integration options for storybook. The potential options are listed below:

Open Source Options
- [Creevey](https://github.com/wKich/creevey)
- [Loki](https://loki.js.org/)
- [StoryShots + Puppeteer](https://github.com/storybookjs/storybook/tree/master/addons/storyshots/storyshots-puppeteer)

Free to Open Source Projects
- [Applitools](https://applitools.com/storybook/)
- [Chromatic](https://www.chromatic.com/)
- [Happo](https://happo.io/)

I believe the next step in narrowing down selection is determining whether we would like to continue with one of the open source options or the free to open source options. In general, the services that normally charge a fee have significantly better documentation than the open source alternatives, interactive UIs, and can easily integrate with github and circleci. However, the initial setup of non-open source options will require a little more effort because all of them require contact and communication with each of the companies to qualify react-uswds for the free open source specials. Additionally, the non-open source options feel potentially a little excessive in terms of bells and whistles for what this library needs. 

I would love to hear people's opinions and perspective on this.

eamahanna:
[...]

PR #158, opened 2020-05-14
Install loki #40

eamahanna:
## PR Description
This PR adds loki to preform visual regression testing for storybook. This tool looks at the rendered stories within storybook, creates *.png files of the stories, and compares them to the reference *.png files. This tool will add an additional testing step for developers. When a story's visual representation is changed, or potentially changed, the developer will need to complete the following steps:

`docker-compose run --rm storybook bash`
`yarn build-storybook`
`yarn loki test --chromeConcurrency 1 --requireReference --reactUri file:./storybook-static`
- Creates .png files of the current state of the stories and places them within the `.loki/current` directory. It also creates diff .png files (stored in `.loki/difference`) comparing the current .png files and the reference .png files (stored in `.loki/`). If there are any differences between the current and reference stories, the test for that story will fail. This is normal and expected. 

`yarn loki approve` <- must do this prior to committing changes
- Updates the reference images with the current images. 

Developers cannot run yarn loki test without being in docker because the dev's local environment will produce slightly different images (screenshots) than circleci's environment. This will lead to tests failing on circleci.

## For Reviewers 
I am unsure how to best document this new developer step. I would love suggestions on how to make the process clearer.

Additionally, I don't really like having to go through the 4 step process to test the new changes. Does it make sense to create a dockerfile and make file for local development? This would simplify the steps developers have to do to test.

P.S. This looks like a huge PR because it is adding all of the reference images for each component. The files are auto-generated.

## Potential Future Work

* Install `git-lfs`
* Execute `git lfs track "*.png"` (as suggested via in the loki documentation)
* Execute `git add .gitattributes` to maintain the lfs tracking

Milmove has had trouble using loki and IE11. Is this a concern for us? If so, I will need to change the storybook config file a bit. 

## References
[Loki](https://loki.js.org/getting-started.html)
[git-lfs](https://git-lfs.github.com/)


trussworks-infra:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 385bfcc18e52b58fb73d87140a16c1d7ffaab2c5
</p>

suzubara:
Can we include loki without requiring Docker? I know that applications like MilMove use it but since this is a library I don't want to require Docker as part of the dev environment. Also, I know we have to check in the loki reference images but do we also have to check in the static storybook files?

eamahanna:
> Can we include loki without requiring Docker? I know that applications like MilMove use it but since this is a library I don't want to require Docker as part of the dev environment. Also, I know we have to check in the loki reference images but do we also have to check in the static storybook files?

I don't believe we can use this library without docker. I first attempted to do this, but when loki is ran locally, the images created are slightly different than those created on circleci which led to circleci failing every time. 

We don't have to check in the storybook static files. I just updated the .gitignore and removed the files.

eamahanna:
@suzubara I just heard back from chromatic. Unfortunately the open source free plan only allows from chrome testing. It also has a snapshot limit of 35,000 per month. Given it only tests in chrome, I am not sure the collaboration tools out weight being able to in different browsers. What are your thoughts?

Issue #166, opened 2020-05-18
feat: Add `nav` variant to Link


suzubara:
We have an existing `Link` component that should be adjusted to handle a new `variant` prop called `nav`.

Requirements:
- Add `nav` as another possible value to the `StyledLinkProps` `variant` type definition
- Adjust the logic in the `linkClasses` function so that it includes the `usa-nav__link` class when the `variant` prop is `nav`
- Add a storybook example to `Link.stories.tsx` for `<Link variant="nav">`

haworku:
This could be accomplished by adding a `navLink` variant to the new `<Link />` #309

SirenaBorracha:
@suzubara or @haworku could you please review and let me know if anything looks off? Otherwise I'll put up a PR. My branch: https://github.com/trussworks/react-uswds/tree/ak-feat-add-nav-variant-to-link-166

<img width="616" alt="Screen Shot 2021-02-03 at 2 49 27 PM" src="https://user-images.githubusercontent.com/16230705/106819795-27133b80-662f-11eb-85aa-ae5ef552c631.png">

haworku:
@SirenaBorracha hello, new contributor 😄 

The ternary looks close but I think there's a change we still need. It is easier to discuss if we can comment on the code diff directly in a PR rather than here in an issue away from the code. Go ahead and open a PR from your branch. You can make a ["draft" PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/) and/or you can put the yellow "wip" work in progress label on it.

PR #181, opened 2020-05-18
Feat: Add Mobile styles to FooterNav

haworku:
# `<FooterNav>` and `FooterExtendedNavList>`

## Completed
- Add mobile styles and behavior
- Refactor out new `FooterExtendedNavList`
- Add tests

## Testing
View `Footer`, `FooterNav`, `FooterExtendedNavList` components in mobile styles.


src/components/Footer/FooterExtendedNavList/FooterExtendedNavList.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+      <FooterExtendedNavList nestedLinks={links} />
+    )
+    expect(container.querySelectorAll('a')).toHaveLength(5)
+    expect(getAllByText('Purple Rain')).toHaveLength(3)
+    expect(getAllByText('Cheetah')).toHaveLength(2)
+  })
+
+  describe('isMobile', () => {
+    it.todo(
+      'renders mobile view when client window width is less than threshold'
```
haworku, line 53: I can't figure out how to test window resize with running into typescript `read-only` limitations on reassigning `window` objects during a test.

src/components/Footer/FooterExtendedNavList/FooterExtendedNavList.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+    info: `
+     Display grouped nav items in an extended nav. Used in  "big" USWDS 2.0 Footer component.
+    
+    Source: https://designsystem.digital.gov/components/form-controls/#footer
+    `,
+  },
+}
+
+export const Default = (): React.ReactElement => (
+  <div className="usa-footer--big ">
```
haworku, line 18: It doesn't feel great React-wise to have to wrap a component to get it to display mobile styles. But there is a dependency in the CSS here.

src/components/Footer/FooterExtendedNavList/FooterExtendedNavList.tsx
```diff
@@ -0,0 +1,96 @@
...
+  React.HTMLAttributes<HTMLElement>): React.ReactElement => {
+  const classes = classnames('grid-row grid-gap-4', className)
+
+  const [isMobile, setIsMobile] = React.useState(false)
+  const [sectionsOpenState, setSectionsOpenState] = useState<boolean[]>(
+    Array(nestedLinks.length).fill(false)
+  )
+
+  useEffect(() => {
+    const isClient = window && typeof window === 'object'
```
haworku, line 26: 🤷‍♀️  I tried this after first implementing with an `isMobile` prop. It seemed strange, once I built out stories, to have a component that didn't expand itself on a wider screen. 

For reference mobile styles only start happening via CSS `.usa-footer--big .usa-footer__primary-content--collapsible.hidden .usa-list--unstyled {
    display: none;
}`

Happy to go back to to `isMobile` being a prop, would just remove the hook to implement and adjust tests. Just wanted to try out another option. Thoughts on this?

src/components/Footer/FooterNav/FooterNav.tsx
```diff
@@ -41,7 +43,9 @@ export const FooterNav = ({
 
   return (
     <nav className={navClasses} {...elementAttributes}>
-      {big && isExtendedNavLinks(links) && <ExtendedNav nestedLinks={links} />}
+      {big && isExtendedNavLinks(links) && (
+        <FooterExtendedNavList nestedLinks={links} />
+      )}
 
       {!isExtendedNavLinks(links) && (
         <ul className="grid-row grid-gap">
```
haworku, line 51: I was thinking to refactor `NavList` to have props/styles for `footerNav` and then use that component here.

src/components/Footer/FooterExtendedNavList/FooterExtendedNavList.stories.tsx
eamahanna: I ran into this in the header, and I couldn't find a way around it either. I agree, it definitely felt a little dirty...

src/components/Footer/FooterExtendedNavList/FooterExtendedNavList.tsx
```diff
@@ -0,0 +1,96 @@
+import React, { useState, useEffect } from 'react'
+import classnames from 'classnames'
+
+export type ExtendedNavLinksType = React.ReactNode[][]
+
+type FooterExtendedNavListProps = {
+  /* 
+    Multidimensional array of grouped nav links. Sub-arrays are column sections, first element is used as a heading.
+  */
+  nestedLinks: ExtendedNavLinksType
```
eamahanna, line 10: Would it be a bit more concise to just have the `nestedLinks: React.ReactNode[][]` ?

src/components/Footer/FooterExtendedNavList/FooterExtendedNavList.tsx
```diff
@@ -0,0 +1,96 @@
...
+  return (
+    <div className={classes}>
+      {nestedLinks.map((links, i) => (
+        <div
+          key={`linkSection-${i}`}
+          className="mobile-lg:grid-col-6 desktop:grid-col-3">
+          <Section
+            onToggle={isMobile ? (): void => onToggle(i) : undefined}
+            // eslint-disable-next-line security/detect-object-injection
+            isOpen={isMobile ? sectionsOpenState[i] : true}
```
eamahanna, line 58: Is it worrisome that you had to disable the eslint for this line?

trussworks-infra:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 50af0e66578d17725984969737525f4c509bc353
</p>

src/components/Footer/FooterExtendedNavList/FooterExtendedNavList.tsx
eamahanna: I think the only advantage to having isMobile being a prop, is it allows whoever is implementing the component to have more control over the component. I think this is in general what we have been attempting to do, but 🤷‍♀️

src/components/Footer/FooterExtendedNavList/FooterExtendedNavList.tsx
haworku: I don't think so. It was a warning, not an error,  basically it dislikes the index variable used as a lookup (injection would be if index was a user controlled that variable but its not user controlled).

src/components/Footer/FooterNav/FooterNav.tsx
haworku: Actually realized `NavList` is not as useful here since I have conditional styles on the `li` themselves. I can use the component in `FooterExtendedNavList` but not here without adding more props for list item classes to be passed in. Maybe this is where the #166 would be useful.

src/components/Footer/FooterExtendedNavList/FooterExtendedNavList.tsx
haworku: Hey maybe! I used a named type to export it and use it in multiple files. But that might be overkill since this isn't a complex type.

src/components/Footer/FooterExtendedNavList/FooterExtendedNavList.tsx
haworku: I'm just gonna make it a prop for simplicity at this point and to move this along.

src/components/Footer/FooterExtendedNavList/FooterExtendedNavList.stories.tsx
suzubara: Agreed, but ultimately this is a shortcoming of the USWDS CSS so IMO all we can do is work with it :( IMO as long as we document clearly that this component should only ever be used within the "big" Footer, that's good enough

src/components/Footer/FooterExtendedNavList/FooterExtendedNavList.test.tsx
suzubara: could you use the `resizeTo` method instead of reassigning the dimensions? https://developer.mozilla.org/en-US/docs/Web/API/Window/resizeTo

imo this would be a great test to have!

src/components/Footer/FooterExtendedNavList/FooterExtendedNavList.tsx
suzubara: Yeah, this is always a tricky problem. It looks like you've pretty closely replicated what USWDS is doing with their JS, so I say this is good for now. perhaps a future enhancement could be adding an `isMobile` prop that overwrites this behavior, for users who want more control?

src/components/Footer/FooterExtendedNavList/FooterExtendedNavList.tsx
```diff
@@ -0,0 +1,91 @@
...
+    Array(nestedLinks.length).fill(false)
+  )
+
+  useEffect(() => {
+    const isClient = window && typeof window === 'object'
+    if (!isClient) return
+
+    function handleResize(): void {
+      const isMobileWidth = window.innerWidth < 480
+      setIsMobile(isMobileWidth)
```
suzubara: my only suggestion here would be to _only_ call `setIsMobile` if the value has changed, to avoid extraneous state updates

src/components/Footer/FooterExtendedNavList/FooterExtendedNavList.tsx
```diff
@@ -0,0 +1,91 @@
...
+  isOpen = false,
+  links,
+  onToggle,
+}: {
+  isOpen: boolean
+  links: React.ReactNode[]
+  onToggle?: () => void
+}): React.ReactElement => {
+  const primaryLinkOrHeading = links[0]
+  const secondaryLinks = links.slice(1)
```
suzubara, line 64: this is purely fancy syntax stuff but you can also do
```
const [ primaryLinkOrHeading, ...secondaryLinks ] = links
```

src/components/Footer/FooterExtendedNavList/FooterExtendedNavList.tsx
```diff
@@ -0,0 +1,91 @@
...
+}
+
+export const FooterExtendedNavList = ({
+  className,
+  nestedLinks,
+}: FooterExtendedNavListProps &
+  React.HTMLAttributes<HTMLElement>): React.ReactElement => {
+  const classes = classnames('grid-row grid-gap-4', className)
+
+  const [isMobile, setIsMobile] = React.useState(false)
```
suzubara: This is super minor but I noticed there is a brief flash of the desktop nav at mobile sizes on render. I think that could be resolved by doing the initial window size check here to default the state to the correct value (instead of `false`) and then that would remove the need to trigger `handleResize()` in the effect hook

haworku:
@suzubara I incorporated the changes you suggested.  Tried to rename variables a bit to help clarify the difference between `isMobile` the prop and then the fallback state based on screen width.   However, still hitting a wall with the tests for screen size.  Lets manual test for now?

 I can create an issue about figuring this out seems like it would be a useful testing pattern to know how to implement.

src/components/Footer/FooterExtendedNavList/FooterExtendedNavList.test.tsx
haworku: Tried out `resizeTo` but its not implemented. jest incorporates JSDOM to override `window` - and JSDOM doesn't support all the same methods.

src/components/Footer/FooterExtendedNavList/FooterExtendedNavList.tsx
```diff
@@ -14,16 +14,37 @@ type FooterExtendedNavListProps = {
...
+  )
   const [sectionsOpenState, setSectionsOpenState] = useState<boolean[]>(
     Array(nestedLinks.length).fill(false)
   )
 
+  // Use isMobile prop, fallback to calculated state if undefined
+  const useMobile = isMobile || (isMobile === undefined && isMobileFallback)
+
+  useEffect(() => {
+    if (!isClient) return
```
suzubara, line 36: non-blocking: just for performance, couldn't we also do an early `return` if `isMobile` is defined? since then the fallback isn't used at all and we wouldn't need to attach the event listener.

haworku: YES.

src/components/Footer/FooterExtendedNavList/FooterExtendedNavList.tsx
```diff
@@ -14,16 +14,37 @@ type FooterExtendedNavListProps = {
...
+
+    function handleResize(): void {
+      if (isMobileWidth !== isMobileFallback) {
```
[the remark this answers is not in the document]
haworku: FYI https://github.com/trussworks/react-uswds/pull/181#discussion_r426928438  Also just double checked it this is the exact error in the console`  Error: Not implemented: window.resizeTo` on running jest.

src/components/Footer/FooterExtendedNavList/FooterExtendedNavList.test.tsx
```diff
@@ -108,5 +108,50 @@ describe('FooterExtendedNavList component', () => {
...
+      const elementsWithHiddenClass = container.querySelectorAll('.hidden')
+      expect(elementsWithHiddenClass.length).toEqual(0)
+    })
+
+    describe('when client window width is less than mobile threshold', () => {
+      beforeEach(() => {
+        // Mobile width is less than 480
+        //eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+        //@ts-ignore
+        window.innerWidth = 479
```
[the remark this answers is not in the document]
haworku: FYI https://github.com/trussworks/react-uswds/pull/181#discussion_r426928438  Also just double checked it this is the exact error in the console`  Error: Not implemented: window.resizeTo` on running jest.

src/components/Footer/FooterExtendedNavList/FooterExtendedNavList.tsx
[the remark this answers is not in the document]
haworku: Nice catch I wasn't seeing any issues with my testing and this made me look at again.  Definitely need to called `window.innerWidth` each time  in `handleResize` to get the right behaviors when lots of resizing of the window is happening.

src/components/Footer/FooterExtendedNavList/FooterExtendedNavList.test.tsx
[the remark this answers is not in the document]
suzubara: Ooh I see, got it! 👍

PR #183, opened 2020-05-19
fix: accept id and name props for Search component input field

### LINKED ISSUE #162: Search component should accept id, name props
Right now the Search component renders a label & input with hardcoded id and name attributes (`search-field` and `search`). It should be modified to accept props to define these instead, since the user will likely need full control over id and name attributes.

duncan-truss:
The Search component previously hardcoded the id and name attributes to 'search-field' and 'search', of the rendered input and label elements.

This adds `id` and `name` as props to the Search component allowing overriding of the default values.

Also updates the storybook documentation USWDS link to be correct.

Fixes #162

### DIFF
--- a/src/components/Search/Search.tsx
+++ b/src/components/Search/Search.tsx
@@ -8,6 +8,8 @@ import { TextInput } from '../forms/TextInput/TextInput'
 
 interface SearchInputProps {
   onSubmit: (event: React.FormEvent<HTMLFormElement>) => void
+  id?: string
+  name?: string
   big?: boolean
   small?: boolean
   label?: React.ReactNode
@@ -19,6 +21,8 @@ export const Search = (
 ): React.ReactElement => {
   const {
     onSubmit,
+    id = 'search-field',
+    name = 'search',
     big,
     small,
     label = 'Search',
@@ -41,10 +45,10 @@ export const Search = (
       role="search"
       search={true}
       {...formProps}>
-      <Label srOnly={true} htmlFor="search-field">
+      <Label srOnly={true} htmlFor={id}>
         {label}
       </Label>
-      <TextInput id="search-field" type="search" name="search" />
+      <TextInput id={id} type="search" name={name} />
       <Button type="submit">
         <span className={small ? 'usa-sr-only' : 'usa-search__submit-text'}>
           Search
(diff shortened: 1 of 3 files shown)


src/components/Search/Search.tsx
```diff
@@ -19,6 +21,8 @@ export const Search = (
 ): React.ReactElement => {
   const {
     onSubmit,
+    id = 'search-field',
```
haworku, line 24: Lets name this `inputId` since `id` is an existing HTML attribute that could be captured in`...formProps` and passed down to the `<form>` also.

yarn.lock
```diff
@@ -2404,16 +2404,6 @@
     regexpp "^3.0.0"
     tsutils "^3.17.1"
 
-"@typescript-eslint/experimental-utils@2.32.0":
-  version "2.32.0"
```
haworku, line 2408: Suggestion - delete the `yarn.lock` if you didn't change any dependencies in the PR. Maybe we need to make a separate PR to update `develop` but there's also a release in flight that will get merged back down ... hoping that will clear this out.

src/components/Search/Search.tsx
duncan-truss: Yep that occurred to me, thanks for the explanation.  Would `className` ever clash as well?

yarn.lock
duncan-truss: 👍 I'll revert it to what's on develop

trussworks-infra:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 6c72a1cb834754a2ddcb626cc9af9106e5a9dc92
</p>

src/components/Search/Search.tsx
suzubara: since `className` is part of the destructure assignment, it won't be included in `formProps` so it should be fine. if it _wasn't_ destructured_ (and appended to the other classes using `classnames`) then yes, it could clash with the existing `className` prop, so it's important to handle that one specially :D

Issue #81, opened 2020-04-15
USWDS Component: Range slider


suzubara:
Add Range slider component: https://designsystem.digital.gov/components/form-controls/#range

haworku:
Suggested scope:
- [ ] New component should be added to the existing `Form` folder
- [ ] Props must handle HTML [range input ](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range )behaviors. This include `min`, `max`, `step`and the input `value`.  Will also need to be able to call an event handler on change to update parent.
- [ ] Both keyboard interaction or click to drag should move the slider
- [ ] As per uswds docs,  "set the `step` attribute so it’s not too granular"
- [ ] As per uswds docs, provide a way for optional labels to be added to the beginning and end of slider to describe the min and max values.

Issue #73, opened 2020-04-15
Add more documentation around how to contribute and write new components


suzubara:
Write some more documentation specifically around how to write new components. Include guidance around:
- How to decide whether to add a new component
- Once decided, make sure it is abstract enough for general usage but also fits in the USWDS
- Make sure props type inherits the HTML element attributes
- Make sure to add corresponding stories and tests
- Make sure to export the new component from the package entrypoint (`index.ts`)

Add a PR template.

PR #194, opened 2020-05-21
feat(rangeinput): creates form RangeInput component, test, and stories

duncan-truss:
## Description

This PR adds the [USWDS Range](https://designsystem.digital.gov/components/form-controls/#range) component to the available form inputs.

The component renders as an input of type range and an optional Label.

The min, max, step, and value fields are optional and default to:
**min**: 0
**max**: 100
**step**: 1
**value**: max/2

I copied the ref code from somewhere else but don't know entirely why it's necessary or all of the possible types it could be.  I think I remember there is an old deprecated way to do refs from the official tutorial.

Screen Shot 2020-05-21 at 6 09 42 PM

## Open Questions
- Should the label be excluded all together from the component or passed whole instead of built out of label and hint attributes?
- The USWDS guidance recommends that you label the min and max ends, should we offer that functionality?
- Should the component accept any event listeners?
- I added the list prop to support linking to a datalist element.  This has next to no browser visual support (ticks in chrome if the `usa-range` class isn't applied) at the moment.
- Similar to the start and end labels should there be an `<output>` field that displays the slider value or would that be handled externally by something that uses an event listener?
- Firefox accepts an orient="vertical" attribute but everyone else just uses CSS so I omitted it

[MDN input range element documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range)

#81

### DIFF
--- a/src/components/forms/RangeInput/RangeInput.stories.tsx
+++ b/src/components/forms/RangeInput/RangeInput.stories.tsx
@@ -1,6 +1,5 @@
 import React from 'react'
 import { RangeInput } from './RangeInput'
-import Label from '../Label/Label'
 
 export default {
   title: 'Forms/RangeInput',
@@ -27,7 +26,7 @@ const labelChildren = (
 const labelHint = <>(Some hint)</>
 
 export const defaultRange = (): React.ReactElement => (
-  <RangeInput label="Range slider" />
+  <RangeInput id="range-slider" name="range" />
 )
 
 export const customRange = (): React.ReactElement => (
@@ -35,7 +34,6 @@ export const customRange = (): React.ReactElement => (
     id="custom-range-slider"
     name="rangeValue"
     className="dark-theme"
-    label="Range slider"
     min={1}
     max={11}
     step={2}
@@ -43,15 +41,12 @@ export const customRange = (): React.ReactElement => (
   />
 )
 
-export const richLabelRange = (): React.ReactElement => (
-  <RangeInput label={labelChildren} hint={labelHint} />
-)
-
 // Only tick marks are shown in Chrome but not with usa-range class currently because the appearance property is set to none
 export const dataListRange = (): React.ReactElement => (
   <>
     <RangeInput
-      label="Range from datalist"
+      id="range-slider"
+      name="range"
       min={0}
       max={4}
       defaultValue={2}
--- a/src/components/forms/RangeInput/RangeInput.stories.tsx
+++ b/src/components/forms/RangeInput/RangeInput.stories.tsx
@@ -1,5 +1,6 @@
 import React from 'react'
 import { RangeInput } from './RangeInput'
+import { Label } from '../Label/Label'
 
 export default {
   title: 'Forms/RangeInput',
@@ -18,15 +19,20 @@ const labelChildren = (
       <span>Start:</span> <span>0</span>
     </div>
     <div>
-      <span>End:</span> <span>10</span>
+      <span>End:</span> <span>100</span>
     </div>
   </>
 )
 
-const labelHint = <>(Some hint)</>
+const labelHint = <>(drag to adjust or use arrow keys)</>
 
 export const defaultRange = (): React.ReactElement => (
-  <RangeInput id="range-slider" name="range" />
+  <>
+    <Label htmlFor="range-slider" hint={labelHint}>
+      {labelChildren}
+    </Label>
+    <RangeInput id="range-slider" name="range" />
+  </>
 )
 
 export const customRange = (): React.ReactElement => (
--- a/src/components/forms/RangeInput/RangeInput.stories.tsx
+++ b/src/components/forms/RangeInput/RangeInput.stories.tsx
@@ -1,0 +1,68 @@
+import React from 'react'
+import { RangeInput } from './RangeInput'
+import Label from '../Label/Label'
+
+export default {
+  title: 'Forms/RangeInput',
+  parameters: {
+    info: `
+USWDS 2.0 RangeInput component
+
+Source: https://designsystem.digital.gov/components/form-controls/#range
+`,
+  },
+}
+
+const labelChildren = (
+  <>
+    <div>
+      <span>Start:</span> <span>0</span>
+    </div>
+    <div>
+      <span>End:</span> <span>10</span>
+    </div>
+  </>
+)
+
+const labelHint = <>(Some hint)</>
+
+export const defaultRange = (): React.ReactElement => (
+  <RangeInput label="Range slider" />
+)
+
+export const customRange = (): React.ReactElement => (
+  <RangeInput
+    id="custom-range-slider"
+    name="rangeValue"
+    className="dark-theme"
+    label="Range slider"
+    min={1}
+    max={11}
+    step={2}
+    defaultValue={3}
+  />
+)
+
+export const richLabelRange = (): React.ReactElement => (
+  <RangeInput label={labelChildren} hint={labelHint} />
+)
+
+// Only tick marks are shown in Chrome but not with usa-range class currently because the appearance property is set to none
+export const dataListRange = (): React.ReactElement => (
+  <>
+    <RangeInput
+      label="Range from datalist"
+      min={0}
+      max={4}
+      defaultValue={2}
+      list="range-list-id"
+    />
+    <datalist id="range-list-id">
+      <option>0</option>
+      <option>1</option>
+      <option>2</option>
+      <option>3</option>
+      <option>4</option>
+    </datalist>
+  </>
+)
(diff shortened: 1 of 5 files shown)


eamahanna:
Here are my thoughts on the questions you listed in your description:

_Should the label be excluded all together from the component or passed whole instead of built out of label and hint attributes?_
* I think it is fine including the label or not. The text input component does not include a label however. 

_The USWDS guidance recommends that you label the min and max ends, should we offer that functionality?_
* I think there should at least be a story showing this. I am unsure if the component should contain them because I don't know how they are suppose to be styled. Maybe there is guidance on that.

_Should the component accept any event listeners?_
* I don't think you need to include a listener explicitly. There is on onChange prop defined in the React.InputHTMLElement interface 
```
 interface InputHTMLAttributes<T> extends HTMLAttributes<T> {
        accept?: string;
        alt?: string;
        autoComplete?: string;
        autoFocus?: boolean;
        capture?: boolean | string; // https://www.w3.org/TR/html-media-capture/#the-capture-attribute
        checked?: boolean;
        crossOrigin?: string;
        disabled?: boolean;
        form?: string;
        formAction?: string;
        formEncType?: string;
        formMethod?: string;
        formNoValidate?: boolean;
        formTarget?: string;
        height?: number | string;
        list?: string;
        max?: number | string;
        maxLength?: number;
        min?: number | string;
        minLength?: number;
        multiple?: boolean;
        name?: string;
        pattern?: string;
        placeholder?: string;
        readOnly?: boolean;
        required?: boolean;
        size?: number;
        src?: string;
        step?: number | string;
        type?: string;
        value?: string | string[] | number;
        width?: number | string;

        onChange?: ChangeEventHandler<T>;   }
```
_I added the list prop to support linking to a datalist element. This has next to no browser visual support (ticks in chrome if the usa-range class isn't applied) at the moment._
* Given that it is optional and defined in the InputHTMLAttributes, I don't think it needs to be an explicit prop, but you can show it in a story maybe?

_Similar to the start and end labels should there be an field or would that be handled externally by something that uses an event listener?_
* I think this is the inputHTMLAttribute onChange prop

duncan-truss:
> Here are my thoughts on the questions you listed in your description:
> 
> _Should the label be excluded all together from the component or passed whole instead of built out of label and hint attributes?_
> 
>     * I think it is fine including the label or not. The text input component does not include a label however.
> 
> 
> _The USWDS guidance recommends that you label the min and max ends, should we offer that functionality?_
> 
>     * I think there should at least be a story showing this. I am unsure if the component should contain them because I don't know how they are suppose to be styled. Maybe there is guidance on that.
> 
> 
> _Should the component accept any event listeners?_
> 
>     * I don't think you need to include a listener explicitly. There is on onChange prop defined in the React.InputHTMLElement interface
> 
> 
> ```
>  interface InputHTMLAttributes<T> extends HTMLAttributes<T> {
>         accept?: string;
>         alt?: string;
>         autoComplete?: string;
>         autoFocus?: boolean;
>         capture?: boolean | string; // https://www.w3.org/TR/html-media-capture/#the-capture-attribute
>         checked?: boolean;
>         crossOrigin?: string;
>         disabled?: boolean;
>         form?: string;
>         formAction?: string;
>         formEncType?: string;
>         formMethod?: string;
>         formNoValidate?: boolean;
>         formTarget?: string;
>         height?: number | string;
>         list?: string;
>         max?: number | string;
>         maxLength?: number;
>         min?: number | string;
>         minLength?: number;
>         multiple?: boolean;
>         name?: string;
>         pattern?: string;
>         placeholder?: string;
>         readOnly?: boolean;
>         required?: boolean;
>         size?: number;
>         src?: string;
>         step?: number | string;
>         type?: string;
>         value?: string | string[] | number;
>         width?: number | string;
> 
>         onChange?: ChangeEventHandler<T>;   }
> ```
> 
> _I added the list prop to support linking to a datalist element. This has next to no browser visual support (ticks in chrome if the usa-range class isn't applied) at the moment._
> 
>     * Given that it is optional and defined in the InputHTMLAttributes, I don't think it needs to be an explicit prop, but you can show it in a story maybe?
> 
> 
> _Similar to the start and end labels should there be an field or would that be handled externally by something that uses an event listener?_
> 
>     * I think this is the inputHTMLAttribute onChange prop

Thank you for these answers!

Since the InputHTMLAttributes contain min, max, and step should I remove them from the RangeInputProps interface along with list?

haworku:
@duncan-truss to answer your questions about the type interface: yes - remove them from there. We don't need to retype them.

src/components/forms/RangeInput/RangeInput.tsx
```diff
@@ -0,0 +1,63 @@
...
+    | undefined
+}
+
+export const RangeInput = (
+  props: RangeInputProps & React.InputHTMLAttributes<HTMLInputElement>
+): React.ReactElement => {
+  // Range defaults to min = 0, max = 100, step = 1, and value = (max/2) if not specified.
+  const {
+    id = 'range-slider',
+    name = 'range',
```
suzubara, line 25: this is my 🔥  opinion and is more of a general comment about form inputs, but IMO `id` and `name` are so fundamental to building HTML forms that I don't think we need to provide default values, and I also think they should be required in the prop type.

src/components/forms/RangeInput/RangeInput.tsx
```diff
@@ -0,0 +1,63 @@
...
+    list,
+    inputRef,
+    ...inputProps
+  } = props
+
+  const classes = classnames('usa-range', className)
+
+  return (
+    <>
+      {label && (
```
suzubara, line 41: To answer your question, the label should _not_ be included in the component (following the examples of the other input components, which _only_ render an `<input>` element). It should be up to the user to add labels.

src/components/forms/RangeInput/RangeInput.tsx
```diff
@@ -0,0 +1,63 @@
...
+        data-testid="range"
+        name={name}
+        className={classes}
+        ref={inputRef}
+        type="range"
+        min={min}
+        max={max}
+        step={step}
+        list={list}
+        {...inputProps}
```
suzubara, line 57: suggestion: any of these props that aren't handled specifically in our code (ie, `className`) can be passed in as part of `...inputProps` instead of having to be destructured & passed in explicitly.

suzubara:
> I copied the ref code from somewhere else but don't know entirely why it's necessary or all of the possible types it could be. I think I remember there is an old deprecated way to do refs from the official tutorial.

This `inputRef` prop is one that each `input` should accept, and allows an implementation to have direct access to the input DOM element which is often required for things like form library integration, or managing focus on form elements.

> * The USWDS guidance recommends that you label the min and max ends, should we offer that functionality?

Since USWDS doesn't provide any examples or styles I think we can leave it for now

src/components/forms/RangeInput/RangeInput.tsx
duncan-truss: That's great context, keep the opinions coming I don't have any of my own yet.

haworku:
> This inputRef prop is one that each input should accept, and allows an implementation to have direct access to the input DOM element which is often required for things like form library integration, or managing focus on form elements.  

Ah, makes sense @suzubara  that's great context - I wasn't sure either about this tbh (the use case specifically). This is an example of something we can include in #73 . 
 
Regarding the types, just generally I wonder if something like `inputRef` should be a shared type in the `/forms` folder.  Maybe we just define it in `TextInput` for now and export it for use?

trussworks-infra:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against a6e5fb59e372b6388e5469fbe7508d2e03bec971
</p>

src/components/forms/RangeInput/RangeInput.stories.tsx
```diff
@@ -0,0 +1,63 @@
...
+    <div>
+      <span>Start:</span> <span>0</span>
+    </div>
+    <div>
+      <span>End:</span> <span>10</span>
+    </div>
+  </>
+)
+
+const labelHint = <>(Some hint)</>
```
suzubara, line 26: it looks like there are a few unused variables now that can be deleted. FWIW I think it would be fine to include a story that renders the `RangeInput` alongside a `Label` if you wanted to demo that.

src/components/forms/RangeInput/RangeInput.tsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react'
+import classnames from 'classnames'
+
+import { Label } from '../Label/Label'
```
suzubara, line 4: this import can be deleted

Issue #168, opened 2020-05-18
New component: Card


suzubara:
https://designsystem.digital.gov/components/card/

eamahanna:
## Issue Description
This goal of this issue is the add the card component to the react-uswds library. This component was added to USWDS in version 2.7.0.

## Issue Requirements
1. Create a card component that provides the ability for a user to implement the following card styles:
* Card, Card with media, Media and header first, Inset media, Exdent media, flag default, flag media on right
* Props for the card component could include: 
```
layout: 'standard' | 'flag', 
headerFirst: boolean, 
media?: React.ReactNode, 
mediaStyle?: 'inset' | 'exdent', 
flagOrientation: 'left' | 'right' (left would be the default)
```
2. Create card group component
* Props for this component could include:
```
items/cards: (this would always be card component items)
``` 
3. Create stories to display each of the styles stated above
4. Implement tests 
5. Investigate and mirror mobile UI functionality
6. Add component to exports

PR #205, opened 2020-05-26
feat: add card component #168

eamahanna:
## PR Description
This PR implements the uswds card component 
* https://designsystem.digital.gov/components/card/
* https://components.designsystem.digital.gov/components/detail/card--compare.html

Included in this PR:
* Card component
* CardGroup Component
* Tests for each component
* Stories showing card styling options

## For Reviewers
I am unsure of how to incorporate the grid styling. At this moment it is just hardcoded in the stories. Is there a better way to do this? Also, I am not sure, but it doesn't look like the `usa-card__media--set-aspect.add-aspect-1x1` class is doing anything. Do you know if I should implement this differently?


trussworks-infra:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 145e9524ca782bb2ba5df1d00ee538c5c411e045
</p>

src/components/card/card.stories.tsx
```diff
@@ -0,0 +1,339 @@
...
+      </header>
+    }
+    footer={
+      <div className="usa-card__footer">
+        <Button type="button" className="usa-button">
+          Example Button
+        </Button>
+      </div>
+    }>
+    <div className="usa-card__body">
```
duncan-truss, line 34: What do we get from making this part of children props instead of explicitly assigning it to something like body?

From what I can tell the <header> and <div> elements are always necessary so should that be internal to the component instead of expecting them to be specified everytime with the right classes?  I could see Media maybe being it's own component but less so the header and footer content.

src/components/card/card.stories.tsx
```diff
@@ -0,0 +1,339 @@
...
+      <div className="usa-card__footer">
+        <Button type="button" className="usa-button">
+          Example Button
+        </Button>
+      </div>
+    }
+    media={
+      <div className="usa-card__media usa-card__media--set-aspect.add-aspect-1x1">
+        <div className="usa-card__img">
+          {/* Are we allowed to use this image? */}
```
duncan-truss, line 92: It should be fine to use, do we want to make it a local image though checked into the repo though?

src/components/card/card.stories.tsx
```diff
@@ -0,0 +1,339 @@
...
+    }
+    footer={
+      <div className="usa-card__footer">
+        <Button type="button" className="usa-button">
+          Example Button
+        </Button>
+      </div>
+    }
+    media={
+      <div className="usa-card__media usa-card__media--set-aspect.add-aspect-1x1">
```
duncan-truss, line 90: `add-aspect-1x1` should be it's own class not part of `usa-card__media--set-aspect`.  I also saw there is a `usa-card__media--fixed-aspect` combined with add-aspect-* in the card test.

duncan-truss: You may want to use an image that has a larger height than width to see the effect on at responsive breakpoints.

src/components/card/Card/Card.tsx
```diff
@@ -0,0 +1,49 @@
...
+    },
+    className
+  )
+
+  return (
+    <li className={classes} data-testid="Card">
+      <div className="usa-card__container" {...divProps}>
+        {header}
+        {media}
+        {children}
```
haworku, line 42: The card body styles are important to make sure the flex layout works for this component overall. I think we should pull the styling into our Card component - something like this-   `<div className="usa-card__body">{children}</div>` instead.

haworku: I think our header/footer styles could also be in the main component for the same reason of making the component act as a styles wrapper.  Something like  `header && <header className="whateverClasses>{header}</header>` in the case of the header.

src/components/card/card.stories.tsx
haworku: I agree the body, header, footer style wrappers could be internal to the component (commented on that above). 

Not opinionated on the `children` question. I think its kind of 🎨 style thing here- often children are for the main body of  component that could have lots of different types of content.

src/components/card/Card/Card.tsx
eamahanna: That makes sense :)

src/components/card/card.stories.tsx
suzubara: Yeah, I think this is a style preference.. personally I like using `props.children` where it makes sense semantically (which in this case I think it does) because I find it easier to parse:

```
<Card>
  <p>Here is my content!</p>
</Card>
```

than:
```
<Card
  body={<p>Here is my content!</p>} />
```

especially if it contains a lot of content.

src/components/card/Card/Card.test.tsx
```diff
@@ -0,0 +1,60 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+
+import { Card } from './Card'
+
+describe('CardGroup component', () => {
```
suzubara, line 6: is this meant to be `Card component`?

src/components/card/Card/Card.tsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react'
+import classnames from 'classnames'
+
+interface CardProps {
+  header?: React.ReactNode
+  media?: React.ReactNode
+  footer?: React.ReactNode
+  standardLayout?: 'default' | 'headerFirst'
+  flagLayout?: 'default' | 'mediaOnRight'
+}
```
suzubara, line 10: I'm finding the naming of the layout props confusing, and also could be problematic if you can specify both a `standardLayout` and a `flagLayout`? What if the prop was like:
```
layout: 'standardDefault' | 'headerFirst' | 'flagDefault' | 'flagMediaRight'
```

src/components/card/Card/Card.tsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react'
+import classnames from 'classnames'
+
+interface CardProps {
+  header?: React.ReactNode
+  media?: React.ReactNode
+  footer?: React.ReactNode
```
suzubara, line 7: It's my impression that `header`, `children`, and `footer` should be required (with only `media` being optional). We might want to test how the Card looks when any of those are omitted to see if there are repercussions in the styling if we want them to be optional.

src/components/card/Card/Card.tsx
```diff
@@ -0,0 +1,49 @@
...
+    media,
+    footer,
+    standardLayout,
+    flagLayout,
+    children,
+    className,
+    ...divProps
+  } = props
+
+  // TODO determine how to implement the tablet:grid-col-4 class properly, should this be li class?
```
suzubara, line 26: Assuming the `li` can handle any of the various responsive grid classes, maybe this component can share the grid prop structure and class helper function located in https://github.com/trussworks/react-uswds/blob/develop/src/components/grid/Grid/Grid.tsx#L11 ? That could be extracted out and shared across any components that support grid classes.

src/components/card/Card/Card.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const classes = classnames(
+    'usa-card',
+    {
+      'usa-card--header-first': standardLayout == 'headerFirst',
```
suzubara, line 30: We should be enforcing `===` for equality operators... I'm a little surprised it doesn't seem to be in our eslint config. we should look into that

src/components/card/card.stories.tsx
```diff
@@ -0,0 +1,339 @@
+import React from 'react'
```
suzubara, line 1: nitpicky, but can we keep file name casing consistent? so `Card.stories.tsx`

src/components/card/card.stories.tsx
```diff
@@ -0,0 +1,339 @@
...
+      <div className="usa-card__footer">
+        <Button type="button" className="usa-button">
+          Example Button
+        </Button>
+      </div>
+    }
+    media={
+      <div className="usa-card__media">
+        <div className="usa-card__img">
+          {/* Are we allowed to use this image? */}
```
suzubara, line 59: it looks like https://unsplash.com/ is for free images, so sounds fine with me! I vote for let's not add it to the repo.

src/components/card/Card/Card.tsx
eamahanna: 🤦

src/components/card/Card/Card.tsx
eamahanna: I am going to keep the header, footer, and the body? as separate components so they are most customizable.

src/components/card/Card/Card.tsx
eamahanna: I am not quite sure how to do this? Can you explain further?

src/components/card/Card/Card.tsx
eamahanna: I have made the header and footer as separate components. @haworku @suzubara Do you think there should be media and body components as well?

eamahanna: @suzubara and I discussed the structure of the card component and came up with the following three possibilities:
```
<Card>   
    <CardHeader className="thing">...</CardHeader> 
</Card>  

<Card   header={<CardHeader className="thing">...</CardHeader>}   />

  <Card   header="MY header"   headerClasses="thing" />

```
We decided to go with the first option for simplicity. This will result in header, footer, media and body components.

src/components/card/Card/Card.tsx
eamahanna: Paired with @suzubara to implement.

src/components/card/card.stories.tsx
eamahanna: Resolved by making individual components for header, footer, body, and media. Card just renders children.

src/components/card/Card/Card.tsx
eamahanna: Thanks for finding this! I added a header first prop to the card component.

src/components/card/Card/Card.tsx
```diff
@@ -0,0 +1,46 @@
...
+      'usa-card--flag': layout === 'flagDefault' || layout === 'flagMediaRight',
+      'usa-card--media-right': layout === 'flagMediaRight',
+    },
+    gridClasses,
+    className
+  )
+
+  return (
+    <li className={classes} data-testid="Card">
+      <div className="usa-card__container" {...divProps}>
```
suzubara, line 39: There's a slight inconsistency here that I could see being confusing. take this example:
```
<Card className="my-custom-class" id="myCard">
```
will result in the `className` on the `<li>` and the `id` on the `<div>`:
```
<li className="my-custom-class">
  <div id="myCard">
```

While I agree it may be useful to pass in attributes for the underlying `div`, maybe for this first implementation let's keep it simple and say the implicit props are all for the `li` instead?

src/components/card/CardBody/CardBody.test.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+
+import { CardBody } from './CardBody'
+
+describe('CardGroup component', () => {
```
suzubara, line 6: typo: `CardBody`

src/components/card/CardFooter/CardFooter.test.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+
+import { CardFooter } from './CardFooter'
+
+describe('CardGroup component', () => {
```
suzubara, line 6: typo: `CardFooter`

src/components/card/CardGroup/CardGroup.test.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { CardGroup } from './CardGroup'
+
+describe('CardGroup component', () => {
+  it('renders without errors', () => {
+    const { queryByTestId } = render(<CardGroup />)
+    expect(queryByTestId('CardGroup')).toBeInTheDocument()
+  })
+
+  it('renders its children', () => {
+    const { queryByText } = render(<CardGroup>My Content</CardGroup>)
```
suzubara, line 13: non-blocking, but this may result in a warning since a `<ul>` should only have `<li>` as children

src/components/card/CardHeader/CardHeader.test.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+
+import { CardHeader } from './CardHeader'
+
+describe('CardGroup component', () => {
```
suzubara, line 6: typo: `CardHeader`

src/components/card/CardMedia/CardMedia.test.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+
+import { CardMedia } from './CardMedia'
+
+describe('CardGroup component', () => {
```
suzubara, line 6: typo: `CardMedia`

src/components/card/card.stories.tsx
```diff
@@ -0,0 +1,265 @@
...
+    </CardBody>
+    <CardFooter>
+      <Button type="button" className="usa-button">
+        Example Button
+      </Button>
+    </CardFooter>
+  </Card>
+)
+
+// TODO UPdate with ed
```
suzubara, line 144: is this an outstanding todo?

src/components/card/Card/Card.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react'
+import classnames from 'classnames'
+
+import { GridLayoutProp, applyGridClasses } from '../../grid/Grid/Grid'
+
+interface CardProps {
+  layout: 'standardDefault' | 'flagDefault' | 'flagMediaRight'
```
suzubara, line 7: just a nitpick while I'm testing out adding some more examples to Storybook, could we make `standardDefault` the default value for layout?

src/components/card/CardMedia/CardMedia.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react'
+import classnames from 'classnames'
+
+interface CardMediaProps {
+  exdent?: boolean
```
suzubara, line 5: also a nitpick from testing, it looks like `inset` is another option for CardMedia that we might want to add as a prop (adds class `usa-card__media--inset`, see https://components.designsystem.digital.gov/components/detail/card--test.html)

src/components/card/Card/Card.tsx
suzubara: after adding more examples with Storybook I'm gonna say this is a higher priority change.. if it sounds okay with you @eamahanna I think this should get in before being merged

src/components/card/CardHeader/CardHeader.test.tsx
eamahanna: 🤦

src/components/card/CardMedia/CardMedia.test.tsx
eamahanna: 🤦‍♀️

src/components/card/card.stories.tsx
eamahanna: No it was old

src/components/card/CardMedia/CardMedia.tsx
eamahanna: good catch

src/components/card/Card/Card.tsx
```diff
@@ -0,0 +1,48 @@
...
+    layout = 'standardDefault',
+    headerFirst,
+    children,
+    className,
+    gridLayout,
+    containerProps,
+    ...divProps
+  } = props
+
+  // TODO determine how to implement the tablet:grid-col-4 class properly, should this be li class?
```
haworku, line 25: Remove todo

src/components/card/Card/Card.tsx
```diff
@@ -0,0 +1,48 @@
...
+      'usa-card--flag': layout === 'flagDefault' || layout === 'flagMediaRight',
+      'usa-card--media-right': layout === 'flagMediaRight',
+    },
+    gridClasses,
+    className
+  )
+
+  return (
+    <li className={classes} data-testid="Card" {...divProps}>
+      <div className="usa-card__container" {...containerProps}>
```
haworku, line 41: There are cases in the test cards of this container also needing classes passed down. For example, the `border-ink` class is applied to some of the test cards.

Bug: The way it is implemented right now the `containerProps` className would override the existing `usa-card__container` className.

src/components/card/card.stories.tsx
haworku: @eamahanna can you take a look at this again? To rename files with git you may have to use `git mv` to get the renaming tracked

src/components/card/card.stories.tsx
```diff
@@ -0,0 +1,512 @@
...
+    </CardHeader>
+    <CardMedia imageClass="add-aspect-3x1">
+      <img
+        src="https://images.unsplash.com/photo-1484100356142-db6ab6244067?ixlib=rb-1.2.1&ixid=eyJhcHBfaWQiOjEyMDd9&auto=format&fit=crop&w=962&q=80"
+        alt="A placeholder"
+      />
+    </CardMedia>
+    <CardBody>
+      <p>
+        {' '}
```
haworku, line 73: Lets clear out the `{' '}` that leftovers from prettier I believe

src/components/card/card.stories.tsx
```diff
@@ -0,0 +1,512 @@
...
+    info: `
+USWDS 2.0 Card components
+
+Source: https://designsystem.digital.gov/components/card/
+`,
+  },
+}
+
+const card = (
+  <Card gridLayout={{ tablet: { col: 6 } }}>
```
haworku, line 23: can we have all of our cards here be `col: 4` so its easier to see them side by side.  Now that we have the test stories also included there still are opportunities to see cards with different grid classes applied elsewhere

src/components/card/card.stories.tsx
```diff
@@ -0,0 +1,512 @@
...
+    <CardBody exdent>
+      <p>
+        This is a standard card with the header and media first, media, and a
+        button in the footer. All of which are exdent style.
+      </p>
+    </CardBody>
+    <CardFooter exdent>
+      <Button type="button" className="usa-button">
+        Example Button
+      </Button>
```
haworku, line 183: I can't figure out what these footer and header exdent styles are supposed to show?  The button also looks weird. with these styles. Do we need to add this story example?

eamahanna: I am not quite sure what the styles are suppose to implement either, but they are included in the USWDS documentation for the component, so I thought it best to include them still. I think it makes sense to have a story for them, given they are called out directly in the documentation.

src/components/card/card.stories.tsx
eamahanna: sure

src/components/card/Card/Card.tsx
eamahanna: How would you fix this bug? I don't think it makes sense to add a containerClass prop...

Issue #124, opened 2020-04-30
Decide how we want to handle lang strings in components


haworku:
There are some new components that could include lang strings- see "Return to top" in `Footer` #78  and `SearchInput` #104 . Since sites using the library can be in another language, we are creating components that won't be flexible to internationalization.  

One solution could be that we don't include any hardcode any strings in our components (or only as a default value that can be overridden by a prop). Other thoughts or ideas?

sojeri:
I like the default value overridden by a prop.

If we want to support internationalized content, I wonder if we should update our storybook to showcase this w/ long strings and RTL behaviors.

tinyels:
I like the default prop idea. I'd also like us to think about what sort of recommendations we want to make with respect to localization. @scarequotes and I have had a couple of conversations about how setting up projects for localization would make it easier for content strategists to manage the language in projects. That is, they could provide initial language and even submit changes without having to pull an engineer in.

haworku:
Hey everyone circling back here!  I'd like to propose closing this issue since it sounds like we have agreement about using props (with English default values) to handle any necessary language strings.

 @sojeri raised something we could capture as a separate issue (updating storybook to showcase localization support - RTL, examples of long strings, etc).   I wonder also if work on that issue could also involve in skimming our components generally for other places where simple choices would support localization?  _Does this feel like premature optimization or useful to folks_? For example, I noticed yesterday that #144 should probably be more flexible (so that the consumer can select the order of MM/DD/YYYY , DD/MM/YYYY, or YY/MM/DD for inputs).  

@tinyels do you have any existing documentation of what you all have been talking about in terms of recommendations about how to approach localization? I wonder if this should be a section in our Engineering docs 🤔 .  I'd be interested to work on something like this with a bit of guidance.

PR #217, opened 2020-06-01
feat(charactercount): creates the character count form component

### LINKED ISSUE #169: New component: Forms / Character count
# [Character Count](https://designsystem.digital.gov/components/form-controls/#character-count)
Character count helps users know how much text they can enter when there is a limit on the number of characters.

## Proposed Requirements

```
Props = {
  id: string,
  name: string
  maxLength: number
  defaultValue?: string
  isTextArea?: boolean
  getCharacterCount?: (text: string) => number
  getCharacterCountMessage? : (remainingCount: number) => string
  inputRef: InputRef
} 
```
- Supports both <`input type="text" />` and `<textarea>`
- Character limit is determined by the `maxLength`.  It does not limit the actual input field length, allowing the user to exceed the soft cap.
- Input is invalid when `maxLength` is exceeded
- `getCharacterCount` prop allows custom logic for character count (that accounts for emojis, ligatures, etc).  It defaults to a function that uses `.length`
- `getMessage` returns custom hint messages to display. It defaults to a function that returns English messages.
- Component state tracks the input text length and validity. The consumer can pass in props that calculate the character count and determine which messages to display.

## Other Considerations
- Account for no-JavaScript environments. Add a default message in the message element that refers to the character limit. This will appear in instances when JavaScript does not load. 
- Instead of using `<Label />` inside this component use [render props](https://reactjs.org/docs/render-props.html) as per https://github.com/trussworks/react-uswds/pull/217#pullrequestreview-423722119

duncan-truss:
This creates a new component CharacterCount as part of the `forms/` group.  The requirements are laid out in the [issue](#169) so I won't recopy them all here.

https://designsystem.digital.gov/components/form-controls/#character-count
https://github.com/uswds/uswds/blob/develop/src/js/components/character-count.js

Please give me feedback on the use of state/hooks as I haven't done those before.  Similarly if you feel strongly about how template strings are handled which we discussed a bit on Slack and @haworku's [previously closed issue](https://github.com/trussworks/react-uswds/issues/124) may be worth revisiting.

I also noticed that we (and uswds) aren't using the validity api anywhere else.  Is this because there will usually be a form library used with their own validation functionality?

I had to add type definitions for an inputRef for TextInput and TextArea to make it easier to cast from the prop to the specific type needed for each component.

I did modify it to work with a defaultValue which seems important because we only listen for the onChange event to update the message.

fix #169

### DIFF
--- a/src/components/forms/CharacterCount/CharacterCount.tsx
+++ b/src/components/forms/CharacterCount/CharacterCount.tsx
@@ -42,7 +42,6 @@ export const CharacterCount = (
   const maxNum = maxLength === undefined ? 0 : maxLength
 
   /* Ideally defined as i18n translation strings */
-  const messageContent = `You can enter up to ${maxLength} characters`
   const emptyMessageFormat = `${maxLength} characters allowed`
   const remainingPluralFormat = '$0 characters left'
   const remainingSingularFormat = '$0 character left'
@@ -148,7 +147,6 @@ export const CharacterCount = (
           {label}
           {inputComponent}
         </FormGroup>
-        <noscript>{messageContent}</noscript>
         <span id={messageId} className={messageClasses} aria-live="polite">
           {state.limitMessage}
         </span>
--- a/src/components/forms/CharacterCount/CharacterCount.tsx
+++ b/src/components/forms/CharacterCount/CharacterCount.tsx
@@ -1,0 +1,160 @@
+import React, { useState } from 'react'
+import classnames from 'classnames'
+
+import { TextInput, TextInputRef } from '../TextInput/TextInput'
+import { Textarea, TextareaRef } from '../Textarea/Textarea'
+import { FormGroup } from '../FormGroup/FormGroup'
+
+interface CharacterCountProps {
+  id: string
+  name: string
+  maxLength: number
+  className?: string
+  label?: React.ReactNode
+  isTextArea?: boolean
+  textRef?: // currently getting an error when adding this as ref
+  | string
+    | ((instance: HTMLInputElement | null) => void)
+    | ((instance: HTMLTextAreaElement | null) => void)
+    | React.RefObject<HTMLInputElement>
+    | React.RefObject<HTMLTextAreaElement>
+    | null
+    | undefined
+}
+
+export const CharacterCount = (
+  props: CharacterCountProps &
+    React.InputHTMLAttributes<HTMLInputElement> &
+    React.TextareaHTMLAttributes<HTMLTextAreaElement>
+): React.ReactElement => {
+  const {
+    id,
+    name,
+    className,
+    label,
+    maxLength,
+    isTextArea,
+    textRef,
+    type, // Conflicts with type attribute already set on TextInput component
+    ...inputProps
+  } = props
+
+  const maxNum = maxLength === undefined ? 0 : maxLength
+
+  /* Ideally defined as i18n translation strings */
+  const messageContent = `You can enter up to ${maxLength} characters`
+  const emptyMessageFormat = `${maxLength} characters allowed`
+  const remainingPluralFormat = '$0 characters left'
+  const remainingSingularFormat = '$0 character left'
+  const overSingularFormat = '$0 character over limit'
+  const overPluralFormat = '$0 characters over limit'
+
+  const limitMessage = (len: number): string => {
+    let limitMessage
+    if (len === 0) {
+      return emptyMessageFormat
+    } else if (len <= maxNum) {
+      switch (maxNum - len) {
+        case 1:
+          return remainingSingularFormat.replace('$0', '1')
+        default:
+          return remainingPluralFormat.replace('$0', (maxNum - len).toString())
+      }
+    } else {
+      switch (len - maxNum) {
+        case 1:
+          return overSingularFormat.replace('$0', '1')
+        default:
+          return overPluralFormat.replace('$0', (len - maxNum).toString())
+      }
+    }
+  }
+
+  const [state, setState] = useState({
+    characterLength: 0,
+    limitMessage: emptyMessageFormat,
+    invalidMessage: false,
+  })
+
+  const classes = classnames('usa-character-count__field', className)
+  const messageClasses = classnames(
+    'usa-hint',
+    'usa-character-count__message',
+    { 'usa-character-count__message--invalid': state.invalidMessage }
+  )
+
+  const messageId = id + '-info'
+
+  const handleChange = (
+    e: React.ChangeEvent<HTMLInputElement> &
+      React.ChangeEvent<HTMLTextAreaElement>
+  ): void => {
+    const len = e.target.value.length
+    setState({
+      characterLength: len,
+      limitMessage: limitMessage(len),
+      invalidMessage: len > maxLength,
+    })
+  }
+
+  const handleBlur = (
+    e: React.FocusEvent<HTMLInputElement> &
+      React.FocusEvent<HTMLTextAreaElement>
+  ): void => {
+    const validationMessage = state.invalidMessage
+      ? 'The content is too long.'
+      : ''
+    e.target.setCustomValidity(validationMessage)
+  }
+
+  let inputComponent
+  if (isTextArea) {
+    const ref = textRef as TextareaRef
+    // How will we know if the label or hint has an id value should we add it to the describedby?
+    inputComponent = (
+      <Textarea
+        id={id}
+        data-testid="characterCountInput"
+        name={name}
+        className={classes}
+        onChange={handleChange}
+        onBlur={handleBlur}
+        aria-describedby={messageId}
+        inputRef={ref}
+        {...inputProps}></Textarea>
+    )
+  } else {
+    const ref = textRef as TextInputRef
+    inputComponent = (
+      <TextInput
+        id={id}
+        data-testid="characterCountInput"
+        name={name}
+        type="text"
+        className={classes}
+        onChange={handleChange}
+        onBlur={handleBlur}
+        inputRef={ref}
+        {...inputProps}
+      />
+    )
+  }
+
+  return (
+    <>
+      {/* Moves maxlength attribute off of input so it is not a hard entry limit */}
+      <div className="usa-character-count" data-maxlength={maxLength}>
+        <FormGroup>
+          {label}
+          {inputComponent}
+        </FormGroup>
+        <noscript>{messageContent}</noscript>
+        <span id={messageId} className={messageClasses} aria-live="polite">
+          {state.limitMessage}
+        </span>
+      </div>
+    </>
+  )
+}
+
+export default CharacterCount
--- a/src/components/forms/CharacterCount/CharacterCount.tsx
+++ b/src/components/forms/CharacterCount/CharacterCount.tsx
@@ -33,12 +33,15 @@ export const CharacterCount = (
     className,
     label,
     maxLength,
+    defaultValue,
     isTextArea,
     textRef,
     type, // Conflicts with type attribute already set on TextInput component
     ...inputProps
   } = props
 
+  const defaultLength =
+    defaultValue === undefined ? 0 : defaultValue.toString().length
   const maxNum = maxLength === undefined ? 0 : maxLength
 
   /* Ideally defined as i18n translation strings */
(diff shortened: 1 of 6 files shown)


trussworks-infra:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against ea0e04e3d5ba162cf3b2853f85820d8ae82b5d51
</p>

src/components/forms/CharacterCount/CharacterCount.tsx
```diff
@@ -0,0 +1,165 @@
...
+    id,
+    name,
+    className,
+    label,
+    maxLength,
+    defaultValue,
+    isTextArea,
+    textRef,
+    type, // Conflicts with type attribute already set on TextInput component
+    ...inputProps
```
eamahanna, line 40: I'm not sure it makes sense to lump the Input and TextArea attributes. I think they are distinct and should be applied to the textArea or Input respectively.

src/components/forms/CharacterCount/CharacterCount.tsx
```diff
@@ -0,0 +1,165 @@
...
+        inputRef={ref}
+        {...inputProps}
+      />
+    )
+  }
+
+  return (
+    <>
+      {/* Moves maxlength attribute off of input so it is not a hard entry limit */}
+      <div className="usa-character-count" data-maxlength={maxLength}>
```
eamahanna, line 152: You don't need to include data-maxLength in the rendered component because this isn't an actual HTML attribute.

src/components/forms/CharacterCount/CharacterCount.tsx
```diff
@@ -0,0 +1,165 @@
...
+  let inputComponent
+  if (isTextArea) {
+    const ref = textRef as TextareaRef
+    // How will we know if the label or hint has an id value should we add it to the describedby?
+    inputComponent = (
+      <Textarea
+        id={id}
+        data-testid="characterCountInput"
+        name={name}
+        className={classes}
```
eamahanna, line 121: the maxLength attribute should be added to each of the input component types: 
`maxLength={maxLength}`

eamahanna: See https://designsystem.digital.gov/components/form-controls/#character-count in the implementation section.

duncan-truss: If we do that then the field will not allow input over `maxlength` characters.  Their javascript code reads that attribute, removes it then adds it as the `data-maxlength` to the character-count container in the uswds code.

src/components/forms/CharacterCount/CharacterCount.tsx
duncan-truss: I can remove it, was just mirroring their implementation.

src/components/forms/CharacterCount/CharacterCount.tsx
```diff
@@ -0,0 +1,165 @@
...
+  const {
+    id,
+    name,
+    className,
+    label,
+    maxLength,
+    defaultValue,
+    isTextArea,
+    textRef,
+    type, // Conflicts with type attribute already set on TextInput component
```
haworku, line 39: Another way to deal with this is just in order of props.
```
<TextInput
        id={id}
        data-testid="characterCountInput"
        name={name}
        className={classes}
        defaultValue={defaultValue}
        onChange={handleChange}
        onBlur={handleBlur}
        aria-describedby={messageId}
        inputRef={ref}
        {...inputProps}
        type="text"
      />
```

We could also write a test to make sure when you pass in a different type it doesn't apply (and document the expected behavior that way rather than through a comment).

src/components/forms/CharacterCount/CharacterCount.tsx
```diff
@@ -0,0 +1,165 @@
...
+    ...inputProps
+  } = props
+
+  const defaultLength =
+    defaultValue === undefined ? 0 : defaultValue.toString().length
+  const maxNum = maxLength === undefined ? 0 : maxLength
+
+  /* Ideally defined as i18n translation strings */
+  const emptyMessageFormat = `${maxLength} characters allowed`
```
haworku, line 48: I think we agreed to pass lang strings as props until we have a bigger discussion. Suggestion - the lang could be a less props if we ignore singular v. plural for now- `emptyText`, `remainingText`, `overLimitText`. We can write the english defaults as "character(s) over limit" for now.

src/components/forms/CharacterCount/CharacterCount.tsx
```diff
@@ -0,0 +1,165 @@
...
+import { FormGroup } from '../FormGroup/FormGroup'
+
+interface CharacterCountProps {
+  id: string
+  name: string
+  maxLength: number
+  className?: string
+  label?: React.ReactNode
+  isTextArea?: boolean
+  textRef?: // currently getting an error when adding this as ref
```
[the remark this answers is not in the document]
duncan-truss: 🤦 Whoops that's why I made the interfaces then I didn't update the props

src/components/forms/CharacterCount/CharacterCount.tsx
```diff
@@ -0,0 +1,165 @@
...
+  const handleChange = (
+    e: React.ChangeEvent<HTMLInputElement> &
+      React.ChangeEvent<HTMLTextAreaElement>
+  ): void => {
+    const len = e.target.value.length
+    setState({
+      characterLength: len,
+      limitMessage: limitMessage(len),
+      invalidMessage: len > maxLength,
+    })
```
haworku, line 99: A few nitpicks here:
- Is there a need to track the `characterLength` in state? 
- Suggestion -  rename `invalidMessage` to make it clearer, maybe`isOverLimit`
- Suggestion -  spread state first to prevent bugs when making all of state is  one object
``` 
setState((state) => ({
      ...state,
      limitMessage: limitMessage(len),
      invalidMessage: len > maxLength,
    }))
```

src/components/forms/CharacterCount/CharacterCount.tsx
suzubara: the value of `type` should be whatever the user passed in via `inputProps`.. why does this need to be explicitly passed in here?

src/components/forms/CharacterCount/CharacterCount.tsx
suzubara: I don't think it's too bad... the thing is we'd need to figure out some conditionally-required props, ie. require `textareaProps` if it's a textarea and `textInputProps` if not. I'm not sure if this is possible with TS but I think treating them as the same is okay, and TS checking should catch issues if you tried to pass input-only props to a textarea and vice-versa. An alternative could be to create separate `CharacterCountTextInput` and `CharacterCountTextarea` components 🤷‍♀️

src/components/forms/CharacterCount/CharacterCount.tsx
suzubara: thinking long term for components like this that render complex strings (of which I doubt there will be many in this lib), this is where I think it would be more useful to expose a function prop that gets called with the current `len` value and returns the whole message string, so the user can have complete control over the logic and the copy. and the default value of this prop can be the existing `limitMessage` fn with the default English strings encapsulated in the fn.

src/components/forms/CharacterCount/CharacterCount.tsx
```diff
@@ -0,0 +1,165 @@
...
+  const classes = classnames('usa-character-count__field', className)
+  const messageClasses = classnames(
+    'usa-hint',
+    'usa-character-count__message',
+    { 'usa-character-count__message--invalid': state.invalidMessage }
+  )
+
+  const messageId = id + '-info'
+
+  const handleChange = (
```
suzubara, line 90: both `handleChange` and `handleBlur` must implement `onChange, onBlur` props that may be passed in as part of `inputProps` in order to preserve any custom event handlers the user may pass in.

src/components/forms/CharacterCount/CharacterCount.tsx
```diff
@@ -0,0 +1,165 @@
...
+    const validationMessage = state.invalidMessage
+      ? 'The content is too long.'
+      : ''
+    e.target.setCustomValidity(validationMessage)
+  }
+
+  let inputComponent
+  if (isTextArea) {
+    const ref = textRef as TextareaRef
+    // How will we know if the label or hint has an id value should we add it to the describedby?
```
suzubara, line 115: according to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-describedby_attribute `aria-describedby` is a space-separated list of element IDs. so IMO let's treat this like `className`, where we append the message ID to the user-defined `aria-describedby` which can be passed in as part of `inputProps`

src/components/forms/CharacterCount/CharacterCount.tsx
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  const handleBlur = (
+    e: React.FocusEvent<HTMLInputElement> &
+      React.FocusEvent<HTMLTextAreaElement>
+  ): void => {
+    const validationMessage = state.invalidMessage
+      ? 'The content is too long.'
+      : ''
+    e.target.setCustomValidity(validationMessage)
```
suzubara, line 109: this is not a blocker, but also just thinking long-term to support whatever validation method the user has, we'll probably want to expose a flag prop that enables/disables native validation and lets them pass in their own invalid state somehow

src/components/forms/CharacterCount/CharacterCount.tsx
haworku: @suzubara did you see this comment? https://github.com/trussworks/react-uswds/pull/217#pullrequestreview-423042971 seems like we are on the same page.  Can we just do that now?  I think that's more useful.

src/components/forms/CharacterCount/CharacterCount.tsx
duncan-truss: Agreed, good points! Yep I did read about setState not being a merge operation in this case.

src/components/forms/CharacterCount/CharacterCount.tsx
suzubara: ohh got it. So, yes I think we agree, my only suggestion is that there be two props, one that returns the message copy (ie, `message`), and one that indicates validity (ie, `isValid` or `hasError`) (see my note about disable native validation). It is very likely that a user who is doing a custom implementation for copy will use the same code on their end to feed both props, but I think the component API will be easier to work with if they are decoupled. To me, a prop called `validateLength` sounds like an imperative function that might trigger validation, when really it should just return the message string.

src/components/forms/CharacterCount/CharacterCount.tsx
ahobson: I am but an 🥚 but since they are need mutually exclusive props (and are separate html tags), I would vote for separate components.

duncan-truss: In my opinion I don't think the implementation is different enough to separate the components.  Textarea just happens to have some quirky attributes that textinput doesn't like rows, columns, wrap behavior, spellcheck ... USWDS seems to ignore rows and column attributes anyway at least within a default form group.  Honestly there are more `<input>` types besides `type="text"` to probably consider supporting or that would diverge more than textarea does.

duncan-truss:
> What if we use a render prop to render the input, so the user doesn't need to pass in a label/hint/etc.? I'd like to avoid re-rendering a `<Label>` (and other possible related components) in `CharacterCount`. More info at https://reactjs.org/docs/render-props.html but I'm imagining something like:
> 
> ```
> <CharacterCount
>   render={(input) => <FormGroup>My custom thing {input}</FormGroup>}
>   />
> ```

I think I'm understanding your suggestion.

> > I also noticed that we (and uswds) aren't using the validity api anywhere else. Is this because there will usually be a form library used with their own validation functionality?
> 
> Yes, I think validation (beyond what html5 does) should be happening outside our components, as you suggested, through form libraries or validation libraries. There's so much about validation that is specific to the use case. Even in our little example, there could be bugs if we hard code this too much - for example, the`defaultLength` calculation probably would have to happen differently if the text included things like emojis.
> 
> I wonder if this component instead needs a prop `validateLength`. This would be a function called whenever the input changes and return a string with the validation message for the character count. It would take in parameters of the current input value, the max length, and (for our specific case/story) do the calculations in `limitMessage`. That would also remove the need to individually pass down the different lang strings to the component as props.
> 
> Def interested in other ideas here. The problem I see is that the USWDS js implementation seems opinionated and our version will probably have to be more flexible.

Ahhhh so true about unicode 💩 and maybe ligatures too 😩

haworku:
Closing this for now it can be picked up when someone takes on this issue again.

Issue #187, opened 2020-05-20
Proposal: start using enums instead of booleans for prop variants


haworku:
We have several components such as `Header`, `Footer`,  `Button`, and maybe even aspects of `Grid`, with  boolean props that can conflict. 

For example, you should not have the `big` and `medium` prop applied to the same footer. Some of this should be on the developer to use our props ... properly 😝... AND we also could starting using variants as well when helpful. 

Instead of booleans we could move towards a single prop named something like `size` or `type` that accepts one of these types. The types would be enums. How do folks feel about this? Is this something we could start code reviewing for and document as practice? 

Proposed requirements to close this issue:
- Add a type variant to the `Button` component (keep the old way of doing these props too but add a deprecation warning - it can be removed fully in next major release ).
- Write out additional issues for other components with this problem.
- Add description of best practices with props to #73

suzubara:
agreed this is the way to go!

haworku:
For requirements,  lets make reword this ticket to be about adding the variant to the `Button` component to start.

PR #229, opened 2020-06-04
feat(button): size enum on Button component

ahobson:
Provide a size enum to eventually replace the small and big boolean props on Button

re #187


trussworks-infra:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 589ec5e88904d962d48ef255b0fb7cca67c8ec92
</p>

src/components/Button/Button.tsx
```diff
@@ -36,6 +44,14 @@ export const Button = (
...
   } = props
 
+  let isBig = big || false
+  let isSmall = small || false
+  if (size === 'big') {
+    isBig = true
+  }
+  if (size === 'small') {
+    isSmall = true
+  }
```
haworku, line 54: I think another way to do this is  
``` 
const isBig = size === 'big' || (!size && big)
const isSmall = size === 'small' || (!size && small)
```

haworku: That's assuming we want to ignore the deprecated props if they pass down `size`. @suzubara I'm blanking on what we decided here... I know we said we wouldnt fix if they passed down `big small` together but do we want to guard against `size='big' small` ?

ahobson: Also, I very much like your suggestion, but I have a counter suggestion.  This is slightly into code golf territory, but how would you feel about

```typescript
  const isBig = big ? big : size === 'big'
  const isSmall = small ? small : size === 'small'
```

It changes things to prefer the old `big` and `small` in case of both, but that feels appropriate to me

ahobson: Or to prefer `size`
```typescript
  const isBig = size ? size === 'big' : big
  const isSmall = size ? size === 'small' : small
```

suzubara: I like the last option (that prefers the `size` prop). I don't think we need to deal with guarding against both `big small`, that is their choice and will be deprecated 🤷‍♀️

ahobson:
> Let's also update all of `Button.stories.tsx` to make sure we only show examples with size (and none with deprecated props).
> 
> 🤔 Is it worth also adding tests for "ignore deprecated big/small props if size exist"? In other words if someone uses `size` that's what we follow first (even if big/small passed down).

I've added a commit that does both of these

haworku:
updated PR title (since it generates our changelog) to note that this work also adds `deprecationWarning` util

Issue #161, opened 2020-05-18
Button should take a Component prop


suzubara:
Right now, the Button component will only render a `button`. However it's very common for a user to want to implement button styles on an `<a>` element or even a custom link component (such as if they are using client-side routing). So the Button component should be modified to accept an optional `Component` prop (probably of type `ReactNode`) to support that use case. It should still default to `button` if no `Component` prop is defined.

haworku:
@suzubara Are you satisfied with the work for #309 to cover this ticket? P

Personally I could still see us implementing a similar solution on `Button` as well. Or else exporting styles from Button to be used in Link directly with Button-like props (though that creates dependency  between the two, since props on Button can change).

suzubara:
@haworku Yes I think this can be closed! IMO we could abstract the Button styles/classes out so they are more easily shared between the Button & Link components, but that would probably be a new issue.

PR #267, opened 2020-06-18
fix: Use JSX.ImplicitElements to derive valid property names

ahobson:
# Summary

Use the proper react types to indicate which props are allowed on components.

## Related Issues or PRs
This was found as part of exploration for #161

## How To Test
This should not change functionality other than allowing the proper attributes

### DIFF
--- a/src/components/Tag/Tag.tsx
+++ b/src/components/Tag/Tag.tsx
@@ -7,9 +7,9 @@ interface TagProps {
 }
 
 export const Tag = (
-  props: TagProps & React.HTMLAttributes<HTMLSpanElement>
+  props: TagProps & JSX.IntrinsicElements['span']
 ): React.ReactElement => {
-  const { children, background, className } = props
+  const { children, background, className, ...spanProps } = props
 
   const style: React.CSSProperties = {}
   if (background) {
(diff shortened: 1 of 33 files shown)


src/components/header/MegaMenu/MegaMenu.tsx
```diff
@@ -19,7 +19,7 @@ export const MegaMenu = (
       <div className="grid-row grid-gap-4">
         {items.map((listItems, i) => (
           <div className="usa-col" key={`subnav_col_${i}`}>
-            <NavList items={listItems} type="megamenu" {...ulProps} />
+            <NavList items={listItems} megamenu={true} {...navListProps} />
```
haworku, line 22: suggestion:  for `megamenu={true}` and I think I saw this one other place.  Leave off the attribute assignment when its `true.` Just by including the name of the attribute on the element it is by default true.

src/components/Tag/Tag.tsx
```diff
@@ -19,7 +19,11 @@ export const Tag = (
   const tagClasses = classnames('usa-tag', className)
 
   return (
-    <span data-testid="tag" className={tagClasses} style={{ ...style }}>
+    <span
+      data-testid="tag"
+      className={tagClasses}
+      style={{ ...style }}
+      {...spanProps}>
```
haworku, line 26: Thanks for catching this, I think I saw a couple other places where this rewrite was kind of a cleanup for us.

ahobson:
> Looks good to me! Just need to take [WIP] out of the PR title and make it match the semantic changelog format.

Ok!  Should I create a corresponding issue about it?  Is this a bugfix or a feature?

haworku:
@ahobson I think `fix` is fine with the same PR title you have.  Also don't feel a need to make an issue since this came up in #161 somewhat organically.

Issue #191, opened 2020-05-21
[bug] uswds typography mixin not being applied in storybook


haworku:
**Describe the bug**
The `:hover` styles on primary links in Footer(#78) do not match styles applied in the [USWDS design system docs](https://designsystem.digital.gov/components/footer/). Specifically the blue color that comes from the [`typeset-link` ](https://github.com/uswds/uswds/blob/8729a83824a1eeb31027c5caae0b7bce799376e0/src/stylesheets/core/mixins/_typography.scss#L63-L65) mixin is not being applied in our code.

**Expected behavior**
We want our storybook examples to look like the examples on uswds design system docs.  In this case, all links should be blue when hovered. 

**Additional context**

- Look more into how [`$theme-global-link-styles` ](https://github.com/uswds/uswds/blob/c5715969bccd6c33e372d1732ee6c3eef0b8c770/src/stylesheets/theme/_uswds-theme-typography.scss#L69)is used we probably need to turn this on in our repo. 
- Also, revisit how we bring in uswds styles our storybook config. We should probably be importing `.scss` instead of instead of as compiled `.css`.  
- @suzubara knows more about all of this

haworku:
The root cause of this is we weren't bringing in all uswds modules into storybook and settings things up to be customized. Any settings variables we set in storybook were  not working for that reason. 

To fix, we need to first  make storybook bring in all uswds styles and then turn on  `$theme-global-link-styles`.

haworku:
Ran into an issue with autoprefixing with postcass in storybook after bringing in all the uswds`.scss` in #696.  It described [in this comment ](https://github.com/trussworks/react-uswds/pull/696#issuecomment-749113191).  Since we've already invested considerable time, the solution fails due to an an unrelated bug with storybook, and the bug does not impact the library code (just the storybook styles), we decided to close for now as **won't fix**. When Storybook 7x comes out we can revisit.

PR #285, opened 2020-06-26
[WIP] Link component with cloneElement

### LINKED ISSUE #145: [feat] Link Component
# <Link />

Make a basic `Link` component that wraps a link in the classes we need for an internal link `usa-link`  versus and external link `usa-link usa-link--external`.  

This is another way to encapsulate styles. See  Typography > [Links](https://designsystem.digital.gov/components/typography/#links).

By default renders an anchor tag, but has an optional prop to accept a different component.

haworku:
# Summary

- Adds new Link component 
- Uses `PossibleLinkProps` union type.
-  `variant` prop adjusts styles.  Style as button by using `variant: 'unstyled'` and passing in custom classes 

### Random learnings about types
- Typescript doesn't seem to have a straightforward XOR type.  [Union types are not exclusive](https://stackoverflow.com/questions/50346958/unexpected-behaviour-using-index-type-with-union-type-signature/50355289#50355289)
- [Overloads](https://www.typescriptlang.org/docs/handbook/functions.html#overloads) seem to mostly address the issue (errors when expected, though messages can be strange).  

### Note on broken "visited" storybook example
The "visited" link story example isn't working.  I *think* think is related to #191. Styles aren't being inherited from custom stylesheets. 

## Related Issues or PRs
- closes #145 
- #161

## How To Test

- See new Link component stories


haworku:
Hey I've pushed a new commit of how this could work using `cloneElement` instead of `createElement`.  This seems like a more straightforward way to simply add the styles (props) we want for link

haworku:
Closing in favor of #309

PR #309, opened 2020-07-07
feat: Link Component #161

ahobson:
# Summary
 
Adds new Link component. Includes handling for `asCustom` prop.

## Related Issues or PRs

relates to #161 

## How To Test

- check new Link examples on Storybook


src/components/Link/Link.tsx
```diff
@@ -0,0 +1,77 @@
...
+import classnames from 'classnames'
+
+type StyledLinkProps = {
+  variant?: 'external' | 'unstyled'
+  className?: string
+}
+
+type LinkProps = {
+  href: string
+  children: React.ReactNode
```
haworku, line 11: Added these back.  Trying to follow the pattern of [requiring props that are fundamental to the element](https://github.com/trussworks/react-uswds/blob/develop/docs/adding_new_components.md#props).  

Without these props `<Link className="usa-button" variant="unstyled" />` is valid  code. It compiles without warning ... and a regular anchor tag in our app would warn about this for missing `href`  and `children` (since we use jsx-a11y to check valid anchor tags).

src/components/Link/Link.tsx
```diff
@@ -0,0 +1,77 @@
...
+    : classnames(
+        'usa-link',
+        {
+          'usa-link--external': isExternalLink,
+        },
+        className
+      )
+}
+
+export function Link<FCProps = LinkProps>(
```
haworku, line 46: I think we still need [overloads](https://mariusschulz.com/blog/function-overloads-in-typescript#version-4-function-overloads) to get accurate intellisense and type suggestions. What I notice is right now when the `asCustom` is not present  and `Link` has invalid props, it only warns about not including `asCustom`, maybe because it has the broadest types. It doesn't warn for anchor tag. 

Here's what an overload before line 46 could look (I think): 

```
// Overloads
export function Link<T>(asCustomProps: AsCustomProps<T>): React.ReactElement
export function Link(linkProps: LinkProps): React.ReactElement
```

 Lmk if you want to pair on this or if I'm off base.

src/components/Link/Link.stories.tsx
```diff
@@ -0,0 +1,84 @@
...
+  },
+}
+
+export const Default = (): React.ReactElement => (
+  <p>
+    <Link href={'#'}> This</Link> is a text link.
+  </p>
+)
+
+export const Visited = (): React.ReactElement => (
```
haworku, line 19: Let's just remove the Visited story all together and just comment TODO for now-  since it doesn't work.  My guess is the styling issue is related to custom style themes -  #191

src/components/Link/Link.tsx
ahobson: Thank you for this suggestion.  I'll have to look into this.  I think I will have time on Friday

ahobson: Thank you!  I think I've simplified the type signatures and added the overloads.  It seems to do what we expect in VSCode now.

src/components/Link/Link.stories.tsx
ahobson: I've commented it out

src/components/Link/Link.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+export const Default = (): React.ReactElement => (
+  <p>
+    <Link href={'#'}> This</Link> is a text link.
+  </p>
+)
+
+/*
+ * TODO: sort out how to test visited links in storybook
+ *
+ * export const Visited = (): React.ReactElement => (
```
suzubara, line 22: Looking at the USWDS source (https://designsystem.digital.gov/components/typography/#links) I think they just added the `usa-color-text-visited` class to their docs theme for demonstration purposes. We probably could do the same in a Storybook-only CSS file (these can be added in `./storybook`):

```
.usa-color-text-visited {
  color: color($theme-link-visited-color);
}
```

suzubara: we may also want to do something similar to the `Default` story so that it _never_ gets the `:visited` style applied. IMO adding CSS to Storybook to enforce the intended styles for demonstration purposes is a 💯   valid thing to do.

src/components/Link/Link.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+    <Link variant="external" href={'http://www.truss.works'}>
+      external website
+    </Link>
+    .
+  </p>
+)
+
+export const StyledAsButton = (): React.ReactElement => (
+  <p>
+    <Link className="usa-button" variant="unstyled" href={'#'}>
```
suzubara, line 51: suggestion for future iteration: I have a feeling the `usa-button` and all of its variations will be a common use-case for `Link` components, so adding the possible button style/className props to the `Link` component might be a good idea.

src/components/Link/Link.stories.tsx
haworku: @suzubara does https://github.com/trussworks/react-uswds/pull/309/commits/813c393bbe7471bd0e0b0b1b8e4b3a1b1f4093df workable for this??

suzubara: yeah that looks great!

Issue #216, opened 2020-06-01
Move to a single branch release flow on "main" branch


suzubara:
As @haworku and @eamahanna witnessed, our current release flow is flawed:
- Merging the release branch into `master` requires that we also have to keep `develop` up-to-date with the release branch changes. This means an extra PR that must be approved
- Since we've transitioned to using squash & merge on this repo exclusively, the workflow of `develop -> release branch -> master -> develop` breaks with a linear commit history, and also results in a loss of commit history since release PRs are squashed on master

My suggestion is the following:
- Deprecate the current `master` branch and use only one main branch for active development and stable versions (naming can be TBD)
- Stable versions will always be accessible via their associated tag.
  - Version tags should be checked out if the code at a specified version needs to be examined, or when making maintenance fixes.
- The new release flow will be: `main branch -> create release branch, update version/changelog -> approve release branch PR -> merge release into main branch -> tag and publish`

Pros:
- Removes extra step of merge into `master` and also into `develop`
- Removes potential confusion of which branch a release PR should be based off
- Keeps a cleaner commit history using squash & merge (the release branch merges will always just reflect the `chore: release` commit, which should always only be the new version number and changelog updates)
- Release hotfixes should be made directly off of `develop` (and merged back into it)

Cons:
- From the time of creating the release branch, all other merges to `develop` should be prevented until the release has been completed. Otherwise we risk publishing a release that has unaccounted-for changes.
  - Mitigation: I would argue that right now, this project is small enough that we can mitigate that from happening with good team communication. Release PRs should take priority over any other PR reviews, and anyone issuing a new release can communicate via Slack that no PRs should be merged until it's completed.
  - Mitigation: _if_ new work _does_ get merged in while a release PR is open, it can be resolved by requiring `release-*` branches always require being up to date with their base branch before being merged (this is a Github branch protection setting)

eamahanna:
I think this makes sense given it reduces the overall complexity with releases. I would just want to make sure the release documentation reflects that you need to make a PSA to the react-uswds channel when doing a release (just to make sure this doesn't get overlooked).

duncan-truss:
I've seen a git workflow where we'd merge the release branch back into develop instead of closing it and merging master back into develop.  Would that simplify anything?

suzubara:
@duncan-truss yes that was my suggestion

duncan-truss:
[...]

PR #314, opened 2020-07-13
ci(main branch): use single branch, remove references to develop or master

haworku:
# Summary

Move to single branch (`main`). Adjust docs and ci accordingly. 
## Related Issues or PRs

 #216 

## How To Test

- all PR checks should still run, but against main branch


.github/workflows/deploy-storybook.yml
```diff
@@ -3,7 +3,7 @@ name: Deploy Storybook
 on:
   push:
     branches:
-      - master
+      - main
```
haworku, line 6: Hmmm... what should we do about this? Random ideas
- we make this workflow instead run when a new tag is created on GH 
- we make a `yarn publish` that pulls down latest main, runs npm publish, and then runs storybook

suzubara: hmmmm good question... I think I prefer the publish happening in GH, and it looks like we can trigger it to run when a tag is pushed. what if we filter on tags like this to capture all v1 (and we can add `2, 3,` etc down the line) - https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags
```
- 1.*
```

PR #345, opened 2020-07-23
fix: Export NavDropDownButton

cablej:
# Summary

It appears that NavDropDownButton cannot be used as it is not exported (following code samples from https://trussworks.github.io/react-uswds/?path=/story/header-navdropdownbutton--default-drop-down-menu). This PR exports NavDropDownButton so that it can be used.


haworku:
@cablej thanks for this! I'll make sure this gets merged and published in the next release.

Issue #80, opened 2020-04-15
New Component: ValidationChecklist


suzubara:
Add Validation component: https://designsystem.digital.gov/components/form-controls/#validation

- We can use/extend the existing `Alert` component
- Probably add a new component `<ValidationChecklist />`
- Validation checking will depend on what forms solution the application is using. For now I'd recommend this component accept a prop that indicates which validations have been met somehow.
- Stories should showcase validation working with react-uswds `Form`

PR #351, opened 2020-07-27
feat(Validation): Add new Validation components

haworku:
# Summary

Add `ValidationChecklist` and `ValidationItem` components. See https://designsystem.digital.gov/components/form-controls/#validation

## Related Issues or PRs
#80


example/src/pages/Forms.tsx
```diff
@@ -0,0 +1,100 @@
...
+                id="email"
+                value={values.email}
+              />
+
+              {touched.password && errors.password && (
+                <Alert type="info" validation heading="Password Requirements">
+                  <ValidationChecklist id="validate-password">
+                    <ValidationItem
+                      id="required"
+                      isValid={values.password.length > 0}>
```
haworku, line 71: Ideally `isValid` would use more complex logic than this - just did something simple for the example

suzubara: ahh I see, yeah I wouldn't worry too much about this since it's really a Formik implementation detail..

PR #381, opened 2020-08-04
build(webpack): support browsers and node 

### LINKED ISSUE #380: Components won't build in Gatsby SSR
**Describe the bug**
When using these components in Gatsby, they work in the develop stage but not when building for server-side rendering.

**To Reproduce**
1. Create new Gatsby site and `yarn add @trussworks/react-uswds`
2. Use any components in pages
3. `gatsby develop` works
4. `gatsby build` fails

**Expected behavior**
Initial issue is with use of `window` in index.js: 
```
ERROR #95312 
"window" is not available during server side rendering.
```

This can be fixed by using setting `gatsby-node.js` to use webpack null-loader for `/react-uswds/`, per instructions at https://www.gatsbyjs.org/docs/debugging-html-builds/#fixing-third-party-modules

Subsequent issue is harder to debug, errors when using any of these components and building.
```
  Error: Minified React error #130; visit https://reactjs.org/docs/error-decoder.html?invariant=130&args[]=undefined&args[]= for  the full message or use the non-minified dev environment for full errors and additional helpful warnings.  
```

Because there's not a lot of information in this error message, it's difficult to tell what's going on. And since there's no error when running develop, there's not a lot more to go on here.

**Additional context**
My example site in development is at https://github.com/spacedogXYZ/crush2020, if this is helpful to debug. Other React USWDS implementations do not appear to have this problem (eg, https://github.com/18F/federalist-uswds-gatsby) but this repo looks like it is currently under development and would be great if it worked for this use case.

**Environment:**
 - Mac OS Catalina 10.15.5
 - Node versions v13.8.0 and v12.18.3 (LTS)

ahobson:
# Summary

Update webpack config to [support umd build for both browsers and node](https://webpack.js.org/configuration/output/#outputglobalobject)
## Related Issues or PRs

closes #380

## How To Test
Create the file `ssr.js` in the root of the `react-uswds` project
```js
const react = require('react')
const domServer = require('react-dom/server')
const uswds = require('./lib')

const TagFactory = react.createFactory(uswds.Tag)
const app = domServer.renderToString(TagFactory())
console.log(app)
```

1. `yarn build`
1. `NODE_ENV=production node ssr.js`

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -12,6 +12,7 @@ module.exports = {
     filename: '[name].js',
     library: 'ReactUSWDS',
     libraryTarget: 'umd',
+    globalObject: 'this',
   },
   externals: {
     react: {


ahobson:
> LGTM - following testing instructions. @ahobson you also checked it against the scenario described in #380 correct?

I did.  I'll update that ticket with instructions after I merge

PR #383, opened 2020-08-05
feat(CharacterCount): Add new component CharacterCount

### LINKED ISSUE #169: New component: Forms / Character count

haworku:
# Summary

Add `CharacterCount` based on [uwds spec ](https://designsystem.digital.gov/components/form-controls/#character-count). See also discussion in #169 

## Related Issues or PRs

closes #169 

## How To Test

Check Forms/ Character Count in Storybook


src/components/forms/CharacterCount/CharacterCount.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+    info: `
+USWDS 2.0 Character count component
+
+Source: https://designsystem.digital.gov/components/form-controls/#character-count
+`,
+  },
+}
+
+export const textInput = (): React.ReactElement => (
+  <FormGroup>
```
ahobson, line 19: I think we want form component stories to be inside a `Form`, not just a `FormGroup`

ahobson:
> * write a test that `textRef` prop behaves as expected.  Having some trouble testing this  (specifically creating a ref that our current typescript definitions like).  Realized this is first time we are testing the ref prop on a input so our types could be wrong.

Is this testing what you had in mind?
```typescript
    it('supports inputRef', () => {
      const tRef = React.createRef<HTMLTextAreaElement>()
        <CharacterCount
          id="character-count"
          name="character-count"
          maxLength={10}
          isTextArea
          inputRef={tRef}
        />
      )
      expect(queryByTestId('textarea')).toBe(tRef.current)
    })
```

haworku:
> > * write a test that `textRef` prop behaves as expected.  Having some trouble testing this  (specifically creating a ref that our current typescript definitions like).  Realized this is first time we are testing the ref prop on a input so our types could be wrong.
> 
> Is this testing what you had in mind?
> 
> ```ts
>     it('supports inputRef', () => {
>       const tRef = React.createRef<HTMLTextAreaElement>()
>       const { queryByTestId } = render(
>         <CharacterCount
>           id="character-count"
>           name="character-count"
>           maxLength={10}
>           isTextArea
>           inputRef={tRef}
>         />
>       )
>       expect(queryByTestId('textarea')).toBe(tRef.current)
>     })
> ```

@ahobson Yes, something like that. Was using `useRef`.  However, either way I can't get those tests actually passing - get typescript errors. In the case of this test `Type '{ id: string; name: string; defaultValue: string; maxLength: number; ref: RefObject<HTMLInputElement>; }' is not assignable to type 'ClassAttributes<HTMLTextAreaElement>'. Let's pair on this.

PR #386, opened 2020-08-07
feat: add license finder

mdawn:
# Summary

Adds license finder as part of circleci check. Also converts jobs to workflows.

Little bit about the strategy: 
My first impulse was to cram Ruby & Node into a single container but [that's a Docker no-no](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#each-container-should-have-only-one-concern), so I converted the jobs to workflows. Am getting a Happo fail (was getting that on main branch as well) which [was mentioned in slack](https://trussworks.slack.com/archives/C0122MPMXL3/p1596637105486100), so not sure if that's something y'all want to work out or if infra reserve look at that.


ahobson:
It doesn't seem like CircleCI thinks the config file is valid

ahobson:
So ... it looks like the license finder does need node, so it looks like we need to create an image that has the license finder for each language we use.  Blearugh
```
LicenseFinder::Yarn: is not installed
```

mdawn:
> So ... it looks like the license finder does need node, so it looks like we need to create an image that has the license finder for each language we use. Blearugh
> 
> ```
> LicenseFinder::Yarn: is not installed
> ```

I TOLD YOU SO lolol

Will you pair with me later on this?

ahobson:
> > So ... it looks like the license finder does need node, so it looks like we need to create an image that has the license finder for each language we use. Blearugh
> > ```
> > LicenseFinder::Yarn: is not installed
> > ```
> 
> I TOLD YOU SO lolol
> 
> Will you pair with me later on this?

I can, but I think this intersects with the work Ren is doing to slim the circleci images.  This is just another tool that needs to be installed

ahobson:
> > > So ... it looks like the license finder does need node, so it looks like we need to create an image that has the license finder for each language we use. Blearugh
> > > ```
> > > LicenseFinder::Yarn: is not installed
> > > ```
> > 
> > 
> > I TOLD YOU SO lolol
> > Will you pair with me later on this?
> 
> I can, but I think this intersects with the work Ren is doing to slim the circleci images. This is just another tool that needs to be installed

Also: it seems pretty terrible that the license finder failed when it couldn't find yarn and yet the job passed

mdawn:
> 
> Also: it seems pretty terrible that the license finder failed when it couldn't find yarn and yet the job passed

Hard agree.

Another PR bites the dust! I'm closing this PR until we hash out what we want containerized vs. in workflows. Thanks for hopping in, Andrew. Looking forwards to the coming battle royale. 😉 🐋 ⚔️

PR #399, opened 2020-08-13
feat: add combo box component

### LINKED ISSUE #170: New component: Combo box
# [Combo Box](https://designsystem.digital.gov/components/form-controls/#combo-box)
A combo box helps users select an item from a large list of options.

## Proposed Requirements
```
Props = {
  id: string,
  name: string
  children: React.ReactNode
  options: [],
  onSelect: (option) => void,
  defaultValue? : string
  required?: boolean
} & JSX.IntrinsicElements['select']
``` 
- base element is `<select>`
- ability clear selections or already typed text
- See [uswds combo box js](https://github.com/uswds/uswds/blob/develop/src/js/components/combo-box.js)

jim:
# Add Combo Box

This PR is a first pass at adding the [combo box](https://designsystem.digital.gov/components/form-controls/#combo-box) component. We are merging to a feature branch for now since we anticipate refactoring the component.

The initial implement uses a prop for handling the options list and `setFieldValue` to control the input value.  

There are significant tests added, which will help be a guard us as future refactors are made. 

## Related Issues or PRs

related to #170 

## How To Test

1. Open storybook and verify that the behavior matches what is shown in the [USWDS version](https://designsystem.digital.gov/components/form-controls/#combo-box).


yarn.lock
```diff
@@ -14609,11 +14609,20 @@ use@^3.1.0:
   resolved "https://registry.yarnpkg.com/use/-/use-3.1.1.tgz#d50c8cac79a19fbc20f2911f56eb973f4e10070f"
   integrity sha512-cwESVXlO3url9YWlFW/TA9cshCEhtu7IKJ/p5soJ/gGpj7vbvFrAY/eIioQ6Dw23KjZhYgiIo8HOs1nQ2vr/oQ==
 
+<<<<<<< HEAD
```
jim, line 14612: This needs to be fixed.

package.json
```diff
@@ -64,7 +64,7 @@
     "happo-plugin-storybook": "^2.5.3",
     "react": "^16.10.0",
     "react-dom": "^16.10.0",
-    "uswds": "^2.7.0"
+    "uswds": "2.7.1"
```
haworku, line 67: Good catch

src/components/forms/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,284 @@
...
+        role="combobox"
+        ref={inputRef}
+        value={inputValue}
+        onBlur={handleInputBlur}
+        onClick={handleInputClick}
+        onKeyDown={handleInputKeyDown}
+        onChange={(e): void => setInputValue(e.target.value)}
+        // these do not type check:
+        // autocapitalize="off"
+        // autocomplete="off"
```
haworku, line 214: Try ```autoCapitalize="off"
         autoComplete="off"```

src/components/forms/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,284 @@
...
+              }}>
+              {option.label || option.value}
+            </li>
+          )
+        })}
+      </ul>
+
+      <div className="usa-combo-box__status usa-sr-only" role="status"></div>
+      <span id={assistiveHintID} className="usa-sr-only">
+        When autocomplete results are available use up and down arrows to review
```
haworku, line 276: We will need to surface this as a prop so our assistive text can be i18n ready.  We can set a default value of English if needed as per #124

src/components/forms/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,284 @@
...
+  None,
+  Input, // The textfield
+  Item, // One of the list items
+}
+
+interface ComboBoxProps {
+  id: string
+  name: string
+  className?: string
+  options: ComboBoxOption[]
```
haworku, line 31: I could be missing something but how would we load this component with an option already selected?  For example, of we are displaying form data coming back from an api. In my mind there has to be some kind way to set the selected item from outside the component (whether that's tracking selected state within the `ComboBoxOption`, or having a separate `selectedOption` prop, or using the current input value to select an option on initial load).

src/components/forms/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,284 @@
...
+      ref={containerRef}>
+      <select
+        className="usa-select usa-sr-only usa-combo-box__select"
+        name={name}
+        aria-hidden
+        tabIndex={-1}
+        value={selectedValue}
+        {...selectProps}>
+        {options.map((option) => (
+          <option key={option.value} value={option.value}>
```
haworku, line 190: The consumer has to make sure their option values are unique. Maybe we should rename value to`id` in `ComboBoxOption` to suggest this more in the naming.

src/components/forms/ComboBox/ComboBox.tsx
jim: Thanks! It's working now.

src/components/forms/ComboBox/ComboBox.tsx
jim: This was missing. I added support for passing in a `defaultValue` prop, which I think will satisfy this and I think is the way you'd do it with a standard `select` in React? There is a story that covers this case now, but not a unit test yet.

src/components/forms/ComboBox/ComboBox.tsx
jim: I made this a string prop (`assistiveHint`) since I don't think the message is dependent on any state (in the character count component we use a function and pass in the current value). Do you think that is OK, or should we always make this kind of thing a function prop for consistency?

jim:
There is a bug where the option list displays as empty if there is already a value selected. I think it stems from the filtering logic.

src/components/forms/ComboBox/ComboBox.tsx
haworku: Oh that makes sense!

src/components/forms/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,317 @@
...
+        setFocusMode(FocusMode.Input)
+      } else {
+        // eslint-disable-next-line security/detect-object-injection
+        const newOption = filteredOptions[newIndex]
+        setFocusedValue(newOption.id)
+      }
+    }
+  }
+
+  const handleInputKeyDown = (event: KeyboardEvent): void => {
```
haworku, line 123: Enter key should select to option

src/components/forms/ComboBox/ComboBox.tsx
```diff
@@ -215,8 +214,8 @@ export const ComboBox = (
         value={selectedValue}
         {...selectProps}>
         {options.map((option) => (
-          <option key={option.value} value={option.value}>
-            {option.label || option.value}
+          <option key={option.id} value={option.id}>
+            {option.label || option.id}
```
haworku, line 220: Should just be  `option.label` since label  is required now.  (same for all  other `option.label || option.id` cases)

src/components/forms/ComboBox/ComboBox.tsx
haworku: That works!  I don't think we should be using a function prop like that too often (hopefully) -  validations might be the main case.

src/components/forms/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,317 @@
...
+        id={selectID}
+        className="usa-combo-box__input"
+        type="text"
+        role="combobox"
+        ref={inputRef}
+        value={inputValue}
+        onBlur={handleInputBlur}
+        onClick={handleInputClick}
+        onKeyDown={handleInputKeyDown}
+        onChange={(e): void => setInputValue(e.target.value)}
```
haworku, line 240: _*I think*_ we want to allow the consumer to still be able to call their own `onChange` passed in from the `selectProps.onChange`

yarn.lock
jim: (It was fixed)

src/components/forms/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,385 @@
...
+        return {
+          ...state,
+          focusedOption: action.option,
+          focusMode: FocusMode.Item,
+        }
+      case 'CLEAR':
+        return {
+          ...state,
+          inputValue: '',
+          selectedOption: undefined,
```
haworku, line 174: Also need to reset `filteredOptions` on this action to `props.action`

jim: I added a test stub to cover this case.

src/components/forms/ComboBox/ComboBox.tsx
jim: Added a test and implemented this.

haworku:
@jim  Checking in! Thanks for your work so far on this. What is the outstanding work here? Would love to get this merged and can help make sure that happens.

jim:
@haworku There are some missing tests to implement, and probably a few to write.

The biggest TODO, though, is to make a decision about how to proceed in terms of what props are required to use the ComboBox. Right now, we expect a `setFieldValue` function to be passed in, which is out of alignment with the rest of the components in this library (which only need to pass `onChange`).

haworku (approved):

PR #521, opened 2020-10-22
fix: Require Typescript 3.8+

### LINKED ISSUE #520: [fix] Require Typescript 3.8+
**Describe the bug**
We use some [Typescript 3.8 features](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export) in exporting the `Modal` type.  We need to update the `package.json` to reflect this.

haworku:
# Summary
- adjust typescript minimum version and fix `yarn build`

## Related Issues or PRs

resolves #520 

## How To Test
- clearing out `node_modules`, `yarn` and then `yarn build` should all run without bugs. 
- Tests and Happo should pass.

### DIFF
--- a/package.json
+++ b/package.json
@@ -119,7 +119,7 @@
     "stylelint-prettier": "^1.1.1",
     "stylelint-scss": "^3.17.1",
     "ts-jest": "^26.1.2",
-    "typescript": "^3.6.3",
+    "typescript": "^3.8.0",
     "url-loader": "^4.0.0",
     "webpack": "^4.41.0",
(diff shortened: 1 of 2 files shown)


ahobson (approved):
This is great!  I think we could also test with the example app being on a pre 3.8 version of typescript and seeing what happens?

haworku:
@ahobson it works as long as they have enough flexibility in their typescript declaration in their `package.json` such as `^3.x.x`.

Issue #538, opened 2020-10-26
Update example dependencies and dependabot config


ahobson:
1. Most of the example dependencies are not in `devDependencies` and they should be.  `dependabot` is only configured to update dev dependencies.
1. Dependabot only checks example once a month, but we should check it weekly as we don't want the example app to have security issues

PR #539, opened 2020-10-26
build(deps-dev): Allow dependabot to update all example deps

ahobson:
# Summary

Also move example deps to devDependencies where appropriate

## Related Issues or PRs

#538 

## How To Test

Merge and then check to see if dependabot does the right thing


haworku (approved):
LGTM

PR #694, opened 2020-12-07
feat: ComboBox component

### LINKED ISSUE #170: New component: Combo box

haworku:
# Summary
USWDS component: https://designsystem.digital.gov/form-controls/03-combo-box/

Current Status:
- ✅ basic functionality
     - Main component code is `ComboBox.tsx` and `ComboBox.test.tsx`. 
     - Includes storybook examples (`ComboBox.stories.tsx`) and implementation in example app (`/example` files). 
-  🔎 5 tests are skipped with comments. 
    - All represent valid minor bugs🐛, most relating to tab and hover behavior in the options dropdown.
    - _Should we consider these blockers to merging this first pass?_

## Related Issues or PRs
closes #170 

## How To Test
- Forms > ComboBox in storybook

### Screenshots (optional)
Screen Shot 2020-12-07 at 12 19 31 PM


src/components/forms/ComboBox/ComboBox.test.tsx
```diff
@@ -0,0 +1,1111 @@
...
+      const { getByTestId } = render(
+        <ComboBox
+          id="favorite-fruit"
+          name="favorite-fruit"
+          options={fruitOptions}
+          onChange={jest.fn()}
+        />
+      )
+
+      userEvent.type(getByTestId('combo-box-toggle'), 'app')
```
suzubara, line 382: is this meant to be `combo-box-input`?

src/components/forms/ComboBox/ComboBox.test.tsx
```diff
@@ -0,0 +1,1111 @@
...
+          onChange={jest.fn()}
+        />
+      )
+
+      userEvent.type(getByTestId('combo-box-input'), 'b')
+      expect(getByTestId('combo-box-input')).toHaveFocus()
+
+      userEvent.click(getByTestId('combo-box-clear-button'))
+      fireEvent.blur(getByTestId('combo-box-clear-button'))
+      expect(getByTestId('combo-box-input')).toHaveFocus()
```
suzubara, line 483: should the clear button be tabbable? I wasn't able to in chrome

src/components/forms/ComboBox/ComboBox.test.tsx
```diff
@@ -0,0 +1,1111 @@
...
+        />
+      )
+
+      fireEvent.click(getByTestId('combo-box-input'))
+      fireEvent.focus(getByTestId('combo-box-option-yuzu'))
+      fireEvent.keyDown(getByTestId('combo-box-option-yuzu'), {
+        key: 'ArrowDown',
+      })
+
+      expect(getByTestId('combo-box-option-yuzu')).toHaveFocus()
```
suzubara, line 723: fwiw, in Chrome when I press the down arrow after the last option it does not change focus

src/components/forms/ComboBox/ComboBox.test.tsx
```diff
@@ -0,0 +1,1111 @@
...
+      )
+
+      // Apple is the item at top of list
+      fireEvent.focus(getByTestId('combo-box-option-apple'))
+      fireEvent.keyDown(getByTestId('combo-box-option-apple'), {
+        key: 'ArrowUp',
+      })
+
+      expect(getByTestId('combo-box-option-apple')).toHaveFocus()
+      expect(getByTestId('combo-box-option-list')).toBeVisible()
```
suzubara, line 812: this is not the behavior I'm seeing.. in Chrome when I have selected Apricot and the Apple option is focused and I press the up arrow, the menu closes

src/components/forms/ComboBox/ComboBox.test.tsx
```diff
@@ -0,0 +1,1111 @@
...
+      const { getByTestId } = render(
+        <ComboBox
+          id="favorite-fruit"
+          name="favorite-fruit"
+          options={fruitOptions}
+          onChange={jest.fn()}
+        />
+      )
+
+      fireEvent.click(getByTestId('combo-box-input'))
```
suzubara, line 889: should this use the [dblclick event](https://developer.mozilla.org/en-US/docs/Web/API/Element/dblclick_event)?

src/components/forms/ComboBox/ComboBox.test.tsx
```diff
@@ -0,0 +1,1111 @@
...
+
+      expect(getByTestId('combo-box-input')).toHaveAttribute(
+        'aria-expanded',
+        'true'
+      )
+      expect(getByTestId('combo-box-option-list')).toBeVisible()
+    })
+
+    it('hides options list when clicking away', () => {
```
suzubara, line 899: should the options list be hidden if an option has focus and you click away? it hides when the input has focus and I click away but not if I've hovered over one of the options first

src/components/forms/ComboBox/ComboBox.test.tsx
```diff
@@ -0,0 +1,1111 @@
...
+          options={fruitOptions}
+          onChange={jest.fn()}
+        />
+      )
+
+      const input = getByTestId('combo-box-input')
+      userEvent.type(input, 'yu')
+      userEvent.click(getByTestId('combo-box-option-yuzu'))
+
+      expect(getByTestId('combo-box-input')).toHaveValue('Yuzu')
```
suzubara, line 997: It might be worth testing that the `onChange` handler was called with the value here

suzubara (commented):
wow, this is an incredibly impressive PR and v1 of this component 👏  !! I think it is really solid and fine to follow up on the outstanding bugs as additional issues after merging. I left some comments, mostly noting differences between the tests and my UX manually testing in Chrome, and some questions.

The only thing I would consider a blocker is that I am seeing the selected value clear out whenever I blur, making it impossible to actually persist a value in the form (see gif). if we can fix that then I think this is good to go!

ComboBox bug

src/components/forms/ComboBox/ComboBox.test.tsx
haworku: Yes. This is documented in [a failing test](https://github.com/trussworks/react-uswds/blob/f6a08d3f59155e54c790f050c836a3826ff0c882/src/components/forms/ComboBox/ComboBox.test.tsx#L726-L746). Should it be a blocker? I'm thinking yes.

src/components/forms/ComboBox/ComboBox.test.tsx
haworku: Yes. This documented in [a failing test](https://github.com/trussworks/react-uswds/blob/f6a08d3f59155e54c790f050c836a3826ff0c882/src/components/forms/ComboBox/ComboBox.test.tsx#L706-L724).

src/components/forms/ComboBox/ComboBox.test.tsx
haworku: 🐛  That's a bug, will fix and write a test.

src/components/forms/ComboBox/ComboBox.test.tsx
haworku: 🐛 This is a bug and this test should be failing.  Changing `fireEvent.focus` to `userEvent.hover(getByTestId('combo-box-option-apple'))` makes it fail as it should.  Great catch!

haworku:
> The only thing I would consider a blocker is that I am seeing the selected value clear out whenever I blur, making it impossible to actually persist a value in the form (see gif). if we can fix that then I think this is good to go!

Fixed in e0ebc24. Thanks!!

suzubara (approved):
Verified that the fixes look great in Storybook! They don't seem to be reflected in the example app but that is an existing issue. This is awesome!! 🚢  🚀  🔥  🎉 

(I took a look at the Happo diffs and they look fine, maybe will be resolved once this branch is updated from `main` or they could be a result of Happo updating their browsers)

suzubara (approved):

PR #696, opened 2020-12-09
docs: All uswds scss modules now applied in Storybook

### LINKED ISSUE #191: [bug] uswds typography mixin not being applied in storybook

haworku:
# Summary
Fix #191 and add comments in other stylesheet files to better explain whats happening.

## Related Issues or PRs
resolves #191 

## How To Test
-  Hover on footer nav links should show blue color (if not already visited). If your links have already been visited you may need to clear history for localhost.

### DIFF
--- a/.storybook/custom-styles.scss
+++ b/.storybook/custom-styles.scss
@@ -1,7 +1,3 @@
-// Import react-uswds scss
-@import '../src/styles/index.scss';
-
-// Add custom styles for storybook implementation
 .custom-class {
   background: #fa9441;
 }
--- a/.storybook/custom-styles.scss
+++ b/.storybook/custom-styles.scss
@@ -5,9 +5,3 @@
 .usa-color-text-visited {
   color: color($theme-link-visited-color);
 }
-
-.usa-link:visited {
-  &:not(.usa-color-text-visited) {
-    color: color($theme-link-color);
-  }
-}
--- a/.storybook/index.scss
+++ b/.storybook/index.scss
@@ -1,0 +1,5 @@
+// Import react-uswds scss
+@import '../src/styles/index.scss';
+
+// Custom storybook styles
+@import './custom-styles.scss';
--- a/.storybook/index.scss
+++ b/.storybook/index.scss
@@ -1,5 +1,14 @@
-// Import react-uswds scss
-@import '../src/styles/index.scss';
+// Establish settings needed to customize uswds further
+$theme-image-path: '~uswds/src/img';
+$theme-font-path: '~uswds/src/fonts';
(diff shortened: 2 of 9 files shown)


haworku:
Turning back to draft since Happo is showing some interesting discrepancies that I will look into more.

haworku:
This ticket adds all USWDS scss to Storybook in order to use custom variables. However, the change is causing some components to break in certain browsers as seen in the [Happo diff](https://happo.io/a/371/p/453/compare/c0858b236549029f4caf81beaed62f702b8eb353/c8dc6528f8a1590acd869d41bcbf2ea2c1793cf7) and the screenshot below. This seems like an autoprefixing issue.
 
Notes:
-  UWSDS scss requires autoprefixing to work (see [USWDS installation docs](https://designsystem.digital.gov/documentation/developers/#installation) - Sass Compilation Requirements). 
- We need to adjust Storybook webpack configuration to do this, for example by using `autoprefixer` and `postcss-loader`).  It is a relatively [a simple change,](https://github.com/postcss/autoprefixer#webpack) particularly since Storybook already includes the `postcss-loader`.  
- However 😇 Storybook's version is of `postcss-loader` is out of date. It needs to be updated to version 8+ to work, which is [well documented](https://github.com/storybookjs/storybook/issues/13491).  So right now everything breaks when we add autoprefixing (see  3a14ad0 ).
- In the screenshot below the left side is the changes from this ticket, the right is whats on our live storybook. You can see the autoprefixing issue in the CSS
Screen Shot 2020-12-21 at 10 36 29 AM

haworku:
Closing as "won't fix" due to reasons described in #191

PR #726, opened 2020-12-21
docs: Add request for implementation issue template

haworku:
# Summary
The intention for this issue template is to provide an even easier way for existing Truss projects to suggest a new component or share their implementation with react-uswds  maintainers.  This was an action item from the monthly react-uswds meeting sync.


ahobson (approved):
I'm good with this, but it would be nice to have someone that has needed an implementation to also take a look

haworku:
@sarboc you would be a good additional reviewer for this ticket since your team has components they'd like to easily share back

suzubara (approved):
I know you're waiting on a review from someone who's had this use case, but looks good to me anyway!

PR #767, opened 2021-01-06
docs: Update license to 2021

haworku:
# Summary
It's a new year. Also there were a bunch of broken links 🤕 from an earlier refactor.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -174,28 +174,3 @@
       of your accepting any such warranty or additional liability.
 
    END OF TERMS AND CONDITIONS
-
-   APPENDIX: How to apply the Apache License to your work.
-
-      To apply the Apache License to your work, attach the following
-      boilerplate notice, with the fields enclosed by brackets "[]"
-      replaced with your own identifying information. (Don't include
-      the brackets!)  The text should be enclosed in the appropriate
-      comment syntax for the file format. We also recommend that a
-      file or class name and description of purpose be included on the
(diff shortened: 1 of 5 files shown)


LICENSE
```diff
@@ -174,28 +174,3 @@
...
-      To apply the Apache License to your work, attach the following
-      boilerplate notice, with the fields enclosed by brackets "[]"
-      replaced with your own identifying information. (Don't include
-      the brackets!)  The text should be enclosed in the appropriate
-      comment syntax for the file format. We also recommend that a
-      file or class name and description of purpose be included on the
-      same "printed page" as the copyright notice for easier
-      identification within third-party archives.
-
-   Copyright 2019, TrussWorks, Inc.
```
ahobson, line 189: I don't think we need that APPENDIX paragraph, but don't we want the bits below, including the Copyright?

ahobson (changes requested):

LICENSE
haworku: Hmmm... so this text is what we already have main README - See https://github.com/trussworks/react-uswds#license 

By keeping it here too we have to update things in 2 places.  Pretty sure we have things proper in the README alone is enough. And then this is just a link to actual text.  

Thoughts?

ahobson: Ok, I see now. Thank you for explaining!

ahobson (approved):
👍

brandonlenz (approved):

PR #782, opened 2021-01-12
feat: Update GovBanner to match USWDS 2.8.0 release

### LINKED ISSUE #456: [feat] Update GovBanner to include HTTPS section
**Does your feature request relate to a specific USWDS feature?**
This issue is about improving our `GovBanner` component to match improvements to USWDS [Banner ](https://designsystem.digital.gov/components/banner/). The 2.8.0 release [made changes to the Banner](https://designsystem.digital.gov/whats-new/updates/2020/09/08/Improvements-banner/) that still need to be reflected in our library

- Add HTTP section to `GovBanner`
- Update the default English text our in Storybook examples (`GovBanner.stories.tsx`) to match the new default text.
- Make any other changes to Storybook that make our example more in sync with [the USWDS implementation](https://designsystem.digital.gov/components/banner/)

brandonlenz:
# Summary

## Related Issues or PRs

closes #456 

## How To Test

Verify copy changes in Storybook for .gov and .mil versions of the banner for both English and Spanish

### Screenshots (optional)
Storybook stories:
image

defaults to "english" & ".gov":
image

with "english" and ".gov" passed:
image

with "english" and ".mil" passed:
image

with "spanish" and ".gov" passed:
image

with "spanish" and ".mil" passed:
image

### DIFF
--- a/src/components/GovBanner/GovBanner.tsx
+++ b/src/components/GovBanner/GovBanner.tsx
@@ -5,14 +5,122 @@ import classnames from 'classnames'
 import flagImg from 'uswds/src/img/us_flag_small.png'
 import dotGovIcon from 'uswds/src/img/icon-dot-gov.svg'
 import httpsIcon from 'uswds/src/img/icon-https.svg'
+import lockIcon from 'uswds/src/img/lock.svg'
+
+enum Language {
+  ENGLISH = 'en',
+  SPANISH = 'es',
+}
+
+enum TLD {
+  DOT_GOV = '.gov',
+  DOT_MIL = '.mil',
+}
+
+interface GovBannerCopy {
+  header: string
+  headerAction: string
+  tldSectionHeader: string
+  tldSectionContent: JSX.Element
+  httpsSectionHeader: string
+  httpsSectionContent: JSX.Element
+}
+
+const getCopy = (language: Language, tld: TLD): GovBannerCopy => {
+  const lock = (
+    <span className="icon-lock">
+      <img src={lockIcon} className="usa-banner__lock-image" alt="lock" />
+    </span>
+  )
+
+  switch (language) {
+    case Language.ENGLISH:
+      return {
+        header: 'An official website of the United States government',
+        headerAction: 'Here’s how you know',
+        tldSectionHeader: `The ${tld} means it’s official.`,
+        tldSectionContent: ((): JSX.Element => {
+          switch (tld) {
+            case TLD.DOT_GOV:
+              return (
+                <>
+                  A <strong>{TLD.DOT_GOV}</strong> website belongs to an
+                  official government organization in the United States.
+                </>
+              )
+            case TLD.DOT_MIL:
+              return (
+                <>
+                  A <strong>{TLD.DOT_MIL}</strong> website belongs to an
+                  official U.S. Department of Defense organization.
+                </>
+              )
+          }
+        })(),
+        httpsSectionHeader: `Secure ${tld} websites use HTTPS`,
+        httpsSectionContent: (
+          <>
+            A <strong>lock ({lock})</strong> or <strong>https://</strong> means
+            you’ve safely connected to the {tld} website. Share sensitive
+            information only on official, secure websites.
+          </>
+        ),
+      }
+    case Language.SPANISH:
+      return {
+        header: 'Un sitio oficial del Gobierno de Estados Unidos',
+        headerAction: 'Así es como usted puede verificarlo',
+        tldSectionHeader: `Los sitios web oficiales usan ${tld}`,
+        tldSectionContent: ((): JSX.Element => {
+          switch (tld) {
+            case TLD.DOT_GOV:
+              return (
+                <>
+                  Un sitio web <strong>{TLD.DOT_GOV}</strong> pertenece a una
+                  organización oficial del Gobierno de Estados Unidos.
+                </>
+              )
+            case TLD.DOT_MIL:
+              return (
+                <>
+                  Un sitio web <strong>{TLD.DOT_MIL}</strong> pertenece a una
+                  organización oficial del Departamento de Defensa de EE. UU.
+                </>
+              )
+          }
+        })(),
+        httpsSectionHeader: `Secure ${tld} websites use HTTPS`,
+        httpsSectionContent: (
+          <>
+            Un <strong>candado ({lock})</strong> o <strong>https://</strong>{' '}
+            significa que usted se conectó de forma segura a un sitio web {tld}.
+            Comparta información sensible sólo en sitios web oficiales y
+            seguros.
+          </>
+        ),
+      }
+  }
+}
+
+interface GovBannerProps {
+  tld?: TLD
+  language?: Language
+}
 
 export const GovBanner = (
-  props: JSX.IntrinsicElements['section']
+  props: GovBannerProps & JSX.IntrinsicElements['section']
 ): React.ReactElement => {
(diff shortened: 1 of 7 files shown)


trussworks-infra:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against f8db555f412ea59ea6758ad81bcbb63eee9ae1c2
</p>

brandonlenz:
> Warnings
> ⚠️	This PR does not include changes to tests, even though it affects source code.
> Generated by 🚫 [dangerJS](https://danger.systems/js) against [f8db555](https://github.com/trussworks/react-uswds/commit/f8db555f412ea59ea6758ad81bcbb63eee9ae1c2)

I'm curious what reviewers think about for testing this story. 

I'm cautious to test against specific verbiage being rendered, as that makes for tests that are low in value in terms of likelihood in catching breaking changes, but do make future updates more disproportionally cumbersome. Using something like Enzyme would allow testing of actual property values, which would allow us to test default values for props when not explicitly passed. What are your thoughts on that?

In 7efcce7 I added a basic test verifying render with the newly added props. I'd be happy to expand on that by testing the actual verbiage if maintainers agree there is value in that kind of test.

sojeri:
@suzubara / @haworku --- I didn't make it all the way through the happo diff, but is it normal to see so many false positives? when I saw the uswds bump I thought that might be the trigger, but switching to diff view the vast majority of components seem to have no visible changes. any idea why these are showing seemingly empty diffs or whether that's a normal behavior here?

literally the only change I saw in the components I reviewed was the warning icon changing ever so slightly.

brandonlenz:
> @suzubara / @haworku --- I didn't make it all the way through the happo diff, but is it normal to see so many false positives? when I saw the uswds bump I thought that might be the trigger, but switching to diff view the vast majority of components seem to have no visible changes. any idea why these are showing seemingly empty diffs or whether that's a normal behavior here?
> 
> literally the only change I saw in the components I reviewed was the warning icon changing ever so slightly.

@sojeri I meant to address that here as well with my understanding of a lot of it. Indeed, it looks like some of the icons may have changed ever so slightly. I also noticed the note at the top of Happo:

```
Some updates were made to Happo workers that might cause unexpected diffs:
1/11/2021, 5:08:00 AM - Our Chrome workers have been updated to run Chrome version 88.
```

Many of the diffs _were_ from Chrome.

There are some valid diffs from the 2.8.0 upgrade, for instance, the proper alignment of the collapse-icon for the GovBanner:
image

image

All that said, I am new to Happo, so I'm not sure what is normal to see.

haworku:
> In 7efcce7 I added a basic test verifying render with the newly added props. I'd be happy to expand on that by testing the actual verbiage if maintainers agree there is value in that kind of test.

Thanks for asking about tests, @brandonlenz !  imo we should add a basic test for the case where these new props are not passed in at all. That is how most people using this component right now out in the wild (since this is new work).  The component should still render with the English text and not break for those folks. 

I hear you on Enzmye,  sounds like a bigger topic we could discuss. Want to make an issue about it? I don't think this should block this work from going through now. This def an outlier component because all it does really is display very specific text which we try to avoid (#124).

> I didn't make it all the way through the happo diff, but is it normal to see so many false positives?
Um... sometimes. 😇 Lately Happo has been catching bugs for us and has been pretty on point. However, this does look like a bunch of false positives maybe related to what you all already identified. Especially when its just a few pixels change for unrelated components I usually push it through. Its approved now. ✅

brandonlenz:
> > In [7efcce7](https://github.com/trussworks/react-uswds/commit/7efcce79c8df482f0be72852014fc2b2414c1e53) I added a basic test verifying render with the newly added props. I'd be happy to expand on that by testing the actual verbiage if maintainers agree there is value in that kind of test.
> 
> Thanks for asking about tests, @brandonlenz ! imo we should add a basic test for the case where these new props are not passed in at all. That is how most people using this component right now out in the wild (since this is new work). The component should still render with the English text and not break for those folks.

@haworku The existing test for rendering the component without props should work for this purpose:

```javascript
    const { queryByTestId } = render(<GovBanner />)
    expect(queryByTestId('govBanner')).toBeInTheDocument()
  })
```

> I hear you on Enzmye, sounds like a bigger topic we could discuss. Want to make an issue about it? I don't think this should block this work from going through now. This def an outlier component because all it does really is display very specific text which we try to avoid (#124).

Sure, I'll create an issue to start discussion on that. I can definitely see an argument being made to keep this repo lean. 

Also thanks for linking #124. I definitely agree this could be slippery slope territory. For the purposes of this PR, if we only want to consider english for now I can remove the logic specific to "spanish", and add optional props in the spirit of 

>it sounds like we have agreement about using props (with English default values) to handle any necessary language strings (#124)

Otherwise I could see someone taking a look at what I've done here and implementing similar logic in the name of convention elsewhere.

brandonlenz:
> Sure, I'll create an issue to start discussion on that. I can definitely see an argument being made to keep this repo lean.

Added #783

haworku (dismissed):
🏅 Thanks for making the Enzyme issue as well. 

I think this implementation and the way props are handled make sense for the use case. It's an exception to the rule, yes, and we always have those. Good work

src/components/GovBanner/GovBanner.tsx
```diff
@@ -5,15 +5,117 @@ import classnames from 'classnames'
...
+      <img src={lockIcon} className="usa-banner__lock-image" alt="lock" />
+    </span>
+  )
+
+  switch (language) {
+    case 'english':
+      return {
+        header: 'An official website of the United States government',
+        headerAction: 'Here’s how you know',
+        tldSectionHeader: `Offical websites use ${tld}`,
```
suzubara, line 35: typo: `Offical > Official`

suzubara (commented):
This looks awesome! I found one typo that should be resolved before merging. I think it would be good to consider a future abstraction where each of the `copy` values can be passed in as props in order to easily allow for other localizations, but since the USWDS spec only explicitly provides copy for English & Spanish I think this is sufficient.

as for the testing question, I think this might be a good situation to use Jest snapshot testing? I agree writing RTL tests that look for specific copy will be tedious to maintain, and this component is unique in that it's purpose is to render specific content. I know we already get visual tests from Happo, but if we want more confidence that this component will not change unexpectedly, snapshot tests would be my suggestion.

src/components/GovBanner/GovBanner.tsx
brandonlenz: Good eye! 1378092

brandonlenz: Added snapshot tests in d8b3a91

I hadn't actually used those before, so this was a good learning experience. Please verify what I ended up doing is in line with convention for Snapshot tests at your convenience 😄

trussworks-infra:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against d8b3a91328130c5ac5f5101566543cdff025f0f7
</p>

suzubara (approved):
awesome, it's a ✅  from me!

suzubara (approved):

PR #803, opened 2021-01-19
feat: DatePicker component

### LINKED ISSUE #338: New Component: Date Picker
# [Date Picker](https://designsystem.digital.gov/components/form-controls/#date-picker)
Add a new `DatePicker` component following our [adding new components](https://github.com/trussworks/react-uswds/blob/main/docs/adding_new_components.md) guide. The USWDS implementation in linked in the title text of this issue.

## Description
A date picker helps users select a single date. Use the new `DatePicker` when scheduling or when day of the week is important.  Use existing `DateInput`  when desired date is already known (such as a birthday). 

## Requirements (proposed)
**Prop Interface**
``` 
DatePickerProps = {
  id: string,
  name: string,
  inputRef: InputRefType,
  isValid?: (inputValue) => boolean
  defaultValue?: string (YYYY-MM-DD)
```
**More Details**
- display with existing `<Label />` (which also has `hint` prop)
- current `DatePicker` implementation only validates for `mm/dd/yyyy`. Even `mm-dd-yyyy` is invalid.  We support a variety of validation formats with an `isValid` function prop.  
- We may want other props to handle setting a default date and setting the range the calendar displays. See [docs](https://designsystem.digital.gov/components/form-controls/#date-picker) for the usage of `data-` attributes (`data-default-date`, `data-range-date` etc). 
- see [uswds data picker js](https://github.com/uswds/uswds/blob/develop/src/js/components/date-picker.js) for several helpful date utilities

suzubara:
# Summary

This PR adds the [USWDS DatePicker component](https://designsystem.digital.gov/components/form-controls/#date-picker).

I worked off of the [USWDS DatePicker source code](https://github.com/uswds/uswds/blob/develop/src/js/components/date-picker.js) and attempted to re-implement as closely to that spec as possible using React paradigms and features where appropriate. The `constants.ts` and `utils.ts` are almost exact copies of the constants and utility methods from the source, other than being converted to TypeScript.

The exported component and its props interface are:
```
<DatePicker />

interface DatePickerProps {
  id: string
  name: string
  className?: string
  disabled?: boolean
  defaultValue?: string
  minDate: string
  maxDate?: string
  rangeDate?: string
```

This renders an internal hidden text `<input />` and an external visible text `<input />`, a button for toggling the date picker dialog, and hidden status text for screen readers. As described in the USWDS spec, dates passed into any of the `props` must use the format `YYYY-MM-DD` and a date typed directly into the external input must use the format `MM/DD/YYYY`.

When opening the date picker dialog, the `<Calendar />` subcomponent is displayed and can be used to select a date from the current month, navigate to previous & next months and years, as well as select specific months and years by clicking on the currently selected ones (which toggles display of the `<MonthPicker />` and `<YearPicker />` subcomponents).

The `minDate` prop defaults to `01-01-0000` and there is no default `maxDate`. If a min and/or max are specified, dates outside of that range become disabled in the Calendar UI, and if an invalid date is typed into the input the input itself [becomes invalid](https://developer.mozilla.org/en-US/docs/Web/API/HTMLObjectElement/setCustomValidity). No error UI is implemented - this is up to consumers to handle.

----

**Note:** There is one small difference in functionality between this component and the USWDS implementation, related to validating the input. The USWDS implementation validates when:
- setting the initial value based on the default value passed in
- clicking on a date in the calendar UI
- typing the Enter key in the external text input
- on focusout (blur) of the external text input

Because this component uses the `useEffect` hook to trigger validation whenever the date value changes (regardless of how), the React DatePicker will validate when:
- setting the initial value based on the default value passed in (same as above)
- clicking on a date in the calendar UI (same as above)
- on input (change) of the external text input

This is a minor discrepancy but one still worth noting, so I am adding it to the Storybook docs for the component. It's also worth mentioning that validation in this case is just calling [`setCustomValidity`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLObjectElement/setCustomValidity) on the external text input, and library users should be able to determine how & when they want invalid UI to display by inspecting the [ValidityState](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState) of the external input. We may find that we want to expose props for custom event handlers or even a ref to the component for better integration with 3rd party form libraries, but I would recommend leaving that for a future story once we have a better understanding of how this component will be used.

----

## Related Issues or PRs

Fixes #338

## How To Test

I recommend starting Storybook and viewing the Date picker stories. The unit tests should cover all expected interaction and functionality, so reading through those and reproducing them manually in the browser is also a good way to verify that it works as expected.

Subcomponent stories (such as those under Date picker / Calendar, Date picker Day, etc.) are for testing visual cases only and may not work as expected without the context of the full Date Picker.

There are several different cases to test:
- Default Date Picker (no default value, no min/max date)
- Disabled (no interaction should be possible)
- With Default Value (date picker is initialized with an existing value)
- With Default Invalid Value (date picker is initialized with an existing value that falls outside the specified min/max dates, so it should be coerced to a valid value when opened)
- With Min Max in Same Month (date picker with a min/max date in one month, so any navigation to other months/years should be disabled)
- With Min Max (min/max dates further out, navigation should be disabled once the bounds are reached)
- With Range Date (this is a visual effect only, that highlights the dates between the given `rangeDate` prop and the currently selected date. this prop will be used when implementing the [Date Range Picker](https://designsystem.digital.gov/components/form-controls/#date-range-picker))
- With External Input (renders an additional input element in order to test closing the Date Picker when focus is lost)

## Screenshots

**Calendar hidden (default state on mount)**
image

**Calendar open**
image

**Month picker**
image

**Year picker**
image


suzubara:
As noted in the PR description, I've encountered a blocker to finishing this, which is that React 17 is required in order to implement the `focusout` event correctly (this will close the Calendar dialog when focus is lost). Since we already have an [issue for updating React](https://github.com/trussworks/react-uswds/issues/805) I will plan to complete that before coming back to this.

In the meantime, any feedback on the general structure of this component is appreciated! My biggest open question is whether it's worth converting any of the existing state hooks to `useReducer`.. I started to work on this but felt undecided as to whether it made sense, since a lot of the handlers depend on the context of the component (such as knowing what the min/max dates are).

src/components/forms/DatePicker/Calendar.tsx
```diff
@@ -0,0 +1,327 @@
+import React, { useState, useRef, useEffect } from 'react'
+
+import {
+  DAY_OF_WEEK_LABELS,
+  DAY_OF_WEEK_SHORT_LABELS,
+  MONTH_LABELS,
+} from './constants'
```
haworku, line 7: Can you please surface these as optional props for i18n purposes?

src/components/forms/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,288 @@
...
+      )
+
+      setCalendarDisplayValue(displayDate)
+      setCalendarPosY(datePickerEl?.current?.offsetHeight)
+
+      const statuses = [
+        'You can navigate by day using left and right arrows',
+        'Weeks by using up and down arrows',
+        'Months by using page up and page down keys',
+        'Years by using shift plus page up and shift plus page down',
```
haworku, line 172: 🤔  More lang that is not i18n friendly

src/components/forms/DatePicker/utils.tsx
```diff
@@ -0,0 +1,522 @@
+import React from 'react'
+
+import { DEFAULT_EXTERNAL_DATE_FORMAT, INTERNAL_DATE_FORMAT } from './constants'
+
+/**
+ * This file contains the USWDS DatePicker date manipulation functions converted to TypeScript
+ */
```
haworku, line 7: So I get that we are just doing this for speed and copy/pasting USWDS tests but I'm a bit weirded out that this doesn't have 100% unit tests coverage for every function for maintenance reasons. Maybe we should backlog a story to either test everything here or else swap with an external dependency (some date/time lib) that would already be well tested?

haworku (commented):
This all looks good so far 😇, thanks for keeping the hooks organized that helps a lot. I didn't run yet just cuz I know things in progress .

Will be interested to see what ends up being similar to Combobox or not. Originally I was thinking we could share some logic but I can see how different they are when I look closer.

Made some comments about stuff to consider.

src/components/forms/DatePicker/utils.tsx
suzubara: USWDS didn't seem to have tests for these methods at all. I started writing some but I haven't gotten them all.. I'm hoping to get as much coverage on these as possible but some of them are just tricky because dates

haworku: Oh wow. 🥇 Okay well then thank you for starting on this then, I misunderstood!  Do you not want to lean on a lightweight date/time lib or util? I can see why we wouldn't want to do that just for this component... but just checking

suzubara: yeah I thought about it, but it would mean importing a new lib and I'm making the assumption that USWDS wrote their own utils to keep things lightweight. I think I want to just follow their decision on this.

brandonlenz (commented):
This is looking awesome! Beyond what's been highlighted already I've just got a small point of feedback at this point.

src/components/forms/DatePicker/utils.tsx
```diff
@@ -0,0 +1,522 @@
...
+
+/**
+ * Set date from month day year
+ *
+ * @param {number} year the year to set
+ * @param {number} month the month to set (zero-indexed)
+ * @param {number} date the date to set
+ * @returns {Date} the set date
+ */
+export const setDate = (year: number, month: number, date: number): Date => {
```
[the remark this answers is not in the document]
brandonlenz: Silly me didn't see that this is exactly how USWDS has them named as well. Better to follow what they've done as a convention then.

trussworks-infra:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 170cab8156060321489cd0b37e34e14ba5ddbd95
</p>

src/components/forms/DatePicker/utils.tsx
suzubara: this also occurred to me later on, but if USWDS makes updates to their DatePicker code then it will be harder to reconcile differences if we deviate from them.

src/components/forms/DatePicker/Calendar.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { Calendar } from './Calendar'
+import { FocusMode } from './DatePicker'
+import { parseDateString } from './utils'
+
+export default {
+  title: 'Components/Form controls/Date picker/Calendar',
+  component: Calendar,
+  argTypes: {
+    handleSelectDate: { action: 'select date' },
+    setStatuses: { action: 'set statuses' },
```
haworku, line 12: Right now all the` <Calendar> `and `<YearPicker>` stories break for me in Storybook when I try to interact with the components `TypeError: setStatuses is not a function`. Perhaps Storybook actions don't work nicely for mocking behavior in this case? 

Happy to pair on this and figure out a workaround. Or, another idea, we could comment out the export line for stories that aren't reflected in components we export from `index.ts`. Seems like these stories are more for our internal use as library developers, we could comment them in when we need to check stuff, but otherwise I don't think they need to go into the published Storybook.

haworku (commented):
Things look good. I am still reading through everything but I spent a good amount of time manual testing.

About Storybook

src/components/forms/DatePicker/Calendar.tsx
suzubara: Issue added: https://github.com/trussworks/react-uswds/issues/866

suzubara:
> Things look good. I am still reading through everything but I spent a good amount of time manual testing.
> 
> About Storybook

@haworku I'm not seeing either of these issues myself! I've since reconciled this branch with the latest `main`, do you mind double checking to see if they are still happening? and/or maybe we can pair and see if we can figure out what's going on?

Also I think you have a good point about not exporting stories for components that are not meant to be used externally. They're super useful for development, but are probably confusing if included in our published Storybook. I will comment those out!

haworku:
@suzubara ✅  I see everything now. I cleared out node modules again something must have been in a really funky state.

src/components/forms/DatePicker/DatePicker.test.tsx
```diff
@@ -0,0 +1,527 @@
...
+import userEvent from '@testing-library/user-event'
+
+import { DatePicker } from './DatePicker'
+import { today } from './utils'
+import { DAY_OF_WEEK_LABELS, MONTH_LABELS } from './constants'
+
+/* 
+OUTSTANDING:
+- focus out on date picker external input validates
+- keydown on date picker input validates if keycode is enter
```
haworku, line 12: Are these validation tasks still outstanding? If so lets file bugs for them.  I think one of these has a skipped test later on in the file, seems like we could remove it for now from here and add to the issue.

haworku (dismissed):
LGTM ✅ 

small housekeeping things were:
- comment out stories for components that aren't exported to reduce confusion 
- backlog an issue if there are any outstanding validations bugs (code comments suggest there are?). I couldn't find any in terms of manual testing.

package.json
```diff
@@ -66,8 +66,7 @@
   "devDependencies": {
     "@babel/core": "^7.10.5",
     "@babel/preset-react": "^7.10.4",
-    "@storybook/addon-docs": "^6.1.5",
-    "@storybook/addon-viewport": "^6.1.5",
```
brandonlenz, line 70: These two dependencies are mentioned in [`.github/dependabot.yml`](https://github.com/trussworks/react-uswds/blob/main/.github/dependabot.yml). Can we update it to reference '@storybook/addon-essentials"?

src/components/forms/DatePicker/DatePicker.stories.tsx
```diff
@@ -0,0 +1,114 @@
...
+- clicking on a date in the calendar UI
+- typing the Enter key in the external text input
+- on focusout (blur) of the external text input
+
+Because this component uses the useEffect hook to trigger validation whenever the date value changes (regardless of how), the React DatePicker will validate when:
+- setting the initial value based on the default value passed in (same as above)
+- clicking on a date in the calendar UI (same as above)
+- on input (change) of the external text input
+
+It's also worth mentioning that validation in this case is just calling [setCustomValidity](https://developer.mozilla.org/en-US/docs/Web/API/HTMLObjectElement/setCustomValidity) on the external text input, and library users should be able to determine how & when they want invalid UI to display by inspecting the [ValidityState](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState) of the external input.
```
brandonlenz, line 33: Is it worth adding a storybook story demonstrating this? I was able to test the validity as follows:
1. Assign an id to the `input` in a storybook story
2. Enter an invalid date
3. Check validity via `document.getElementById('my-id').checkValidity()`
    - returns `false`
4. Select/enter valid date
    - returns `true`

I wouldn't hold up this PR just for this story, so maybe we make it an issue as a nice-to-have for the storybook docs?

brandonlenz (commented):
I have just a few questions, but otherwise this looks amazing! I know a lot of work went into this. Awesome job!

src/components/forms/DatePicker/DatePicker.stories.tsx
```diff
@@ -0,0 +1,114 @@
...
+    defaultValue="2021-01-20"
+    rangeDate="2021-01-08"
+  />
+)
+
+export const withExternalInput = (): React.ReactElement => (
+  <>
+    <DatePicker id="birthdate" name="birthdate" />
+    <input type="text" data-testid="test-external-element" />
+  </>
```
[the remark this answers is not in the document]
suzubara: @brandonlenz I _think_ this was to demonstrate that the Calendar closes if focus moves to another element and/or opening the Calendar creates a focus trap (so you can't tab to the next input until it's closed). I probably just added it to test and don't mind removing if it's confusing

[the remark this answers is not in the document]
brandonlenz: Ohhh! Got it!

That makes sense. 

I'd say it's a valuable story to have. I would consider renaming the story to be more specific to how it relates to focusing.

I was confused because I was expecting the external input to be magically hooked into the value/validation of the `DatePicker` itself (since `DatePicker` internally has an "externalInput") for some use case in this story 🤦

src/components/forms/DatePicker/DatePicker.stories.tsx
suzubara: after thinking about this I started going down a bit of a rabbit hole regarding exposing refs to input components that are more complex and use refs to the inputs internally themselves. I believe I have somewhat figured it out, but it’s a new chunk of code that to me feels like maybe outside of “mvp” for DatePicker. the factors I’m considering are:
- exposing refs to input components is usually important so that consumers can use them to do things like imperatively focus on an input, and check the native validity state of it
- I think the answer is to use `React.forwardRef` on the DatePicker component, and `useImperativeHandle` to expose important methods/properties back to the parent ref
- this was mostly through exploration from the last hour or so, and I don’t feel comfortable merging these additions without more known use cases of where this feature is needed
- React very explicitly says if you start using `forwardRef` in a library, treat it as a breaking change. Meaning if we merge the DatePicker without it, and then we add it in, it will be a major version update
- I have a suspicion that the ComboBox component will need similar treatment, so adding `forwardRef` there would also require a major version update

based on all of the above, this is my inclination:
- Merge the DatePicker component as-is, with no exposed ref
- Do some more exploratory work, using the Example app or other consumer projects, to use the DatePicker and ComboBox inputs in Real Life Forms and feel out the limitations caused by not having access to the input ref
- Create a new issue for resolving the above, with the assumption that this will most likely require using `fowardRef` on these components and trigger a new major version of the library

References:
- https://reactjs.org/docs/forwarding-refs.html#note-for-component-library-maintainers
- https://reactjs.org/docs/hooks-reference.html#useimperativehandle
- https://medium.com/@ilxanlar/useimperativehandle-by-examples-99cbdc8e3c3a

src/components/forms/DatePicker/DatePicker.stories.tsx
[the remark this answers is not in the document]
suzubara: I got rid of this story and added a second input to the "Complete Datepicker" example to hopefully illustrate the same thing in a less confusing way!

package.json
suzubara: done

brandonlenz (approved):
All good from my end. Great work!

PR #804, opened 2021-01-19
feat: New Component: Breadcrumb

### LINKED ISSUE #337: New Component: Breadcrumb
# [Breadcrumb](https://designsystem.digital.gov/components/breadcrumb/)

Add a new Breadcrumb component following our [adding new components](https://github.com/trussworks/react-uswds/blob/main/docs/adding_new_components.md) guide. The USWDS implementation in linked in the title text of this issue.

## Description
 Breadcrumbs provide secondary navigation to help users understand where they are in a website. 
> A breadcrumb bar (or breadcrumbs, or breadcrumb trail) is a series of links that describes the hierarchical relationship between the current page and the overall site structure. Breadcrumbs allow a user to navigate “up” instead of using the “Back” button to return to the previous page. A breadcrumb bar facilitates orientation when a user navigates to an interior page from a web search or link from another site.

## Requirements 
**Prop Interface**
``` 
BreadcrumbProps = {
  variant: 'default' | 'wrap' 
  withMetadata: boolean,
 children: React.React Node // required, these are the list items
}
```
**More Details**
- `<nav>` root element
-  breadcrumbs are an ordered list - `<ol>` inside the nav.  _"This allows assistive technology to enumerate the items in the breadcrumbs and allows shortcuts between list items."_
- children should already be in hierarchy order
- `withMetadata` adds markup SEO support, includes `<meta>` tags and adds special attributes to the `ol`. Make this prop a boolean and only support JSON-LD for first go around.  
- use `aria-current="page"` with the value of the current page in the Storybook example.  

## Other Considerations for future work (out of scope for now)
- Alternatives to wrapping
>  In general, rely on truncating the title of the current page over wrapping breadcrumb text. But usability comes first. Consider alternative approaches if the title of the current page is completely truncated. For example, a mobile-friendly breadcrumb may show only a page's direct parent. Sites with very long breadcrumb trails might ultimately need to wrap breadcrumbs, or consider flattening the information architecture of the site. 
- More robust metadata support.  Allow people to specify, or feed in attributes for,  JSON-LD (recommended by uswds) or RDFa for metadata schema.

brandonlenz:
# Summary

Add [USWDS Breadcrumb Component](https://designsystem.digital.gov/components/breadcrumb/).

## Related Issues or PRs

Fixes #337 

## How To Test

Storybook remains the best way to interact with the component, especially regarding meta data. I would recommend inspecting the elements from storybook to verify:

```
git checkout --track origin/bl-breadcrumb-337
yarn storybook
```

There are also several jest tests that cover basic rendering and uswds class application.

### Screenshots (from storybook)

---

**Default Breadcrumb:**
image

Note: non-wrapping breadcrumbs truncate:
image

---

**Breadcrumb with RDFA Metadata:**
image
image

---

**Wrapping Breadcrumb:**
image

image

---

**Breadcrumb using custom Link components:**
image

---


trussworks-infra:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  
<tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against fd4aac985c39f2d0d3f72da5cbe3fa5d1fc22666
</p>

src/components/breadcrumb/BreadcrumbNav/BreadcrumbNav.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+export const DefaultBreadcrumb = (): React.ReactElement => (
+  <BreadcrumbNav>
+    <Breadcrumb label="Home" href="#" />
+    <Breadcrumb label="Federal Contracting" href="#" />
+    <Breadcrumb label="Contacting assistance programs" href="#" />
+    <Breadcrumb
+      label="Women-owned small business federal contracting program"
+      current
+    />
+  </BreadcrumbNav>
```
brandonlenz, line 26: I'm going to kind of word vomit out a few thoughts on this PR as it stands, starting here!

I like the developer api of this component as used here, and it aligns with the Props Interface requirement defined in the [github issue](https://github.com/trussworks/react-uswds/issues/337#issue-663147524) 

That said, in the case of metaData, there are optional props being passed to Breadcrumb that might make the developer api overly confusing. I would prefer those props to be implementation details. Similar to our FooterNav, for example, we could specify an object structure for the breadcrumb links, and pass those into a BreadcrumbNav rather than relying on `children`. That would require slight changes to the props interface, but I think it would be worth it.

While this approach gives us similar flexibility to the uswds implementation, it also allows the user to specify props on individual breadcrumbs that might be ignored when the BreadcrumbNav tries to enforce metaData on all of its children.

src/components/breadcrumb/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+export const Breadcrumb = (
+  props: BreadcrumbProps & JSX.IntrinsicElements['li']
+): React.ReactElement => {
+  const {
+    label,
+    href,
+    current = false,
+    withMetaData = false,
+    position = undefined,
```
brandonlenz, line 20: These props specifically do not need to be exposed to developers using our library. I'd like to encapsulate them as noted in my previous comment

src/components/breadcrumb/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,80 @@
...
+      typeof: assignMetaDataValue('ListItem'),
+    },
+    a: {
+      property: assignMetaDataValue('item'),
+      typeof: assignMetaDataValue('WebPage'),
+    },
+    span: {
+      property: assignMetaDataValue('name'),
+    },
+  }
```
brandonlenz, line 41: The main uncertainty I have is regarding meta data as a whole. The current implementation hardcodes the metadata as shown by uswds [here](https://github.com/uswds/uswds/blob/develop/src/components/breadcrumb/_breadcrumb--metadata.njk)

In #337 we have the following note: 

> withMetadata adds markup SEO support, includes <meta> tags and adds special attributes to the ol. Make this prop a boolean and only support JSON-LD for first go around.

I'm reading up on JSON-LD and RDFa now, but I suspect the implementation for metaData will look a bit different than it currently does.

brandonlenz: This was particularly helpful when learning about RDFA vs JSON-LD meta data: https://developers.google.com/search/docs/data-types/breadcrumb#json-ld_1

It seems that JSON-LD meta data should be built and included at the root level of an application, so I'm hesitant to attach it to this component. I would imagine a developer will use a specific JSON-LD npm module to include meta data in their site if they choose to go that way. In the name of keeping this repo lightweight with a focus on presentational components, I recommend leaving JSON-LD implementation to implementing teams.

We still include RDFA meta data tags here, though, since they are attached directly to the Breadcrumbs, and we'll be mirroring the uswds project by doing so (Their implementation linked in the above comment for reference)

suzubara:
this looks like a solid start, thanks for working on this! I want to ask if we should consider a more compositional approach to these components, to allow for more flexibility in case users need to pass other props down. for example something like this:
```
<BreadcrumbNav listProps={{ vocab: "http://schema.org" }}>
  <BreadcrumbItem><a href="#">Link</a></BreadcrumbItem>
  <BreadcrumbItem current>Current Link</BreadcrumbItem>
</BreadcrumbNav>
```

with `Breadcrumbs` taking props to pass to the `nav, ol` elements, and `BreadcrumbItem` taking props for the `li` element. I think we'd also want to make sure to support custom link elements and not just `a`.

brandonlenz:
> this looks like a solid start, thanks for working on this! I want to ask if we should consider a more compositional approach to these components, to allow for more flexibility in case users need to pass other props down. for example something like this:
> 
> ```
> <BreadcrumbNav listProps={{ vocab: "http://schema.org" }}>
>   <BreadcrumbItem><a href="#">Link</a></BreadcrumbItem>
>   <BreadcrumbItem current>Current Link</BreadcrumbItem>
> </BreadcrumbNav>
> ```
> 
> with `Breadcrumbs` taking props to pass to the `nav, ol` elements, and `BreadcrumbItem` taking props for the `li` element. I think we'd also want to make sure to support custom link elements and not just `a`.

@suzubara I'm with you on the compositional approach being more flexible. I originally pursued that path, My main reasoning for moving away from it was to be able to apply RDFA meta tags to the child components keeping parity with the [uswds example](https://github.com/uswds/uswds/blob/develop/src/components/breadcrumb/_breadcrumb--metadata.njk). If we go with a compositional approach we would lose control over adding RDFA meta data to the children (without knowing what types of elements will be received). I imagine going that route, we would not be able to support the `withRdfaMetaData` prop.

As far as passing in custom links, not just `a`, we face a similar problem of not being able to enforce the uswds class `usa-breadcrumb__link`. For that I could introduce a `BreadcrumbLink` component for standard use, giving us something that looks like this:

```javascript
<BreadcrumbBar listProps={{ vocab: "http://schema.org" }}>
  <Breadcrumb><BreadcrumbLink href="#">Link 1</BreadcrumbLink></Breadcrumb> //This combination of components could remain in a `LinkingBreadcrumb` component
  <Breadcrumb><CustomLink href="#">Link 2</CustomLink></Breadcrumb>
  <Breadcrumb>Current Link</Breadcrumb>
</BreadcrumbBar>
```

If there's some other React magic you'd recommend I use to get around these hiccups, please let me know. Curious on your thoughts.

haworku:
>  I think we'd also want to make sure to support custom link elements
✅ That should have been part of the initial issue description too as well. Whoops! Whatever solution we go with - let's write story examples in storybook with that use case to show how that would be handled.

brandonlenz:
> > I think we'd also want to make sure to support custom link elements
> 
> ✅ That should have been part of the initial issue description too as well. Whoops! Whatever solution we go with - let's write story examples in storybook with that use case to show how that would be handled.

Thanks @haworku and @suzubara. I'm working towards and interface with out-of-the-box components that directly mirror uswds implementations that looks like this:

```javascript
<BreadcrumbBar>
  <LinkingBreadcrumb href="#">Home</LinkingBreadcrumb>
  <LinkingBreadcrumb href="#">Federal Contracting</LinkingBreadcrumb>
  <LinkingBreadcrumb href="#">Contacting assistance programs</LinkingBreadcrumb>
  <Breadcrumb current>Women-owned small business federal contracting program</Breadcrumb>
</BreadcrumbBar>
```

The `Breadcrumb` component itself will be flexible, allowing child elements to be passed through. This means the onus for handling meta data will be entirely on the implementor, but I'll provide props to allow RDFA (and a storybook example for how to do so), so that developers can go that route if they so choose. For JSON-LD the onus is on the implementing developer anyways, so no changes there.

PR to be updated shortly.

suzubara (commented):
this is looking really good! I have 2 main pieces of feedback:

1) for components that pass intrinsic props onto one element, we've been following a pattern of using intersection types to add those props to the component (as opposed to nesting them inside of another prop). So for example, the `Breadcrumb` component would look like:

```
export const Breadcrumb = (props: BreadcrumbProps & JSX.IntrinsicElements['li']
```

I know for components that are passing intrinsic props to multiple elements, they need to be grouped together, but unless necessary I would suggest following the above pattern.

2) I think I don't quite understand the advantage of using the `LinkingBreadcrumb` component over using `Breadcrumb` and `BreadcrumbLink` outright.. I realize it's packaging those two components together, but the need to pass separate sets of props for the `li` and the `a` feels overcomplicated to me. Also it enforces a structure where all of the `children` are rendered inside the `a`, which would not facilitate the USWDS example with RDFa metadata where the `<meta>` tag is a sibling of `<a>`. My suggestion would be to:

- forgo `LinkingBreadcrumb`
- modify `BreadcrumbLink` so that it mirrors the existing `Link` component, applies the `usa-breadcrumb__link` but can also handle a custom element
- resulting in an example implementation like:

```
   <BreadcrumbBar listProps={{ ...rdfaMetadata.ol }}>
      <Breadcrumb { ...rdfaMetadata.li }>
        <BreadcrumbLink href="#" { ...rdfaMetadata.a }>
          <span property="name">Home</span>
        </BreadcrumbLink>
        <meta property="position" content="1" />
      </Breadcrumb>
      <Breadcrumb { ...rdfaMetadata.li }>
        <BreadcrumbLink<CustomLinkProps> to="#" asCustom={CustomLink} {{ ...rdfaMetadata.a }}>
          <span property="name">Federal Contracting</span>
        </BreadcrumbLink>
        <meta property="position" content="2" />
      </Breadcrumb>
      <Breadcrumb current { ...rdfaMetadata.li }>
        <span property="name">
          Women-owned small business federal contracting program
        </span>
        <meta property="position" content="4" />
      </Breadcrumb>
    </BreadcrumbBar>
```

src/components/breadcrumb/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,30 @@
...
+  className?: string
+  listItemProps?: JSX.IntrinsicElements['li']
+}
+interface BreadcrumbProps extends BaseBreadcrumbProps {
+  current?: boolean
+}
+
+export const Breadcrumb = (props: BreadcrumbProps): JSX.Element => {
+  const { children, current = false, className, listItemProps } = props
+  const classes = classnames(
```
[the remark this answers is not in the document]
brandonlenz: 🤯 That's perfect! Thanks for the tip!

src/components/breadcrumb/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,30 @@
...
+export interface BaseBreadcrumbProps {
+  children: React.ReactNode
+  className?: string
+  listItemProps?: JSX.IntrinsicElements['li']
+}
+interface BreadcrumbProps extends BaseBreadcrumbProps {
+  current?: boolean
+}
+
+export const Breadcrumb = (props: BreadcrumbProps): JSX.Element => {
```
brandonlenz, line 13: Good catch @suzubara on adding the JSX intrinsic props. Adding this comment here to remind myself.

src/components/breadcrumb/BreadcrumbLink/BreadcrumbLink.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+
+interface BreadcrumbLinkProps {
+  children: React.ReactNode
+  href: string
+  anchorProps?: JSX.IntrinsicElements['a']
+  spanProps?: JSX.IntrinsicElements['span']
```
haworku, line 7: Maybe redundant prop? Not used in component

src/components/breadcrumb/LinkingBreadcrumb/LinkingBreadcrumb.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import { BaseBreadcrumbProps, Breadcrumb } from '../Breadcrumb/Breadcrumb'
+import { BreadcrumbLink } from '../BreadcrumbLink/BreadcrumbLink'
+
+interface LinkingBreadcrumbProps extends BaseBreadcrumbProps {
+  href: string
+  anchorProps?: JSX.IntrinsicElements['a']
+  spanProps?: JSX.IntrinsicElements['span']
```
haworku, line 8: Maybe redundant prop? Not used in component

haworku (commented):
Things look good to me! 

**A thought about a way to cut scope for this PR regarding`<LinkingBreadcrumb />`:**
This component is a nicety, not the core behavior of breadcrumb. Consumers could adapt their own version  suited to their use case. They know what types of links they have and exactly what attributes they need.

If at the end of the day we feel weirdness or uncertainty about this component, or if refactoring things to be more like `<Link>` feels like a heavy lift,  I could see us moving this pattern into the Storybook examples, rather than exporting from the lib at this point. We could backlog a story about making a more superpowered `<BreadcrumbLink>` for a later date. Not trying to dissuade you @brandonlenz at all, just want to say I feel like we have the core work for Breadcrumb done (which is great!).

src/components/breadcrumb/BreadcrumbLink/BreadcrumbLink.tsx
brandonlenz: Yep this was a lingering remnant of a previous iteration. It's being removed 😄

src/components/breadcrumb/LinkingBreadcrumb/LinkingBreadcrumb.tsx
brandonlenz: Yea per your advice and @suzubara I'm killing the LinkingBreadcrumb. It felt like a nice candidate for a compositional convenience component for developers who wanted default functionality with low complexity. At the end of the day it doesn't really fit with the convention of this library, and ultimately adds a maintenance burden to maintainers of react-uswds.

haworku: Sounds good.  And i hear you. Its always a bit weird (for me personally) to juggle the difference between the library and the code we write day to day.

src/components/breadcrumb/Breadcrumb/Breadcrumb.tsx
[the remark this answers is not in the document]
brandonlenz: 8abb76f731d22e7a13bebad6b0739191f15fe229

src/components/breadcrumb/BreadcrumbLink/BreadcrumbLink.tsx
```diff
@@ -0,0 +1,23 @@
...
+): React.ReactElement {
+  const { className } = props
+  const classes = classnames(className, 'usa-breadcrumb__link')
+
+  if (isCustomProps(props)) {
+    return <Link<FCProps> {...props} className={classes} variant="unstyled" />
+  }
+
+  return <Link {...props} className={classes} variant="unstyled" />
+}
```
brandonlenz, line 23: I've now got `BreadcrumbLink` wrapping `Link`. I did not want to re-implement any logic from the `Link` component, to prevent duplicated maintenance in the event the Link implementation were to change. 

I was hoping there would be a way to wrap the Link component's signature completely in more of a HOC fashion, and simply inject the `className`. Due to Link's clever overloading involving the optional generic, I had to mirror at least that part here.

src/components/breadcrumb/BreadcrumbLink/BreadcrumbLink.tsx
brandonlenz: Gone in 3b391bade03925a0ada15f4a2dd3881fb808dd3d

suzubara (approved):
OK, I have verified the markup matches the examples in the [USWDS docs](https://designsystem.digital.gov/components/breadcrumb/) _and_ tested this component out on MilMove! everything works as expected. We should use the stray `eslint-disable` lines, as I don't think they are needed, but after that I think this is good to 🚢  🎉

src/components/breadcrumb/Breadcrumb/Breadcrumb.test.tsx
```diff
@@ -0,0 +1,37 @@
+/*  eslint-disable jsx-a11y/anchor-is-valid */
```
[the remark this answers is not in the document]
brandonlenz: a20556a

suzubara (approved):

Issue #312, opened 2020-07-13
[feat] Include compiled `/lib` files in the repo


suzubara:
**Does your feature request relate to a specific USWDS feature?**
No

**Is your feature request related to a problem? Please describe.**
Right now, compiled library files (located in `/lib`) are ignored by git. The most apparent problem from this is that if someone wanted to install a specific branch or tag version that was not published to npm, it would not work out of the box because the library files are not compiled. It seems to be a common practice for library repos to include `/lib` in the checked-in code, so perhaps we should as well.

**Describe the solution you'd like**
- Remove `/lib` from `.gitignore`
- Add checks (probably in `dangerfile.ts`) that require PRs to include updated `lib` files before merging (this will also mean branches must be up-to-date with the main branch)
- Optionally: maybe warn if `lib` files are not updated as a pre-push hook? If someone is testing a WIP branch, it will be important to include compiled files on every push, even before merging to master.

**Describe alternatives you've considered**
- Don't support installing from github -- this is less than ideal, since it makes testing implementations of WIP features difficult
- Use [`npm prepare`](https://docs.npmjs.com/misc/scripts) to build the library on install -- also not ideal, since we'd have to add it explicitly during feature development, and remember to remove it before merging/releasing. It means that implementation code bases are responsible for installing dev dependencies and compiling the code, and this can also cause latency issues in CI.

**Additional context**
This is semi-related to our current inability to automatically publish WIP to npm or Github. We can't auto-publish to NPM because 2FA is required, and NPM does not have any way around that right now. Github package repo does not (yet?) support publishing to multiple release streams (such as `@next`). For now, we should have a low friction way for someone to install a WIP branch of this library, and doing that from the Github repo directly is the easiest way right now.

PR #842, opened 2021-02-03
build: Update yarn publish scripts, move dependencies to devDependencies

suzubara:
# Summary

This PR makes some minor changes to our [npm lifecycle scripts](https://docs.npmjs.com/cli/v6/using-npm/scripts#life-cycle-scripts):
- Use `prepare` to build the library -- this will build the `/lib` directory before publishing, and on local `yarn install`, and before installing from a git tag, resolving https://github.com/trussworks/react-uswds/issues/312
- Adds `prepublishOnly` to ensure tests and linters are passing before publishing

It also moves dependencies into `devDependencies` and adds `uswds` to `peerDependencies`.

I also updated the README because I discovered an issue if you try to use JS components (such as the ComboBox or DatePicker) but are also importing the full USWDS JS in your project (`import 'uswds'`). This will result in the UI initializing twice and it's very buggy.

## Related Issues or PRs

Closes https://github.com/trussworks/react-uswds/issues/312

## How To Test

This PR should be tested by installing this branch on applications and making sure use of ReactUSWDS components still works as it did before. To do this, edit your `package.json` file and point the library to this branch:

```
"@trussworks/react-uswds": "https://github.com/trussworks/react-uswds#sr-build-updates"
```

Then run `yarn install` and start your app as usual. The install might take a little longer than usual, because it has to compile the library since it's not using a published version. You can also verify the correct version is installed by looking in your `yarn.lock` file. The `@trussworks/react-uswds` entry should look like this, with "resolved" pointing to this repo instead of the Yarn package registry:

```
"@trussworks/react-uswds@https://github.com/trussworks/react-uswds#sr-build-updates":
  version "1.10.0"
  resolved "https://github.com/trussworks/react-uswds#3626600feb13f9e5e58a86e6bfa72e450f8123fe"
```


trussworks-infra:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 4e87515cf64dedddc6dd2fde3c5a35604e970d5e
</p>

package.json
```diff
@@ -58,14 +59,7 @@
   "homepage": "https://github.com/trussworks/react-uswds#readme",
   "peerDependencies": {
     "react": "^16.x",
-    "react-dom": "^16.x"
-  },
-  "dependencies": {
```
brandonlenz, line 63: I'm trying to be better versed on the different dependency types here, looking at https://classic.yarnpkg.com/en/docs/dependency-types/

> #### Dependencies
> These are your normal dependencies, or rather ones that you need when running your code (e.g. React or ImmutableJS).

> #### devDependencies
> These are your development dependencies. Dependencies that you need at some point in the development workflow but not while running your code (e.g. Babel or Flow).

> #### peerDependencies
> Peer dependencies are a special type of dependency that would only ever come up if you were publishing your own package.
> Having a peer dependency means that your package needs a dependency that is the same exact dependency as the person installing your package. This is useful for packages like react that need to have a single copy of react-dom that is also used by the person installing it.

Is the reasoning that things like `react` and `uswds` only need to be `peerDependencies` that react-uswds itself doesn't "run", so we expect only the implementing App to include it?

---

Second question, mostly a point of curiosity: 

I've used some FE component libraries that include all their own styles transitively before. For react-uswds it looks like we require the implementing project to include USWDS styles on its own as well. Is this intentional because the assumed use-case is for our components to supplement a project already using USWDS in some capacity?

brandonlenz:
Adding a note here as well to say that I built a Create React App and added this branch as a dependency in my `package.json` and it worked perfectly ✅

brandonlenz (approved):

package.json
```diff
@@ -58,14 +59,7 @@
...
-    "classnames": "^2.2.6",
-    "dotenv": "^8.2.0",
-    "happo-plugin-storybook": "^2.5.3",
-    "react": "^16.10.0",
-    "react-dom": "^16.10.0",
+    "react-dom": "^16.x",
     "uswds": "2.8.0"
```
brandonlenz, line 63: Do we want this peer dependency to specify exactly `2.8.0`, or do we want to allow consumers who might be on more current version of uswds to also use react-uswds without warnings?

haworku (approved):
This worked for me as well pulling into an existing app. Thanks for this PR its important maintenance work forgot about `prepublish` and its good to have that in place in case a local release branch somehow gets in a weird state

package.json
haworku: @brandonlenz The answer to your question (in my understanding at least) is yes.

PR #846, opened 2021-02-03
feat: Add nav variant to Link component

SirenaBorracha:
# Summary
Add a `nav` option to the `variant` property to style the existing `Link` component with className `usa-nav__link`

## Related Issues or PRs

#166 

## How To Test

```yarn test```

```yarn storybook```
Viewable from Typography > Link > Nav Link

### Screenshots (optional)

Storybook
Screen Shot 2021-02-03 at 2 49 27 PM

### CODE REVIEW

File: src/components/Link/Link.tsx [hunk 1]
```diff
@@ -40,16 +40,17 @@ function linkClasses<T>(
 ): string | undefined {
   const unstyled = variant === 'unstyled'
   const isExternalLink = variant === 'external'
+  const isNavLink = variant === 'nav'
 
   return unstyled
     ? className
```
haworku, line 46: The way the [classnames](https://github.com/JedWatson/classnames#usage/) package works is it already conditionally includes the css. This means  we don't need to add another ternary here.  Check out the linked docs if you've never used this package before, its pretty convenient.  

See if you could get this still passing tests but following a pattern that's more similar to this pseudocodey example:
``` javascript
return unstyled?
    className // if its unstyled, only use classes passed in by props
     :  
   classnames( // otherwise use...
       'usa-link', // any classes required for component
       { 'random-class': true},  // and  any conditionally included CSS
       classname  // and also add the classes passed in by props
    )
```

We add `className` last because we always want the styles that come in through props to be applied on top of other styles. There are several examples of this throughout react-uswds codebase to check out too if you want.

haworku:
Swap with `usa-link__nav`

SirenaBorracha:
@suzubara do we use Happo differently with this project than on MM?

example/tsconfig.json
```diff
@@ -17,7 +17,8 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
-    "jsx": "react"
+    "jsx": "react",
+    "noFallthroughCasesInSwitch": true
```
brandonlenz, line 21: Just double checking, it sounds like @suzubara is aware and okay with this change being in the PR. Is that the case?

brandonlenz (commented):
This looks great! 

One question to double check on the change to the `tsconfig.json`, otherwise I'm good to approve!

Congrats on your first react-uswds PR 💯

example/tsconfig.json
SirenaBorracha: Yay! Yes, Suz is aware, here's the [Slack conversation](https://trussworks.slack.com/archives/C0122MPMXL3/p1612301170007100)

brandonlenz (approved):

PR #867, opened 2021-02-08
fix: prevent dependabot verison updates (+ husky fixes)

brandonlenz:
# Summary

Dependabot recently performed an update to React for a major version change. We would like to prevent that in the future.

Additionally, uswds should be updated manually, so it has been added to the dependabot ignore list.


.github/dependabot.yml
```diff
@@ -15,25 +15,29 @@ updates:
...
-          - '^6.0.21'
+          - '^6.1.5'
       - dependency-name: '@storybook/addon-docs'
         versions:
-          - '^6.0.21'
+          - '^6.1.5'
       - dependency-name: '@storybook/addon-viewport'
         versions:
-          - '^6.0.21'
+          - '^6.1.5'
```
brandonlenz, line 39: While I'm at it, I bumped these versions to what we're currently using (since that's what they were before).

.github/dependabot.yml
```diff
@@ -15,25 +15,29 @@ updates:
     ignore:
       - dependency-name: react
         versions:
-          - '^16.11.0'
+          - '>=18.0.0'
```
ahobson, line 18: Why did dependabot not ignore react 17 with this rule? Without reading the docs, it seems like it should have.

brandonlenz: From my reading [here](https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#ignore), the `versions` option uses npm's semver to parse version ranges that should be ignored. `^16.11.0` is equivalent to a range of versions `>=16.11.0` and `<17.0.0`.

What we're unsure about is why this only just now happened. React 17 was released in October of last year. The timing is suspect because we also recently moved `react` to `devDependency`. From my limited research, it sounds like dependabot should treat `devDependencies` with the same config-derived logic, but I also didn't find an answer that felt definitive on that.

ahobson: From what I can tell from [the docs](https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#allow) it seems we were only allowing development dependencies? That doesn't seem to make sense with the PRs we see, though. It does seem like we want `dependency-type: direct` though?

suzubara: @ahobson you're right, I think that is what caused dependabot to upgrade React to 17. tbh this project should only have development & peer dependencies to begin with, but changing `dependency-type: direct` shouldn't hurt.

brandonlenz: @suzubara I'll update that when I fix conflicts here  👍

brandonlenz: 43222c2

suzubara (commented):
@brandonlenz I'm not seeing the `uswds` dependency in the diff anymore, did it get removed by accident? I think once that's added back this is good to go!

brandonlenz:
> @brandonlenz I'm not seeing the `uswds` dependency in the diff anymore, did it get removed by accident? I think once that's added back this is good to go!

Nice catch! Added it back with https://github.com/trussworks/react-uswds/pull/867/commits/ac1025265e46397b5ddcc3185d2a88ee6678a231

suzubara (dismissed):

.husky/.gitignore
```diff
@@ -0,0 +1 @@
+_
```
brandonlenz, line 1: @suzubara heads up, in order to commit after the husky version update in https://github.com/trussworks/react-uswds/pull/867/commits/1d091e847a28b36c2da38f13d5ea92e6542be880 I just had to install husky which added this file.

brandonlenz:
Re-running ci because the happo diff, while concerning, shouldn't be caused by this type of PR.
image

Hopefully it's just a happo fluke.

.github/dependabot.yml
```diff
@@ -11,26 +11,30 @@ updates:
...
         versions:
-          - '^16.10.0'
+          - '>=18.0.0'
       - dependency-name: '@types/react-dom'
         versions:
-          - '^16.10.0'
+          - '>=18.0.0'
       - dependency-name: '@storybook/react'
         versions:
           - '^6.0.21'
```
ahobson, line 33: Am I reading this right that we are ignoring versions of `@storybook/react` that are compatible with `6.0.21`? Same for the `addon-essentials` below? I'm pretty sure I'm the one that added those incorrectly.

brandonlenz: [Based on the docs](https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#ignore), this means we will ignore all versions of `@storybook/react` between `6.0.21` and `7.0.0`. I imagine we probably want the inverse of that? That's why I changed our React version to a `>=` syntax since we are okay with minor and incremental version changes, but we want to ignore major version changes. We just have to remember to update these ignores for every major version migration if we want to keep getting minor updates.

ahobson: Yeah, that's what I think, too. I think I misunderstood when I originally set it up. Should we go ahead and fix everything at the same time?

brandonlenz: That's perfectly reasonable! I know storybook tends to be a bit more brittle, so if we do run into issues with minor version updates after this PR, we can always re-evaluate.

brandonlenz: 2646e3a

.husky/.gitignore
brandonlenz: Actually, I'm not so sure I did everything required here. In https://github.com/trussworks/react-uswds/pull/877 the patch notes mentions updating from husky 4 to 5 being more involved:

>### Migrating
>Important Husky v5 brings a lot of improvements but is also very different from v4. Git hooks won't work if you only upgrade husky dependency, existing config needs to be migrated too.

>The best way to switch to v5 is to follow the new installation instructions and migrate existing hooks command using husky add.

I found https://github.com/typicode/husky-4-to-5 via https://github.com/typicode/husky/issues/854#issuecomment-776126582

brandonlenz: @suzubara @ahobson Okay, so the husky 5 docs don't cover the implementation of how husky 5 works very well (they no longer use package.json, and instead rely on scripts in the .husky directory). Everything is updated in accordance with the [Husky 5 docs](https://typicode.github.io/husky/#/?id=install) now via 58b6d22

ahobson (approved):
I think this looks good to me. Thank you so much for chasing down all the comments and suggestions

ahobson (approved):

Issue #342, opened 2020-07-21
New Component: Tooltip


haworku:
# [Tooltip](https://designsystem.digital.gov/components/tooltip/)
A tooltip is a short descriptive message that appears when a user hovers or focuses on an element.

## Description
 _Any element_ with the class name `usa-tooltip` and a `title` attribute will become a tooltip. 

## Requirements (proposed)
**Prop Interface**
``` 
TooltipProps = {
  label: string
  position: 'top' | 'bottom' | 'left' | 'right'
} 
```
**More Details**
- `position` prop used as `data-position` on tooltip element
- `label` is used as `title` on tooltip element
-  Our tooltip component then takes any element (perhaps as children or as a prop or as hook, developer choice) and adds these attributes.
- The tooltip popover is generated with javascript - see [uswds tooltip js](https://github.com/uswds/uswds/blob/b3a6d7c0656a26edd7329f47f05a2908069c2de2/src/js/components/tooltip.js). We may want to reach for an existing tooltip library or utility rather than entirely re-implementing. 
- Tooltips are progressive enhancements for the title attribute, and will display as the title attribute if the component doesn’t initialize.

PR #871, opened 2021-02-11
feat: Create Tooltip component  

noelledusahel:
# Summary
This story creates a react component modeled after the [USWDS Tooltip](https://designsystem.digital.gov/components/tooltip/)

## Related Issues or PRs

https://github.com/trussworks/react-uswds/issues/342

## How To Test
In storybook see Components > Tooltips

``` 
yarn test
yarn storybook
```


trussworks-infra:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 86ed9068759cc61d17123cf371f2570f91a9f9c5
</p>

src/components/Tooltip/Tooltip.test.tsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+
+import { Tooltip } from './Tooltip'
+
+describe('Tooltip component', () => {
```
suzubara, line 6: Could we add some tests that test:
- the Tooltip is hidden by default
- handling showing/hiding the Tooltip on the relevant JS events (mouseenter, focus, mouseleave, blur, keydown)

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,229 @@
...
+interface TooltipProps {
+  label: string
+  position?: 'top' | 'bottom' | 'left' | 'right'
+  className?: string
+  children: ReactElement
+}
+
+const TRIANGLE_SIZE = 5
+const SPACER = 2
+const tooltipID = `tooltip-${Math.floor(Math.random() * 900000) + 100000}`
```
suzubara, line 18: I _think_ this should probably be initialized inside the `Tooltip` component, probably inside a `useEffect` hook so it just happens once on initial mount.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,229 @@
...
+  return (
+    rect.top >= 0 &&
+    rect.left >= 0 &&
+    rect.bottom <= (win.innerHeight || docEl.clientHeight) &&
+    rect.right <= (win.innerWidth || docEl.clientWidth)
+  )
+}
+
+export const Tooltip = (
+  props: TooltipProps & JSX.IntrinsicElements['span']
```
suzubara, line 35: see my comment about changing the signature of this component, I think this should be `JSX.IntrinsicElements['button']`

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,229 @@
...
+    if (
+      triggerElementRef.current &&
+      tooltipBodyRef.current &&
+      wrapperRef.current
+    ) {
+      const tooltipTrigger = triggerElementRef.current
+      const tooltipBody = tooltipBodyRef.current
+      const wrapper = wrapperRef.current
+      tooltipTrigger.setAttribute('aria-describedby', tooltipID)
+      tooltipBody.setAttribute('id', tooltipID)
```
suzubara, line 65: I think some of these attributes don't change when `isVisible` changes, and could be set declaratively in the returned markup instead

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,229 @@
...
+  const tooltipClasses = classnames('usa-tooltip__body', {
+    'is-set': isVisible,
+    'usa-tooltip__body--top': position === 'top',
+    'usa-tooltip__body--bottom': position === 'bottom',
+    'usa-tooltip__body--right': position === 'right',
+    'usa-tooltip__body--left': position === 'left',
+    'is-visible': isVisible,
+  })
+
+  useEffect(() => {
```
noelledusahel, line 55: I placed the position logic within a useEffect hook because certain calculations like `tooltipBodyWidth` were being generated when the tooltip was offscreen, causing issues with `marginLeft` in `positionLeft`

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,229 @@
...
+        window.getComputedStyle(tooltipBody).getPropertyValue('height'),
+        10
+      )
+      const leftOffset = tooltipTrigger.offsetLeft
+      const tooltipBodyWidth = tooltipBodyRef.current.offsetWidth
+      const adjustHorizontalCenter = tooltipWidth / 2 + leftOffset
+      const adjustToEdgeX = tooltipWidth + TRIANGLE_SIZE + SPACER
+      const adjustToEdgeY = tooltipHeight + TRIANGLE_SIZE + SPACER
+
+      /**
```
suzubara, line 99: I think it would be good to take advantage of React's declarative nature here, and set these attributes (style, classes) using `useState` and pass them to the element(s) inside of the markup instead of manipulating the HTML elements directly

suzubara (changes requested):
This is great work! I have some suggestions/comments:

- Please remove the `package-lock.json` file - this was likely added as a result of using `npm install` but we should be consistent in using `yarn` on this project.

- I think the `button` element with the `title` attribute should be generated by this component, since the USWDS component follows a strict convention (must be a button element and tooltip body is populated with the value of the `title` attribute) instead of using `props.children`. This would mean usage would look more like: `<Tooltip position="top" title="Top" />` and I think intrinsic props should be for the `button` element (since the wrapper `span` in USWDS is not customizable other than adding classes, so we could expose a `wrapperClasses` prop for that)

- I think some additional event handlers are needed, the tooltip should show on focus/mouseenter and hide on mouseleave, blur, keydown according to the [source JS](https://github.com/uswds/uswds/blob/b3a6d7c0656a26edd7329f47f05a2908069c2de2/src/js/components/tooltip.js#L252)

brandonlenz:
> * I think the `button` element with the `title` attribute should be generated by this component, since the USWDS component follows a strict convention (must be a button element and tooltip body is populated with the value of the `title` attribute) instead of using `props.children`. This would mean usage would look more like: `<Tooltip position="top" title="Top" />` and I think intrinsic props should be for the `button` element (since the wrapper `span` in USWDS is not customizable other than adding classes, so we could expose a `wrapperClasses` prop for that)

@suzubara I advised to follow a flexible approach here. While all the uswds examples are buttons, they make a specific mention of using it on more than just buttons in the docs:
image

I could definitely see people wanting to wrap a custom link in a tooltip, which I thought we might want to allow.

suzubara:
> > * I think the `button` element with the `title` attribute should be generated by this component, since the USWDS component follows a strict convention (must be a button element and tooltip body is populated with the value of the `title` attribute) instead of using `props.children`. This would mean usage would look more like: `<Tooltip position="top" title="Top" />` and I think intrinsic props should be for the `button` element (since the wrapper `span` in USWDS is not customizable other than adding classes, so we could expose a `wrapperClasses` prop for that)
> 
> @suzubara I advised to follow a flexible approach here. While all the uswds examples are buttons, they make a specific mention of using it on more than just buttons in the docs:
> image
> 
> I could definitely see people wanting to wrap a custom link in a tooltip, which I thought we might want to allow.

Huh okay, I missed that!! What if we exposed a custom component prop (like we do for the `Link` component) in that case? I'm mostly trying to avoid the possibility of the `title` attribute not being the same value as the Tooltip label, and also not require the user to manually add the CSS classes to the child component

ahobson:
Anyone have experience testing tooltips in storybook? All I found was using this [pseudo state Addon](https://storybook.js.org/addons/@ergosign/storybook-addon-pseudo-states-react), but I don't know if it even works as advertised.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,229 @@
...
+      data-testid="tooltip"
+      ref={wrapperRef}
+      className="usa-tooltip"
+      {...spanProps}
+      onMouseEnter={() => {
+        setVisible(true)
+      }}
+      onMouseLeave={() => {
+        setVisible(false)
+      }}>
```
brandonlenz, line 212: I just realized these (in addition to the focus triggers etc that Suz mentioned) should be on the child (trigger) element. Right now, if the child element has a huge margin, the tooltip will trigger on mousover of the child element's margin as well, because the span encapsulates that as well.

noelledusahel: yep im seeing this today

brandonlenz:
> Huh okay, I missed that!! What if we exposed a custom component prop (like we do for the `Link` component) in that case? I'm mostly trying to avoid the possibility of the `title` attribute not being the same value as the Tooltip label, and also not require the user to manually add the CSS classes to the child component

With Link as an example to follow, that will work! I'm personally a bigger fan of the syntax of just wrapping an element with a tooltip (seems to be common in FE component libraries like [material-ui](https://material-ui.com/components/tooltips/)), and the syntax is familiar to use. That may be because I don't think of a tooltip as a standalone component that should be responsible for creating anything internally, but rather something that exists as an enhancement to other components.

Internally we'll be doing `React.cloneElement` or `React.createElement` either way, right?. Both _should_ allow us to set the 'usa-tooltip__trigger' class internally though. Is that the class you were talking about with the child component?

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,229 @@
...
+  position?: 'top' | 'bottom' | 'left' | 'right'
+  className?: string
+  children: ReactElement
+}
+
+const TRIANGLE_SIZE = 5
+const SPACER = 2
+const tooltipID = `tooltip-${Math.floor(Math.random() * 900000) + 100000}`
+
+function isElementInViewport(
```
brandonlenz, line 20: For some polish, I think this method is a good candidate to add to a `utils`-esque place. Suz has some good examples of that with DatePicker and the new [FileInput](https://github.com/trussworks/react-uswds/pull/900/files#diff-fed0c73187c20beede2d59813f693438b77907b5bbbfc1a54b3d59df4b0a6cfb)

suzubara:
> With Link as an example to follow, that will work! I'm personally a bigger fan of the syntax of just wrapping an element with a tooltip (seems to be common in FE component libraries like [material-ui](https://material-ui.com/components/tooltips/)),, and the implementation syntax is familiar to use. That may be because I don't think of a tooltip as a standalone component that should be responsible for creating anything internally, but rather something that exists as an enhancement to other components.

Yeah I am not solidly on one side or the other here. if we were building a super generic & flexible Tooltip component I would definitely agree with you about using `props.children` and just wrapping, but that can sometimes introduce its own set of complexities. Since the USWDS Tooltip is very basic, and strictly just populates the Tooltip contents with plain text from the `title` attribute, I was focusing on the ideal implementation of just that.
 
> Internally we'll be doing `React.cloneElement` or `React.createElement` either way, right?. Both _should_ allow us to set the 'usa-tooltip__trigger' class internally though. Is that the class you were talking about with the child component?

Yup that's the class I meant, agreed.

src/components/Tooltip/Tooltop.stories.tsx
```diff
@@ -0,0 +1,84 @@
...
+)
+
+export const CustomComponent = (): React.ReactElement => {
+  type MockTooltipProps = React.PropsWithChildren<{
+    to: string
+    className?: string
+  }> &
+    JSX.IntrinsicElements['a'] &
+    React.RefAttributes<HTMLAnchorElement>
+  const CustomLink: React.ForwardRefExoticComponent<MockTooltipProps> = React.forwardRef(
```
brandonlenz, line 61: @suzubara curious on your thoughts here. We had both not used forwardRef before. This seems like the suggested way to get a ref to a child component's html element. I had seen some discussion about forwarding refs on DatePicker and FileInput so I wanted to call this one out for discussion specifically.

For people who just want to wrap an existing component, this _does_ seem cumbersome as they would need to expose a ref in order to use this component.

Would you recommend going this route or a different appraoch?

src/components/Tooltip/Tooltop.stories.tsx
```diff
@@ -0,0 +1,84 @@
...
+export const tooltipLeft = (): React.ReactElement => (
+  <div className="margin-4">
+    <Tooltip position="left" label="Left">
+      Show on left
+    </Tooltip>
+  </div>
+)
+
+export const CustomComponent = (): React.ReactElement => {
+  type MockTooltipProps = React.PropsWithChildren<{
```
brandonlenz, line 55: Since these are the props for the `CustomLink` created below, should `MockTooltipProps` be called `CustomLinkProps` or similar?

brandonlenz (commented):
A couple of suggestions, and wanted to draw attention to the forwardRef to get some more feedback on it where my own knowledge falls short.

src/components/Tooltip/Tooltip.test.tsx
haworku: More tests were added in 112fcd83166bac4c0d82dc579503d5823a3c10cd. Thanks!

Please still also add these tests (for keyboard navigation, `hover` doesn't cover these). 
- hides on blur
- shows on focus

src/components/Tooltip/Tooltip.tsx
haworku: very much agree about avoiding `element.style` since we have JSX `className`.  It is a refactor though and the way the code is written seems fine to me for a first pass (since it just copies uswds). We could backlog an issue about refactoring this. @suzubara thoughts?

suzubara: yes, I'm okay with that!

src/components/Tooltip/Tooltip.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+    expect(queryByTestId('tooltipWrapper')).toBeInTheDocument()
+    expect(queryByTestId('triggerElement')).toBeInTheDocument()
+    expect(queryByTestId('tooltipBody')).toBeInTheDocument()
+  })
+
+  it('hides tooltip by default', () => {
+    const { getByTestId } = render(
+      <Tooltip label="Click me">
+        <button>My Tooltip</button>
+      </Tooltip>
```
brandonlenz, line 23: Sorry, I didn't see this on my first pass. Looks like the unit tests are still using the old implementation syntax for the tooltip. Can we update these to match how Tooltip is used in storybook now that it implements a button internally?

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,369 @@
...
+  onMouseOver: (e: React.MouseEvent) => void
+  onFocus: (e: React.FocusEvent) => void
+  onMouseLeave: (e: React.MouseEvent) => void
+  onBlur: (e: React.FocusEvent) => void
+  onKeyDown: (e: React.KeyboardEvent) => void
+  children: ReactNode
+} & T
+
+interface WithCustomTooltipProps {
+  asCustom: // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
suzubara, line 34: They're not 100% the same but I used [this Formik source code](https://github.com/formium/formik/blob/08d41a7d78290c272ac484a29828f54b74ec4870/packages/formik/src/Field.tsx#L33) as an example of typing a polymorphic component prop.

Other references that have been helpful in figuring this out:
- https://blog.andrewbran.ch/polymorphic-react-components/
- https://react-typescript-cheatsheet.netlify.app/docs/advanced/patterns_by_usecase
- https://github.com/kripod/react-polymorphic-box
- https://github.com/typescript-cheatsheets/react/issues/167#issuecomment-751347673

suzubara (commented):
Do we want to add a `wrapperClasses` prop that mirrors the `data-classes` attribute used by the [USWDS Tooltip](https://designsystem.digital.gov/components/tooltip/)?
> JavaScript generated most of the tooltip’s markup. Apply utility classes to any tooltip’s wrapping element, by including them inside a data-classes attribute, separated by spaces.

noelledusahel:
> Do we want to add a `wrapperClasses` prop that mirrors the `data-classes` attribute used by the [USWDS Tooltip](https://designsystem.digital.gov/components/tooltip/)?
> 
> > JavaScript generated most of the tooltip’s markup. Apply utility classes to any tooltip’s wrapping element, by including them inside a data-classes attribute, separated by spaces.

@suzubara  on the USWDS side data-classes gets applied on the tooltip trigger. are you reffering to a wrapperClass that applies classes to the wrapper span or to the tooltip trigger?

noelledusahel:
> > Do we want to add a `wrapperClasses` prop that mirrors the `data-classes` attribute used by the [USWDS Tooltip](https://designsystem.digital.gov/components/tooltip/)?
> > > JavaScript generated most of the tooltip’s markup. Apply utility classes to any tooltip’s wrapping element, by including them inside a data-classes attribute, separated by spaces.
> 
> @suzubara on the USWDS side data-classes gets applied on the tooltip trigger. are you referirng to a wrapperClass that applies classes to the wrapper span or to the tooltip trigger?

Actually I read further, and see the classes get applied to the wrapper

suzubara:
> > > Do we want to add a `wrapperClasses` prop that mirrors the `data-classes` attribute used by the [USWDS Tooltip](https://designsystem.digital.gov/components/tooltip/)?
> > > > JavaScript generated most of the tooltip’s markup. Apply utility classes to any tooltip’s wrapping element, by including them inside a data-classes attribute, separated by spaces.
> > 
> > 
> > @suzubara on the USWDS side data-classes gets applied on the tooltip trigger. are you referirng to a wrapperClass that applies classes to the wrapper span or to the tooltip trigger?
> 
> Actually I read further, and see the classes get applied to the wrapper. I'm curious whether you think it makes sense to use the same terms as USWDS, and use the prop name data-classes rather than wrapperClasses

I don't feel super strongly about the exact name, but my inclination is to use something explicit, like `wrapperClasses`. The reason is because `data-` attributes are not specific to React, and in the case of USWDS they are being used to easily interface between vanilla HTML & JS. So I would recommend not naming a prop anything like `data-` because it will probably not be clear that it's a React prop.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,347 @@
...
+
+export type CustomTooltipProps<T> = TooltipProps<T> & WithCustomTooltipProps<T>
+
+export function isCustomProps<T>(
+  props: DefaultTooltipProps | CustomTooltipProps<T>
+): props is CustomTooltipProps<T> {
+  return 'asCustom' in props
+}
+
+const tooltipID = `tooltip-${Math.floor(Math.random() * 900000) + 100000}`
```
suzubara, line 36: Because this `tooltipID` is defined outside of the `Tooltip` component, it's the same value for all instances of a Tooltip used within a single app. This should be defined inside of the `Tooltip` component in a `useRef` container:
```
const tooltipID = useRef(`tooltip-${Math.floor(Math.random() * 900000) + 100000}`)

// to access:
tooltipID.current
```

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,347 @@
...
+  useState,
+} from 'react'
+import classnames from 'classnames'
+import { isElementInViewport } from './utils'
+
+type TooltipProps<T> = {
+  label: string
+  position?: 'top' | 'bottom' | 'left' | 'right' | undefined
+  className?: string
+  dataClasses?: string
```
suzubara, line 18: so I tried this out and actually React shows an error in the console because of a potential naming conflict with native `data-` attributes. I think this prop should be renamed to `wrapperClasses`.

image

suzubara (commented):
I tested this out on MilMove and the Tooltip works great! I left two minor comments that I think should be fixed before merging. We should also make sure to create an issue to refactor the positioning code so it uses React state to store class/position, instead of JS DOM APIs.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,347 @@
...
+      },
+      children
+    )
+
+    return (
+      <span
+        data-testid="tooltipWrapper"
+        ref={wrapperRef}
+        className="usa-tooltip"
+        {...customProps}
```
suzubara, line 275: `customProps` are passed into the `triggerElement` below, so they should not be repeated on this wrapper span

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,347 @@
...
+    }
+
+    useTooltip(triggerElementRef, position)
+
+    return (
+      <span
+        data-testid="tooltipWrapper"
+        ref={wrapperRef}
+        className="usa-tooltip"
+        {...spanProps}
```
suzubara, line 314: I think the remaining props are meant to be passed onto the button element, rather than the wrapper span (this should mirror the custom implementation above)

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,347 @@
...
+  useRef,
+  useState,
+} from 'react'
+import classnames from 'classnames'
+import { isElementInViewport } from './utils'
+
+type TooltipProps<T> = {
+  label: string
+  position?: 'top' | 'bottom' | 'left' | 'right' | undefined
+  wrapperclasses?: string
```
noelledusahel, line 17: i removed className prop, as i didn't see the purpose of having both className and wrapperclasses. may have lost some perspective  for why we'd need a separate prop since starting this ticket so your input would be appreciated @suzubara

suzubara: The `className` prop is meant to allow a user to pass a custom CSS class to the trigger element, so that value should be incorporated in the list of classes passed to the button or custom element used. `wrapperClasses` is the same, but for the wrapper element instead (this mirrors the `data-classes` attribute exposed by USWDS).

For example if a user of USWDS did this:
```
    <button type="button" class="usa-button usa-tooltip myCustomButtonClass" data-position="top" data-classes="width-full tablet:width-auto" title="Top">Show on top</button>
```

they should be able to achieve the same result using our Tooltip component with this:
```
    <Tooltip className="myCustomButtonClass" position="top" wrapperClasses="width-full tablet:width-auto" label="Top">Show on top</button>
```

and that should result in the `tooltipWrapper` element having the wrapperClasses value (`width-full tablet:width-auto`), and the `triggerElement` element having the className value (`myCustomButtonClass`) in addition to the other usa tooltip classes

noelledusahel: awesome, I think my latest change should fix it @suzubara

suzubara (approved):
awesome, I think this is a great first pass at the Tooltip! I will create a followup issue for refactoring some of the code to use more React and use less of the JS DOM API

noelledusahel:
> awesome, I think this is a great first pass at the Tooltip! I will create a followup issue for refactoring some of the code to use more React and use less of the JS DOM API

gotcha, when you do please add me as a reviewer!

haworku (dismissed):
⭐  @noelledusahel

src/components/Tooltip/Tooltop.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+  title: 'Components/Tooltip',
+  component: Tooltip,
+  parameters: {
+    info: `
+    USWDS 2.0 Tooltip component
+
+    Source: https://designsystem.digital.gov/components/tooltip/
+    `,
+  },
+}
```
brandonlenz, line 14: Can this be reformatted to match the updated storybook addon-docs format? https://github.com/trussworks/react-uswds/pull/1008

```suggestion
  parameters: {
    docs: {
      description: {
        component: `
### USWDS 2.0 Tooltip component

Source: https://designsystem.digital.gov/components/tooltip/
`,
      },
    },
  },
}
```

brandonlenz (changes requested):
Small change. Otherwise good to approve! Great work!

suzubara (approved):

brandonlenz (approved):

PR #900, opened 2021-02-16
feat: File Input component

### LINKED ISSUE #341: New Component: File Input
# [File Input](https://designsystem.digital.gov/components/form-controls/#file-input)

File input allows users to attach one or multiple files.

## Description
File input built for _synchronous_ upload (all files are submitted once on form submission). Intentionally allows multiple file formats. Has a drag and drop area.

## Requirements (proposed)
**Prop Interface**
``` 
FileInputProps = {
  id: string,
  name: string,
  multiple : boolean, // defaults false
  inputRef: InputRefType
  isValid?: ([files]) = boolean
```
**More Details**
- uses existing `<Label/>` component 
- uses `<input type="file">` attributes- for example, user can limit file types by using `accept`, can accept multiple with `multiple`.
- loads files with generic preview image by default
- expected interaction: When a user selects or drags documents to the file input, the file name and a thumbnail preview are listed. These refresh when new options are selected.  See [uswds js implementation](https://github.com/uswds/uswds/blob/97530c4474ba85bb440399cdd708cd2631070de9/src/js/components/file-input.js). 
- expected safety measure: strip invalid characters from file name.
- The file input component should be a progressive enhancement of `<input type="file" />`. If the component doesn't initialize, it should still work and appear like a standard file input.
- Internet Explorer/Edge - These browsers do not support dragging items to a file input. Instructions to drag files should be removed for these browsers

suzubara:
# Summary

Implements the [USWDS File Input component](https://designsystem.digital.gov/components/form-controls/#file-input).

There are some tests that aren't 100% accurate, because simulating a drop event doesn't seem to accurately trigger the onChange handler for a file input. I left outstanding comments in those cases and maybe they can be improved upon in the future.

## Related Issues or PRs

Closes #341 

## How To Test

**Note:** The FileInput requires updating to USWDS 2.8.1, which is included in this PR. Please make sure you run `yarn install` before testing after you check out this branch.

Test that you can use the input in a multipart form to upload file(s), and the file type validation works as expected.

If you want to handle uploading async on the change event, there is an `onChange` prop exposed you can use to pass in a custom handler. Note that the FileInput component will still display dropped files, and you may want to force a re-render to clear the value after uploading.

### Screenshots (optional)

image

image

image

image


src/components/forms/FileInput/FileInput.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+  const dragText = multiple ? 'Drag files here or ' : 'Drag file here or '
+
+  const filePreviews = []
+  if (files) {
+    for (let i = 0; i < files?.length; i++) {
+      const imageId = makeSafeForID(files[i].name)
+      const key = `filePreview_${imageId}`
+      filePreviews.push(
+        <FilePreview key={key} imageId={imageId} file={files[i]} />
```
haworku, line 56: I wonder if we should surface an optional prop for FilePreview  (defaulting to this component) so people could build their own?

src/components/forms/FileInput/FilePreview.tsx
```diff
@@ -0,0 +1,69 @@
...
+      aria-hidden="true">
+      <img
+        id={imageId}
+        data-testid="file-input-preview-image"
+        src={previewSrc}
+        alt=""
+        className={imageClasses}
+        onError={onImageError}
+      />
+      {name}
```
haworku, line 66: Screen Shot 2021-02-18 at 10 57 24 AM
Got into this weird state when I had a file (screenshot) which had a long name. I wonder if we want to put the name in a block element or something with a set min width.. or should we file as bug to uswds

src/components/forms/FileInput/FilePreview.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+  describe('when the file is done loading', () => {
+    it('renders the file preview image and removes the loading class', async () => {
+      const { getByTestId } = await waitFor(() =>
+        render(<FilePreview {...testProps} />)
+      )
+
+      const expectedSrc = 'data:text/plain;base64,VGVzdCBGaWxlIENvbnRlbnRz'
+
+      const imageEl = getByTestId('file-input-preview-image')
+      await waitFor(() => expect(imageEl).not.toHaveClass('is-loading'))
```
haworku, line 62: nice. I haven't used `waitFor` before TIL.

src/components/forms/FileInput/FileInput.tsx
suzubara: this feels like an enhancement that goes beyond the USWDS implementation, mind adding as a new feature request?

src/components/forms/FileInput/FilePreview.tsx
suzubara: Ah, I believe this is a result of needing to update to USWDS 2.8.1 (this update included some changes to the File Input CSS classes). I get the same result if I was on another branch with 2.8.0, and view the FileInput without running `yarn install`

suzubara:
@haworku I _believe_ for your use case, we'll want to make it so you can programmatically clear/reset the value of the FileInput, which I think requires the same `forwardRef` change mentioned in this thread: https://trussworks.slack.com/archives/C0122MPMXL3/p1613073894012000

This might be a good component to try this out first on, since it will be tested in a real world app right away, and then we'll have a better idea of how to roll out this change to other JS input components as needed. How does that sound?

src/components/forms/FileInput/FileInput.tsx
haworku: Good point https://github.com/trussworks/react-uswds/issues/957

brandonlenz (commented):
Just approved the Happo diffs. There were a lot of very subtle changes to form labels (whitespace and font size), beyond what is mentioned specifically in the [2.8.1 patch notes](https://designsystem.digital.gov/about/releases/#version-281). I suspect these changes were simply not mentioned, but its perfectly reasonable to presume that's where the diffs came from.

haworku:
@suzubara Yep. Sound good. This is a good idea to use to experiment with the pattern here.

src/components/forms/FileInput/FileInput.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+import { FilePreview } from './FilePreview'
+import { makeSafeForID } from './utils'
+
+interface FileInputProps {
+  id: string
+  name: string
+  disabled?: boolean
+  multiple?: boolean
+  accept?: string
```
jim, line 12: It would be nice for the type of `accept?` to be `string | string[]` to make it more obvious how to pass multiple options and avoid needing to split on a string inside the component. I can imagine these use cases:

```jsx
<FileInput accepts="pdf" ... />
<FileInput accepts={["pdf", "docx", "png"]} ... />
```

The second one needs a little more syntax than the current `accept="pdf,docx,png"`, but I think having the type system help a developer provide the options correctly outweighs that.

suzubara: hmm yeah, I think the default string syntax is due to this attribute being passed to the `<input />` (and that's the [syntax for the native `accept` attr](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#attr-accept)) but I could see allowing an array of strings being a nice enhancement. In that case we'd also need to join them into a single string to pass into the `<input />` so there's some extra handling either way.

jim: I didn't realize that `input` had an `accept` attribute natively. In that case it's probably fine as-is, at least in terms of being consistent with how we handle other things in the library. Thanks for the thoughtful response!

src/components/forms/FileInput/FilePreview.tsx
haworku: ha! looks like github deployments is just on time for me over here 😄

haworku (approved):
LGTM. I like having the custom handlers sb example as well.

Issue #923, opened 2021-02-19
[fix] Latest release (1.12.0) cannot be installed


andrewkfiedler:
**ReactUSWDS Version & USWDS Version: 1.12.0**

```
npm ERR! code ELIFECYCLE
npm ERR! syscall spawn
npm ERR! file sh
npm ERR! errno ENOENT
npm ERR! @trussworks/react-uswds@1.12.0 postinstall: `husky install`
npm ERR! spawn ENOENT
npm ERR! 
npm ERR! Failed at the @trussworks/react-uswds@1.12.0 postinstall script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/andrewfiedler/.npm/_logs/2021-02-19T19_20_24_777Z-debug.log
```

**To Reproduce**
1.  Attempt to install 1.12.0 on a project, note the failure above

**Expected behavior**

Installs without errors

**Screenshots**
![Screen Shot 2021-02-19 at 12 30 39 PM](https://user-images.githubusercontent.com/11984853/108552273-50c69680-72ae-11eb-946c-92cd7a85c673.png)

**Additional context**

Rolling back to version 1.11.0 works

brandonlenz:
@andrewkfiedler thanks for reporting this! We'll have a fix available for this in our next release (likely Monday)!

(reopening for now. I will close this manually on release so that you will know when the fix is available).

brandonlenz:
@andrewkfiedler release 1.12.1 is now live! Thanks again, and please let us know if you encounter any other difficulties.

Closing this issue now that the release is live.

PR #925, opened 2021-02-19
fix: disable postinstall script prior to publishing

brandonlenz:
# Note!!:
The alternative to this is to roll back husky to v4: https://github.com/trussworks/react-uswds/pull/926. We should merge one and close the other + delete the branch.

Due to Husky v5 currently causing more headaches than it may be worth, I recommend rolling back to v4 (PR linked above)

# Summary

Husky 5's implementation is a bit different and required adding a `postinstall` script to package.json. This script should only be ran while developing, not for users of the packaged artifact.

Solution is to disable post install scripts before publishing. This is what `pinst` was added for in https://github.com/trussworks/react-uswds/pull/867, however I missed adding the disable. (See [pinst usage ](https://www.npmjs.com/package/pinst#usage) for more info)

closes: https://github.com/trussworks/react-uswds/issues/923

## How To Test

1. Checkout this branch:
```
git fetch
git checkout bl-fix-husky-packaged-post-install-923
```
2. Perform a clean install to verify dev still works (I deleted `node_modules` and ran `yarn` to verify, plus the ability to commit this and see the git hooks still working with husky were good tests)

3. Locally running `yarn prepublishOnly` should run the existing test and linting scripts as well as disabling `postInstall`:
image
image

5. locally running `yarn postpublish` should result in postInstall being enabled once again:
image
image

6. Finally, testing an install of this package after release of this code change will be the true test.


.husky/.gitignore
```diff
@@ -1 +1 @@
-_
+_
```
brandonlenz, line 1: In the original PR I added a newline to this `.gitignore` because it's generally good practice for files to end with newlines. However, every time `husky install` runs (which happens after every `yarn install`, it updates this file. Might as well merge it in this PR so it won't show up as a random diff for anyone.

trussworks-infra:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 09aa81100c29c56e4e12ccb15f69a09185ad06e3
</p>

.husky/.gitignore
brandonlenz: For reference since git doesn't really show this change very well on the `Conversation` tab:
image

brandonlenz:
Closing as we are going with #926 for the near-term at least.

PR #951, opened 2021-02-22
build(circleci): Save and link to CircleCI artifact

ahobson:
# Summary

Deploy storybook as an artifact in CircleCI and link to it from a GitHub deployment

## How To Test

Look at the PR and follow the link to the deployment


ahobson:
The Happo task also builds storybook, but I'm not sure how to save the copy it builds, so we're doing 2x the work for now. If it becomes a bigger deal, we can look into fixing/optimizing things

suzubara:
> The Happo task also builds storybook, but I'm not sure how to save the copy it builds, so we're doing 2x the work for now. If it becomes a bigger deal, we can look into fixing/optimizing things

Just linking this open issue, which is relevant and I remembered seeing before: https://github.com/happo/happo.io/issues/86

brandonlenz (approved):
My limited CI knowledge says this looks good. I'd recommend getting 👀 from someone more well versed in this if time permits, but I won't hold it up.

ahobson:
> My limited CI knowledge says this looks good. I'd recommend getting 👀 from someone more well versed in this if time permits, but I won't hold it up.

I worked with @eeeady on this, so I think we're good to go

Issue #487, opened 2020-09-30
New Component: Step indicator


suzubara:
# [Step indicator](https://designsystem.digital.gov/components/step-indicator/)

A step indicator updates users on their progress through a multi-step process.

![image](https://user-images.githubusercontent.com/2723066/94746951-7e349700-0343-11eb-91ef-aca9580542a2.png)

![image](https://user-images.githubusercontent.com/2723066/94746976-8c82b300-0343-11eb-98f7-c3d42a6e70d1.png)

## Description

A list of steps or screens that indicate a user's progress through a form or process.

> Use the step indicator when the user is working through a form or process that will span several different pages that can be organized into three or more high-level steps or chapters. This helps users understand the progression of long forms and reduces uncertainty as they complete each screen.

> The step indicator is designed to complement standard back/next navigation in a linear sequence, not to be navigation of its own.

## Requirements (proposed)

**Prop Interface**
``` 
StepIndicatorStepProps = {
  status: 'complete' | 'current' | 'incomplete' // default to incomplete
  label: React.ReactNode // required
}

StepIndicatorProps = {
  displayLabels: boolean // default to true
  centered: boolean // default to false
  counterStyle: true | false | 'small' // default to false
}
```

## Example Implementation
```
<StepIndicator centered="true" counterStyle="small" displayLabels="true">
  <StepIndicatorStep status="complete">Personal information</StepIndicatorStep>
  <StepIndicatorStep status="complete">Household status</StepIndicatorStep>
  <StepIndicatorStep status="current">Supporting documents</StepIndicatorStep>
  <StepIndicatorStep>Signature</StepIndicatorStep>
  <StepIndicatorStep>Review and submit</StepIndicatorStep>
</StepIndicator>
```

**More Details**
  -   `<div>`root element with `aria-label="progress"`
  - navigation / links are not part of this component
  - steps are an ordered list `<ol>` inside the root element

## Other Considerations
- custom header render prop?
- language props for "Step x of x", "completed", "not completed"?

hinsxd:
I am interested in making this. Will return to this issue within a day :)

MissyM2:
Hi, there, I would like to work on this, in addition to the alignment of the GovBanner that I noted before.  Let me know if I can do this.

brandonlenz:
Merged to our `uswds-2.9.0` branch

PR #952, opened 2021-02-22
feat: New Component: Step indicator

brandonlenz:
# Summary

Adds new Step Indicator Component.

## Related Issues or PRs

closes: https://github.com/trussworks/react-uswds/issues/487

## Notes

- [ ] Localization/Internationalization: I do not currently have localizations pushed for the screenreader text or the `of` in "3 **of** 5", for example. There are several different ways I can think of to handle this: 
- **Edit:** Looks like uswds is considering options https://github.com/uswds/uswds/issues/3445
- We could pass in a map of localizations as an optional prop to any components that could accept custom verbiage:
```javascript
interface StepIndicatorLocalizations {
  STEP: string
  OF: string
  
}

interface StepIndicatorProps {
  {...}
  localizations?: StepIndicatorLocalizations
}
```
---

```javascript
interface StepIndicatorStepLocalizations {
  COMPLETED: string
  NOT_COMPLETED: string
}

interface StepIndicatorStepProps {
  {...}
  localizations?: StepIndicatorStepLocalizations
}
```

  - We could also work with a more "react-y" solution using useContext for localizations at the App level.
  - This may deserve a larger discussion and github issue (I'll file one if one does not already exist) to decide on the right solution.

## How To Test

Checkout `bl-step-indicator-487` and run storybook to inspect the component variations further

#### Default

image

---
#### No Labels

image

---
#### Centered Labels

image

---
#### Counters

image

---
#### Small Counters

image


trussworks-infra:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 41214f358e8b4a38e20b8ebebdb59198f0214e54
</p>

src/components/stepindicator/StepIndicator/StepIndicator.tsx
```diff
@@ -0,0 +1,70 @@
...
+        {children}
+      </ol>
+      <div className="usa-step-indicator__header">
+        <h2 className="usa-step-indicator__heading">
+          <span className="usa-step-indicator__heading-counter">
+            <span className="usa-sr-only">Step</span>
+            <span className="usa-step-indicator__current-step">
+              {currentStepNumber}
+            </span>
+            &nbsp;
```
brandonlenz, line 59: If you noticed this `&nbsp;` being present here but not in the [source code](https://designsystem.digital.gov/components/step-indicator/), I can add soem context:

In the examples, there are spaces as shown in the image below:
image

These come from [implicit spacing added between `html` `inline` elements](https://stackoverflow.com/questions/5078239/how-do-i-remove-the-space-between-inline-inline-block-elements). 

Since rendered react html is minified, the implicit spacing is not rendered, so I had to explicitly add spaces here.

I may raise an issue with `uswds`. While this is not a quirky bug with a single browser, they may want to know about an inconsistency in display of one of their components when code is minified vs non-minified, similar to https://github.com/uswds/uswds/issues/3997.

haworku:
> This may deserve a larger discussion and github issue (I'll file one if one does not already exist) to decide on the right solution.

Let's definitely make an issue. I like the idea of starting a pattern for `localizations` 🌟 . And also things like _"step X of X"_ are  usually complicated (you don't want to translate a sentence or phrase word by word for i1n because that assumes certain grammar structure).  We might want to throw around more ideas with that or decide its out of scope.

src/components/stepindicator/StepIndicator/StepIndicator.tsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react'
+import classnames from 'classnames'
+import { StepIndicatorStepProps } from '../StepIndicatorStep/StepIndicatorStep'
+
+interface StepIndicatorProps {
+  showLabels?: boolean
+  counters?: true | false | 'small'
```
haworku, line 7: I think this type is pretty confusing because its a boolean or a string. Lets make this more similar to our variant props which use string union type -  `default | small  | none`  or something that?

src/components/stepindicator/StepIndicator/StepIndicator.tsx
```diff
@@ -0,0 +1,70 @@
...
+    const i = children.findIndex((step) => step.props.status === 'current')
+    return i === -1 ? 0 : i
+  }
+  const currentStepIndex = findCurrentStepIndex()
+  const currentStepNumber = currentStepIndex + 1
+  const currentStepLabel = children[parseInt(`${currentStepIndex}`)].props.label
+  const totalNumberOfSteps = children.length
+
+  return (
+    <div className={classes} data-testid="step-indicator" {...divProps}>
```
haworku, line 48: Should we add `aria-label: progress` here since its explicitly called out in the step indicator [guidance?](https://designsystem.digital.gov/components/step-indicator/) it could always be overriden by `divPops`

haworku (commented):
Ready to approve this just would like to get your 👀  on the `counters` type again

src/components/stepindicator/StepIndicator/StepIndicator.tsx
brandonlenz: fixed in ec8cf08

src/components/stepindicator/StepIndicator/StepIndicator.tsx
brandonlenz: Meant to make a separate commit for this, but it was also fixed in ec8cf08

haworku (approved):

PR #983, opened 2021-03-01
feat: Add DateRangePicker Component

### LINKED ISSUE #339: New Component: Date Range Picker
# [Date Picker Range](https://designsystem.digital.gov/components/form-controls/#date-range-picker)

Add a new `DateRangePicker` component following our adding new components guide. The USWDS implementation in linked in the title text of this issue.

## Description
A date range picker helps users select a range between two dates. Use the new `DatePicker` when scheduling or when day of the week is important.  Use existing `DateInput`  when desired date is already known (such as a birthday). 

## Requirements (proposed)
**Prop Interface**
``` 
DatePickerProps = {
  id: string,
  name: string,
  isValid: (input1Value, input2Value) => boolean
  minDate?: string (YYYY-MM-DD)
  maxDate?: string (YYYY-MM-DD)
```
**More Details**
-  This component uses two `DatePicker` components #338.  That work should be completed first.
- `<div class="usa-date-range-picker/>"` is root element
- using dates outside the range results in a validation error
- remaining input props (from `JSX.IntrinsicElements['input']) should be spread on both DatePicker inputs individually.  
- support a variety of validation formats with `isValid` function prop.  
- `minDate` and `maxDate` is used to override `data-min-date` and `data-max-date` defaults on the two picker elements.  The default min date is 0000-01-01. There is no default max.
- see [uswds date range picker](https://github.com/uswds/uswds/blob/develop/src/js/components/date-range-picker.js) js for several helpful date utilities

brandonlenz:
# Summary

Adds [Date Range Picker](https://designsystem.digital.gov/components/form-controls/#date-range-picker) component

## Related Issues or PRs

closes #339 

## How To Test

Thanks to the CI artifacts, you should be able to simply click "show environments" to access storybook and mess with this new component

### Screenshots (optional)

image


trussworks-infra:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against b749362edf1611a4dea46c92da76ec23d3dfd3ed
</p>

src/components/forms/DatePicker/DatePicker.tsx
```diff
@@ -24,20 +24,21 @@ import {
...
-interface DatePickerProps {
+export interface DatePickerProps {
   id: string
   name: string
   className?: string
   disabled?: boolean
   required?: boolean
   defaultValue?: string
-  minDate: string
+  minDate?: string
```
brandonlenz, line 34: Since this prop has a default value, I made it optional so that it isn't required for the props of either datepicker in the DateRangePicker. I think it was intended to be optional in the initial Datepicker release, given the default prop specified for storybook, and the default value when extracting props.

src/components/forms/DateRangePicker/DateRangePicker.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+interface DateRangePickerProps {
+  startDateLabel?: string
+  startDateHint?: string
+  startDatePickerProps: Omit<DatePickerProps, 'rangeDate'>
+  endDateLabel?: string
+  endDateHint?: string
+  endDatePickerProps: Omit<DatePickerProps, 'rangeDate'>
+  className?: string
+}
```
brandonlenz, line 16: I deviated slightly from what USWDS recommends here. It would be no big deal to change though, so I'm curious on reviewers' thoughts. USWDS recommends a min date and max date for the entire date range picker. At present, instead, this implementation allows a range to be specified on each date picker respectively. This would allow for the use case where the start date must fall within a given range, and the end date must fall within a separate range.

Once a start or end date is selected, however, the range of the other date picker is bounded by start/end date OR the specified date range, whichever is more restrictive.

src/components/forms/DateRangePicker/DateRangePicker.tsx
```diff
@@ -0,0 +1,145 @@
...
+        return formatDate(parsedCurrentStartDate)
+      } else {
+        return formatDate(parsedMinEndDate)
+      }
+    } else {
+      return (parsedCurrentStartDate && formatDate(parsedCurrentStartDate)) 
+        || (parsedMinEndDate && formatDate(parsedMinEndDate)) 
+        || undefined
+    }
+  }
```
brandonlenz, line 94: I've just identified a bug with the logic in these methods. Fixing and adding tests.

brandonlenz: 6180ab6 & 2fedaec fix the issues I had identified.

brandonlenz: ad82800 adds the test covering these issues so they do not get reintroduced

src/components/forms/DatePicker/DatePicker.tsx
```diff
@@ -24,20 +24,21 @@ import {
...
   maxDate?: string
   rangeDate?: string
   onChange?: (val?: string) => void
   onBlur?: (
     event: React.FocusEvent<HTMLInputElement> | React.FocusEvent<HTMLDivElement>
   ) => void
+  inputProps?: JSX.IntrinsicElements['input']
```
brandonlenz, line 41: I meant to add a note about this change. I assume this is a breaking change, so I'm looking at alternatives as well.

Since the `DateRangePicker` uses `DatePickerProps` to pass through props to each `DatePicker`, I wanted `DateRangeProps` to include all the props in `DatePicker`'s signature. 

One problem I noticed while keeping the intersection type is that the custom `onChange` prop of `DatePicker` conflicts with `onChange` from JSX.IntrinsicElements['input']:

```
Type '(val?: string | undefined) => void' is not assignable to type '((val?: string | undefined) => void) & ((event: ChangeEvent<HTMLInputElement>) => void)'
```

brandonlenz: For example, the following fixes both issues (prevents breaking changes, and avoids the "onChange" overlap), while still allowing DateRangePicker to only need to know about the exported `DatePickerProps` for passing through to each `DatePicker`:

```
interface BaseDatePickerProps {
  id: string
  name: string
  className?: string
  disabled?: boolean
  required?: boolean
  defaultValue?: string
  minDate?: string
  maxDate?: string
  rangeDate?: string
  onChange?: (val?: string) => void
  onBlur?: (
    event: React.FocusEvent<HTMLInputElement> | React.FocusEvent<HTMLDivElement>
  ) => void
}

export type DatePickerProps = BaseDatePickerProps & Omit<JSX.IntrinsicElements['input'], "onChange">

export const DatePicker = (
  props: DatePickerProps
): React.ReactElement => {
  const {
    id,
    name,
    defaultValue,
    disabled,
    required,
    minDate = DEFAULT_MIN_DATE,
    maxDate,
    rangeDate,
    onChange,
    onBlur,
    ...inputProps
  } = props
```

haworku: Having`BaseDatePickerProps` and then `DatePickerProps` as the intersection with `input` types works. I think we did something like that with `<CharacterCount /> . I'm fine with using `Omit` for now as way to get around the issue with types... but it points to something bigger.

We probably need to be strict with what is an `onChange`. This is is an issue in Combobox as well (which I helped work on haha)  - we have `onChange` handling the value of the input but not the actual ChangeEvent like the native HTML. I don't know what to do about this... I wonder if for v2 we should have clearer naming of our props so that `onChange` is only ever handling `ChangeEvent` and anything else is called `onValueChange` or something like that...

brandonlenz: Updated in de11085

I like the idea of having separate verbiage for onChange vs similar custom implementations!

src/components/forms/DateRangePicker/DateRangePicker.tsx
suzubara: I think I see what you're saying but I need a little more clarification to understand fully.. are you saying the USWDS version would allow for incompatible start/end dates, and the way you did it will guard against that? if so I'm all in favor

src/components/forms/DateRangePicker/DateRangePicker.tsx
```diff
@@ -0,0 +1,159 @@
...
+  const endDatePickerOnChange = getDatePickerOnChangeFn(
+    endDatePickerProps.onChange,
+    setEndDateInternalValue
+  )
+  
+  const classes = classnames(className, 'usa-date-range-picker')
+
+  return (
+    <div className={classes} data-testid="date-range-picker">
+      <div className="usa-form-group">
```
suzubara, line 114: Could we use our own `FormGroup` & `Label` components here? I'm thinking of how we might want to expand this component in order to display an error state for example, and since those components already take an `error` prop that might make it easier

suzubara (dismissed):
When comparing the rendered markup with the USWDS source, I noticed the USWDS passes these classes along into the `DatePicker` components:
```
usa-date-range-picker__range-start
usa-date-range-picker__range-end
```

that and the other comments I left are pretty minor. I tested this out on MilMove and it looks great!! nice work 🎉

src/components/forms/DateRangePicker/DateRangePicker.tsx
brandonlenz: It wouldn't cause problems that we could adjust the logic for internally so much.

If we want this to match USWDS perfectly, we'd just expose a min and max date prop for the DateRangePicker, omit the min and max date pass through props for each DatePicker, and feed them in manually. 

The only reason I didn't do that is that the current solution allows this behavior (pass the min and max dates to each date picker), but also allows an implementer to specify different min and max dates for each date picker if they wanted to have separate min/max dates for each date picker. 

It just seemed like a feature that might be useful. I wanted to call it out and defer to reviewers opinions on that though. I can pretty easily refactor it to constrain the min and max date to apply to the entire date range picker rather than each date.

src/components/forms/DateRangePicker/DateRangePicker.tsx
brandonlenz: Passed those classes through in bd99ba0, as well as updating to using our own FormGroup and Label components

brandonlenz:
@suzubara your review prompted me to take a closer look at the generated markup. Looks like I missed the `aria-describedby` property on the internal and external inputs as well. Fixed in https://github.com/trussworks/react-uswds/pull/983/files/a77722bfeca340bf37934591ee8196dad3ab8a4b..cb45ebe7071e15a135ddcb1a47bcfbf372440d52

src/components/forms/DateRangePicker/DateRangePicker.tsx
suzubara: Ahh I see, that makes sense! I like how you expanded it so I vote to keep it!

suzubara (approved):
🚀

brandonlenz:
@suzubara same story here. Had to merge main and resolve merge conflicts from main, so this will likely need another approval. Thanks for the reviews!

suzubara (approved):

PR #902, opened 2021-02-18
fix: add ActionType.CLEAR_FOCUS to indicate a user has left the field

christopherhuii:
# Summary
The ComboBox has issues with "losing" its focus when clicking outside of the component.

https://user-images.githubusercontent.com/8367504/108426115-35955180-71f0-11eb-91c1-9ea40d8bd920.mov

A related issue is that the `FocusMode` is still set to `FocusMode.Input` even after the user has left the component.

This PR allows the user to:
1. Unfocus from the ComboBox when they click outside
2. Resets the `FocusMode` to `None` when they click outside

I achieved this by creating a new `ActionType` called `CLEAR_FOCUS` to indicate when a user has clicked outside of the combo box. I am open to suggestions on the name. This action is different from `ActionType.CLEAR` because `CLEAR` empties the input, whereas `CLEAR_FOCUS` indicates the user has left the field.

## Related Issues or PRs
Fixes #901 

## How To Test
1. Open Storybook
2. Focus ComboBox by opening its options or typing in the input
3. Click outside of the component and the input should lose focus.

### Screenshots (optional)

After Fix:

https://user-images.githubusercontent.com/8367504/108428519-89556a00-71f3-11eb-8029-d19cf7703133.mov

Issue #901, opened 2021-02-16
[fix] ComboBox forces focus when another field is changed


christopherhuii:
**ReactUSWDS Version & USWDS Version:**

ReactUSWDS `v1.10.0`
USWDS: `v2.8.1`

**Describe the bug**

My form has multiple fields. If I select an option from the `ComboBox`, then type in the input field, the `ComboBox` steals focus and I'm unable to finish filling out my text input.

https://user-images.githubusercontent.com/8367504/108134906-e9210900-706b-11eb-88d1-70b28b51d823.mov

**To Reproduce**
I was able to reproduce this error in Storybook.
```
export const comboBoxBug = (): React.ReactElement => {
  const [ val, setVal ] = useState('');
  const fruitList = Object.entries(fruits).map(([key, value]) => ({
    value: key,
    label: value,
  }))

  return (
    <Form onSubmit={noop}>
      <Label htmlFor="input-ComboBox">Combo Box</Label>
      <ComboBox
        id="input-ComboBox"
        name="input-ComboBox"
        options={fruitList}
        onChange={noop}
      />
      <Label htmlFor="test">Input Field</Label>
      <TextInput
        id="test"
        name="test"
        type="text"
        onChange={(e) => {
          setVal(e.target.value)
        }}
        value={val}
      />
    </Form>
  )
}
```
1. Select an option in the `ComboBox`
2. Type into the `TextInput`
BUG: Notice the focus switch to the `ComboBox` and disrupts the user's ability to type in the `TextInput`

**Expected behavior**
I expect the focus to remain unchanged.

PR #989, opened 2021-03-05
fix: set CombBox focusMode to BLUR when focus is outside of the component

christopherhuii:
# Summary
In #901, we ensured that the combo box's `focusMode` was set to `BLUR` when a user **clicked** outside of the component. This is not sufficient for keyboard shortcuts, such as tabbing. When a user tabs through input fields, the focus on the combo box isn't reset.

*I also added an additional story that puts a text input next to the combo box to make it easier for folks to test. Since we've had some issues surrounding focus, I think it'd be good to have for testing.

## Related Issues or PRs 
#901 
#902

## How To Test
The form must have at least 2 elements to interact with. I used the combo box and a text input.
1. Fill in the combo box (e.g. type value, select from dropdown, etc.)
2. Use `tab` to focus the next field
3. Type in the text field

Prior to the fix, the user's focus would be shifted back to the combo box.
After the fix, the user is able to continue typing in the text input

### DIFF
--- a/src/components/forms/ComboBox/ComboBox.stories.tsx
+++ b/src/components/forms/ComboBox/ComboBox.stories.tsx
@@ -3,7 +3,7 @@ import React from 'react'
 import { ComboBox } from './ComboBox'
 import { Form } from '../Form/Form'
 import { Label } from '../Label/Label'
-
+import { TextInput } from '../TextInput/TextInput'
 import { fruits } from './fruits'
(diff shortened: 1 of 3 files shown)


haworku (approved):
Thanks for the new storybook example - good pattern that we can use in other components too to improve testing. Glad we are making incremental a11y improvements.

PR #1010, opened 2021-03-10
fix: ComboBox "keyboard trap" and misc bugs

christopherhuii:
# Summary

[ReactUSWDS Storybook](https://trussworks.github.io/react-uswds)
[USWDS](https://designsystem.digital.gov/form-controls/03-combo-box/)

This PR fixes a "keyboard trap" issue as well as other miscellaneous bugs.

1. When the user has no matching options, the user can't tab out.
2. When the user types input, no option is focused, and hits "tab", the input should clear and switch focus
3. When the user types input, and the text perfectly matches an option (case insensitive), the input should select/fill combo box with the correct option.


brandonlenz (approved):
Nice. The behavior of the ComboBox seems to be in line with how the USWDS one behaves now. Thanks for taking this on! ✅

PR #1026, opened 2021-03-16
docs(adr): Release and Branching

brandonlenz:
# Summary

ADR for defining/documenting branching strategy following the [engineering playbook](https://github.com/trussworks/Engineering-Playbook/blob/138f807140b6999c0a12d870865cc9ee44eb417c/documentation/adr.md). 

For the purposes of the ADR process and discussion, I included a `decision outcome` here. Please use this PR as discussion and know that the true decision outcome should come from a consensus among maintainers. If there are pros/cons or options/suboptions I have no evaluated, please feel free to add them.


docs/adr/0001-branching-strategy.md
```diff
@@ -0,0 +1,92 @@
+# Document and support branching strategy for allowing parallel development when development dependencies exist
+
+* Status: proposed
+* Deciders: [Codeowners](../../CODEOWNERS)
+* Date: 2020-03-16
+
+## Context and Problem Statement
+
+Due to a combination of temporarily having a large influx of people giving some attention to `react-uswds`, and our intention to be careful about releasing breaking changes, we recently created develop branches corresponding to [Github milestones](https://github.com/trussworks/react-uswds/milestones) parallel to `main` to allow simultaneous streams of work that would have previously been blocked (branches `V2` and `usdws-2.9.0`).
```
ahobson, line 9: I'm still not sure why work was blocked vs creating a release with all of the current work and then starting new work with breaking changes on main. I don't feel like you've discussed that here.

As we talked about when we chatted, the developer workflow and the release workflow are necessarily tied together. I think the top level goals we have are around release workflow and developer workflow is an important consideration as to which release workflow we should choose. There might be multiple developer workflows per release workflow, so I think you might consider listing the top 1 or two per release workflow.

Make sense?

ahobson (changes requested):

docs/adr/0001-branching-strategy.md
brandonlenz: > I'm still not sure why work was blocked vs creating a release with all of the current work and then starting new work with breaking changes on main. I don't feel like you've discussed that here.

The general consensus that I've understood onboarding and interacting with the project is to be very careful about making breaking changes, and ideally bundle multiple into a single release in an effort to be kinder to consumers. If there are disagreements with that philosophy, I would think we should have an ADR to get consensus there. I'd be interested to see @haworku and @suzubara weigh in on that since they have a longer history with the project than I do.

We talked about releases and what constitutes breaking vs non-breaking this week in the "irl" meeting, which we will create an ADR for. Maybe it makes sense to evaluate that before we make a decision on branching as well?

As far as I've understood, there has been no need or desire to change the release process. This ADR was written under that assumption. If true, then our branching options must accommodate the current process, though it is fair to consider the merit of flexibility of a branching strategy for alternative release strategies. That's covered in this ADR.

Ultimately what I'm gathering is we have 3 related, but separate, decisions to make via ADRs: 

1. > Why don't we just release breaking changes as frequently as we want? (ADR based of the concerns noted above)

2. > What constitutes a breaking change? (ADR from IRL meeting)

3. > Should we have a documented branching strategy for navigating development dependencies (This ADR)

And of course, if our answer to the first ADR is "Just release breaking changes immediately, always" then this ADR will no longer be needed.

I think it makes sense to evaluate them in that order, since each builds off of the preceding ADRs.

brandonlenz: It probably actually make sense to define a breaking change first, so the order I would recommend would be (with better titles):

1. What constitutes a breaking change?
2. How should breaking changes be released?
3. What should our documented branching strategy for navigating development dependencies look like?

haworku: > I'm still not sure why work was blocked vs creating a release with all of the current work and then starting new work with breaking changes on main. I don't feel like you've discussed that here

@ahobson the answer to your question in detail-  the scenario is we had someone submit a PR for an existing issue  that required a major bump in the USWDS version in our `package.json`.  We didn't want to just merge that to main because we hadn't supported other USWDS changes from in between versions that including breaking changes from USWDS. That work was blocked for this reason and instead the change was separate branch for that uswds version.

The other scenario is we started a separate v2 branch when we had random PRs with breaking changes to our library, react-uswds (unrelated to uswds version).  That work would come in alongside many other types of changes and PRs.  As mentioned - _maybe we want to start releasing those breaking changes often and bumping major version randomly_  Look forward to a discussing that as needed. I def thought we didn't want to, but I bring that perspective to  from previous experience where we planned our major releases for our UI library in advance, always grouping a few breaking changes together. I thought  sure we might bump with uswds version but not at random for a single change. That perspective might be overkill for this project I realize now.

docs/adr/0002-release-and-branching-strategy.md
```diff
@@ -0,0 +1,208 @@
...
+  * Current `react-uswds` branching strategy.
+  * `main` is the primary develop branch.
+  * Features are PRed to `main`.
+  * Bugfixes are PRed to `main`.
+  * Releases take place via short-lived release branches based off `main`, tagged on `main`.
+* Pros and Cons
+  * Good, because it is suitable for small project with limited time available.
+  * Good, because, as the standard/out-of-the-box branching strategies for new git projects, it should be familiar to developers of all skill levels.
+  * Bad, because it does not scale well when development dependencies exist and a large number of people are actively contributing.
+  * Bad, because it does not provide clear direction for navigating development dependencies.
```
suzubara, line 157:
```suggestion
  * Bad, because it does not provide clear direction for navigating changes to development dependencies.
```

brandonlenz: b1b468c

docs/adr/0002-release-and-branching-strategy.md
```diff
@@ -0,0 +1,208 @@
...
+  * Additional develop branches are the primary extension of the default branching strategy
+  * Releases take place via short-lived release branches based off `main`, tagged on `main`.
+    * As noted below, this strategy also supports custom releases from the additional develop branches, if desired.
+* Pros and Cons
+  * Good, because it scales extremely well to *both* custom developer workflows and custom release strategies
+    * During times of low developer availability or periods without development dependencies the branching strategy can be scaled to be effectively identical to the [Default](#default) strategy.
+    * Its flexibility supports multiple release strategies, ideal for future changes to that process.
+  * Good, because it provides a clear process for navigating development dependencies and unblocking corresponding work.
+  * Good, because it retains the familiar process of PRing off of and to a base develop branch.
+  * Bad because, while based on well documented and widely understood concepts, this custom branching strategy would require additional documentation to make clear for new contributors.
```
suzubara, line 200: Let's mitigate this by adding documentation for:
- How to start a new USWDS version branch (this should probably start with the version upgrade issue)
- Working on USWDS features that are ahead of the current supported version
- Knowing when & how to release any breaking changes (this is primarily a judgement call depending on the severity of the change)

...and also by making sure we're using the USWDS version milestones and `breaking` labels consistently. Maybe we add a label to indicate issues related to the current version as an easy way for contributors to find things to work on?

docs/adr/0002-release-and-branching-strategy.md
```diff
@@ -0,0 +1,208 @@
...
+  * Bad because, it increases developer overhead by adding more considerations to the development process (i.e. "What branch do I use as a base for the work I want to do?").
+    * Bad, because it functions best with supporting automation which would likely be a custom development in this case.
+  * Bad because, and this is a *big* one: existing CI and merge processes would need to be re-evaluated.
```
suzubara, line 203: I think we should have a defined branch prefix (maybe `dev-`) for these branches so we can easily extend CI processes to include them

suzubara (approved):
overall I'm in agreement with the chosen option, mostly because it's close to what we're currently doing and I want to avoid spending _too_ much time with this decision until there's a clear problem to solve. You've obviously put a lot of thought into these options and I think this ADR does a great job of laying the groundwork upon which the project can evolve based on future needs. thank you!!

ahobson (approved):

docs/adr/0002-release-and-branching-strategy.md
```diff
@@ -0,0 +1,208 @@
...
+### Option 2: Release all changes immediately
+* Good, because it lifts restrictions on the developer workflow that breaking or blocking changes would ordinarily cause.
+* Good, because developer overhead is significantly reduced by removing decision energy around planning work.
+* Good, because maintainer overhead is reduced by removing decision energy around planning releases.
+* Good, because it facilitates a consistent release process. There are no conditional steps in the process, just release and increment the appropriate version accordingly.
+* Good, because it eliminates the need for prioritizing the backlog.
+* Bad, because the stability of the library for consumers becomes a secondary concern.
+* Bad, because more frequent breaking releases.
+* Bad, because being a consumer of `react-uswds` may feel chaotic to endure breaking releases between features.
+* Bad, because the level of support in relation to USWDS becomes unclear as features may be released out of "chronological" order (i.e. `react-uswds` depends on 2.10.0 for a released 2.10.0 component before all 2.9.0 components are completed).
```
ahobson, line 109: Do we have any more docs on this? It seems unfortunate to me that if one of our projects needs only one component from a future release we can't release a version of react-uswds with it until we've implemented all components from all prior releases. Or do I misunderstand?

ahobson: Said another way, IMHO our compatibility promise should be that *for all components we have implemented* they will be compatible with a given USWDS release, not that we have implemented all components from a particular USWDS release.

brandonlenz: I don't think we formally have more docs on this. Most of these are coming from discussions I've been a part of. That's an interesting way to look at it though! I don't have reservations with the idea of "sniping" USWDS components by updating our version to do so, so long as we're still intentional about picking up with smaller changes and fixes that often accompany USWDS releases. We've been more diligent about that lately, so that practice would definitely help.

In that scenario I can also see USWDS changes for a particular release being specific to an un-implemented React USWDS component, which we would also just have to be diligent about making issues in the backlog for, or updating the "new component" issue. That way we don't miss out on things to maintain parity while "skipping" components.

Framing the "compatibility promise" a certain way and documenting that publicly would be really helpful here. Now I'm not exactly sure if we've implicitly adopted one, and whether its even the right one to use.

ahobson: I think that compatibility promise has a big impact on the branching strategy and why it is or is not necessary. I'm not trying to derail this ADR, but I am trying to make sure we document the assumptions that go into the ADR. I wonder if it's possible compatibility promise is another pre-requisite ADR?

brandonlenz: No worries! Fully understood and you're absolutely right! I _do_ think the compatibility promise makes or breaks some of these options, and it shouldn't be left as an assumption as I had done. I'm leaning in the direction of a pre-requisite ADR. On its own, I don't think it would be very large.

haworku:
@brandonlenz @ahobson @suzubara I see a lot of good discussion here but it seems like we left things in a place where we aren't fully ready to adopt this ADR as is written. Going to close for now, we can always revisit.

Issue #737, opened 2020-12-22
New Component: Identifier


haworku:
# [Identifier](https://designsystem.digital.gov/components/identifier/)

## Description
The identifier communicates a site’s parent agency and displays agency links required by federal laws and policies.
![Screen Shot 2020-12-28 at 2 14 16 PM](https://user-images.githubusercontent.com/10750442/103240814-fa288600-4916-11eb-80e8-3bb46a15d86e.png)

## Requirements (proposed)
**Prop Interface**
``` 
Props = { }  & JSX.IntrinsicElements['div']
```

**Sample Implementation**
```javascript
 return (
      <div className="usa-identifier">
       {children}
      </div>) 
```

```JSX
<Identifier>
  <IdentifierMasthead />
  <IdentifierLinks / >
  <IdentifierGov>
 </Identifier>
```
**More Details**
  -   `<div>`root element
  -  IdentifierMasthead and IdentifierGov components have `section` root element, IdentifierLinks is `nav`

## Other Considerations
-  "With rare exceptions, if you use the identifier, include the entire identifier. That is, don't delete sections or required links or change any link text beyond the customizations mentioned in the implementation section."
- "Use the identifier component for required links. If your site already includes the federally required links in its site footer, remove them in favor of the links in the identifier."

suzubara:
Fixed by https://github.com/trussworks/react-uswds/pull/1044

PR #1044, opened 2021-03-22
feat: Identifier component MVP

SirenaBorracha:
# Summary

This PR adds the [Identifier component](https://designsystem.digital.gov/components/identifier/) from USWDS. 

I'm not sure about my Storybook examples, some may be unnecessary (others might be missing), and I believe the test suite could be more complete. 

We decided against refactoring `IdentifierLinks` / `IdentifierLinkItem` / `IdentifierLink` as they follow currently practice using custom `Link` - if a simpler way is obvious, please call it out.

## Related Issues or PRs
#737 

## How To Test

`yarn storybook`
`yarn test`

Or, click "show environments" to view this component in Storybook

#### Default, English and Spanish

Screen Shot 2021-03-29 at 6 28 20 PM

#### No logos, English and Spanish
Screen Shot 2021-03-29 at 6 29 22 PM

#### Multiple parent agencies and logos, English and Spanish
Screen Shot 2021-03-29 at 6 28 34 PM

#### With tax disclaimer, English and Spanish
Screen Shot 2021-03-29 at 6 29 44 PM

#### More than two parents and logos // Tax disclaimer and multiple parents and logos
Screen Shot 2021-03-29 at 6 28 52 PM


trussworks-infra:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 59c38455c7bb32c59ec2a81a8edc3bb9981f1bc1
</p>

src/components/Identifier/Identifier/Identifier.stories.tsx
```diff
@@ -0,0 +1,248 @@
...
+
+export default {
+  title: 'Components/Identifier',
+  component: Identifier,
+  parameters: {
+    info: `
+  USWDS 2.0 Identifier component
+  
+  Source: https://designsystem.digital.gov/components/identifier/
+  `,
```
brandonlenz, line 20: We recently changed this convention for Storybook story descriptions. Could you update this one to match? See https://github.com/trussworks/react-uswds/commit/ed171f924c7455351f2d07a7422317191bd1f4bb for details.

src/index.ts
```diff
@@ -7,6 +7,7 @@ export { Accordion } from './components/Accordion/Accordion'
 export { GovBanner } from './components/GovBanner/GovBanner'
 export { Button } from './components/Button/Button'
 export { ButtonGroup } from './components/ButtonGroup/ButtonGroup'
+// export { Identifier } from './components/Identifier/Identifier'
```
brandonlenz, line 10: This looks like an accident that we can remove

src/components/Identifier/IdentifierMasthead/IdentifierMasthead.tsx
```diff
@@ -0,0 +1,90 @@
...
+            <p className="usa-identifier__identity-domain">{domain}</p>
+            <p className="usa-identifier__identity-disclaimer">
+              {copy.identityDisclaimer}
+              &nbsp;
+              <a href={parentAgency.url}>{parentAgency.name}</a>
+              {taxpayerDisclaimer && `. ${taxpayerDisclaimer}`}
+            </p>
+          </div>
+        </div>
+      </div>
```
brandonlenz, line 85: I noticed that there isn't as much whitespace between the logo and the agency & domain text in the storybook screenshots as there is in the USWDS examples. I _think_ that's just because one of the closing `</div>` tags needs to be before the `identifierMasthead-agency-description` div. It's that sneaky one here in their source code:

image

brandonlenz (commented):
A few small things I noticed on a first pass. This is looking great! 🎉

src/index.ts
SirenaBorracha: Should I remove it or comment it in?

src/components/Identifier/IdentifierMasthead/IdentifierMasthead.tsx
SirenaBorracha: Confirming that that fixed it, thank you for spotting that!

Screen Shot 2021-03-26 at 11 10 05 AM

src/index.ts
brandonlenz: Turns out my comment was the 'accident'. I misread which file this was in and just saw a commented line, thinking it should be removed 🤦 

Identifier and any publicly exposed subcomponents _should_ be exported here, so comment it in. I _would_ go ahead give it its own section under a comment though `/** Identifier Components */`

src/components/Identifier/IdentifierMasthead/IdentifierMasthead.tsx
brandonlenz: Nice! Looks great.

src/components/Identifier/Identifier/Identifier.stories.tsx
```diff
@@ -0,0 +1,248 @@
+import React from 'react'
+/*  eslint-disable jsx-a11y/anchor-is-valid, react/jsx-key */
```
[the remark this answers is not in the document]
SirenaBorracha: Removed

src/components/Identifier/Identifier/Identifier.stories.tsx
SirenaBorracha: Updated

src/components/Identifier/IdentifierMasthead/IdentifierMasthead.tsx
```diff
@@ -0,0 +1,90 @@
...
+      <div className="usa-identifier__container">
+        <div className="usa-identifier__logos">
+          {parentAgency.logo && (
+            <IdentifierLogo
+              agencyUrl={parentAgency.url}
+              agencyName={parentAgency.name}
+              src={parentAgency.logo}
+            />
+          )}
+        </div>
```
brandonlenz, line 72: I know you potentially have a refactor with how this is handled, so I won't expect a change in response to this feedback directly, just throwing it out as a heads up. Feel free to fix alongside your other changes to masthead if that makes more sense to you.

Basically, if there are no logos, the `usa-identifier__logos` `div` should not be present at all. For example, with how the code is written now, that would involve the conditional render of the entire div based on `parentAgency.logo` instead of just the contents of the `div`. 

That appears to be responsible for the extra space when no logos are present:
image

suzubara (commented):
Just adding this for the record: @SirenaBorracha and I discussed how to handle the content and localization for this component and thought it would be simpler to allow users to pass in `children` or another prop to `<IdentifierMasthead />` so avoid over-architecting this (and allow for flexibility for other languages). I still think the EN/SP versions should be included in Storybook to demonstrate how to use this component to mirror the USWDS examples, but the copy itself can be passed in there instead of hardcoding content in the components themselves.

src/components/Identifier/Identifier/Identifier.stories.tsx
```diff
@@ -0,0 +1,426 @@
...
+  </div>,
+]
+
+export const identifierDefault = (): React.ReactElement => (
+  <Identifier>
+    <IdentifierMasthead aria-label="Agency identifier">
+      <div className="usa-identifier__container">
+        <div className="usa-identifier__logos">
+          <IdentifierLogo
+            agencyUrl="#"
```
brandonlenz, line 132: Now that the logo is not necessarily connected to a specific agency, we can simplify `agencyUrl` to `href` to make it more like other "Linking" components, which should also make it more obvious what the component does besides displaying an image.

src/components/Identifier/Identifier/Identifier.stories.tsx
```diff
@@ -0,0 +1,426 @@
...
+            className="usa-identifier__logo-custom-class-name">
```
brandonlenz, line 133: Is this `custom-class-name` doing anything for us in the Storybook stories? If not, I'd remove it.

src/components/Identifier/Identifier/Identifier.stories.tsx
```diff
@@ -0,0 +1,426 @@
...
+          </IdentifierLogo>
+        </div>
+        <div
+          data-testid="identifierMasthead-agency-description"
+          className="usa-identifier__identity"
+          aria-label="Agency description">
+          <p className="usa-identifier__identity-domain">domain.edu.mil.gov</p>
+          <p className="usa-identifier__identity-disclaimer">
+            {`An official website of the `}
+            <a href="testlink">Test Agency Name</a>
```
brandonlenz, line 144: In the storybook stories, I think the `href`s should all basically be to "#" (unless they actually point to somewhere real).

That will require you to disable the eslint rule in this file again, since that's an actual href on an anchor tag.

src/components/Identifier/Identifier/Identifier.stories.tsx
```diff
@@ -0,0 +1,426 @@
...
+          </p>
+        </div>
+      </div>
```
brandonlenz, line 147: I like this flexible approach as it does provide value to the consumer, and it takes a complex burden of hardcoding grammar for multiple languages.

There is an enhancement I wanted to surface for reviewers to discuss. I don't think that we necessarily need to do this for _this_ PR, but my inclination would be to create wrapper components for most of this JSX, just to remove the burden on the implementer from needing to interact with the uswds classes. That's one of the main/nice things about our library is that we encapsulate all of that into our components, generally. 

That would involve creating several subcomponents like `IdentifierContainer`, `IdentifierLogos`, `IdentifierAgencyDescription`.

If that is something we agree would be a good idea, I would be happy with creating a Github issue for the enhancement, and merging this as is for now. Then it wouldn't be a breaking change in the future, just a slightly more polished out-of-the-box set of sub-components.

Thoughts?

brandonlenz (commented):
A few comments, and one larger item I wanted to surface for discussion amongst everyone.

src/components/Identifier/Identifier/Identifier.stories.tsx
brandonlenz: And @SirenaBorracha just to be clear, on this point I'm definitely not expecting you to make any changes for this at the moment. Curious what other reviewers thing as well.

src/components/Identifier/Identifier/Identifier.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+interface IdentifierProps {
+  className?: string
+  children: React.ReactNode
+}
+
+export const Identifier = ({
+  className,
+  children,
+}: IdentifierProps & JSX.IntrinsicElements['div']): React.ReactElement => {
```
suzubara, line 12: non-blocking suggestion: We may want to allow users to pass additional props to the `<Identifier>` component that get passed along to the `div`, like how `sectionProps` are used in the `IdentifierGov` component below

src/components/Identifier/IdentifierLogo/IdentifierLogo.tsx
```diff
@@ -0,0 +1,29 @@
...
+  children,
+  anchorProps,
+  className,
+  anchorClassName,
+}: IdentifierLogoProps): React.ReactElement => {
+  const anchorClasses = classnames(
+    'usa-identifier__logo',
+    className,
+    anchorClassName
+  )
```
suzubara, line 23: Since the anchor (`<a>`) is the intrinsic element for this component, can we structure this like we do with the other components where additional props (and `className`) are passed along directly to the `<a>` element? I don't think `anchorClassName` is needed here because `className` serves the same purpose, and `anchorProps` can be the used to define the "rest" of the props (ie, `...anchorProps`) instead of a single named prop. For example, to the user it's the difference between:
```
<IdentifierLogo anchorProps={{ ariaLabel: "My logo" }}> ...
```
and:
```
<IdentifierLogo aria-label="My logo"> ...
```

suzubara (approved):
This is looking great!! I made a couple of suggestions regarding the intrinsic element props, but ✅  for MVP :)

src/components/Identifier/Identifier/Identifier.stories.tsx
suzubara: @brandonlenz yup I definitely agree!

haworku (dismissed):
LGTM.

package.json
```diff
@@ -197,5 +197,6 @@
         "hidden": true
       }
     ]
-  }
+  },
+  "dependencies": {}
```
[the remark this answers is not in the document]
suzubara: oh no, I think this can be deleted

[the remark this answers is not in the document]
brandonlenz: I see this removed in https://github.com/trussworks/react-uswds/pull/1044/commits/8e813b8a681e1190d8e5d97effbb7f79e9ba68a7 now. Thanks @SirenaBorracha!

brandonlenz (approved):
Thanks for taking care of everyone's feedback!

PR #1045, opened 2021-03-23
docs(adr): Increase ReactUSWDS version when updating USWDS version

### LINKED ISSUE #343: [docs] Add note on USWDS versions
Add a brief section to our README that explains:

- what it means for ReactUSWDS to use one version of USWDS, and for a user’s app to potentially use the same one/a different one
- how to determine what version of USWDS we have fully parity with in terms of component support (we could reference milestones)

Capturing this from old discussions in slack.

suzubara:
# Summary

Adds an ADR to help us decide ReactUSWDS versioning as it relates to USWDS versions.

I also copied the ADR template @brandonlenz introduced in https://github.com/trussworks/react-uswds/pull/1026

## Related Issues or PRs

Resolves #343

### DIFF
--- a/docs/adr/0001-update-major-version-with-uswds-minor-version.md
+++ b/docs/adr/0001-update-major-version-with-uswds-minor-version.md
@@ -1,0 +1,100 @@
+# Increase ReactUSWDS version when updating USWDS version
+
+- Status: proposed
+- Deciders: Maintainers
+- Date: 2021-03-21
+
+## Context and Problem Statement
+
+As the primary driver of requirements and new development on this project, new releases of USWDS can often necessitate implementation changes of existing components in ReactUSWDS. USWDS is also listed as both a `devDependency` and `peerDependency` in this project, and these versions should always be in alignment. It is important that applications use the same version of USWDS that was used to build the version of ReactUSWDS they're using. A version mismatch may result in unexpected markup & CSS combinations.
+
+Therefore, any time we decide to update the version of USWDS used in ReactUSWDS, we should expect to perform the following changes:
+
+- Update the USWDS version listed in both `devDependencies` and `peerDependencies`
+- Perform any API or implementation changes to existing components detailed in the USWDS release notes
+  - For example, from [2.10.1](https://github.com/uswds/uswds/releases/tag/v2.10.1):
+    > Use proper semantic markup for footer title: Now we're using a p instead of an h3 for usa-footer\_\_logo-heading since the name of the project or agency is hot a heading in this context.
+
+The questions this ADR seeks to answer are:
+
+- When performing a USWDS version update as described above, what kind of release should that result in for ReactUSWDS? (i.e., major / minor / patch)
+  - At minimum, this update will always require users to also update their version of USWDS.
+- Should USWDS version updates be timed and released in sequence with the components they implement?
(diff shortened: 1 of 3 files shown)


docs/adr/0001-update-major-version-with-uswds-minor-version.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Decision Drivers
+
+- We want to provide as many USWDS components as possible to the widest audience as possible, taking into account users who may be version-locked to a legacy USWDS version for whatever reason
+- We don't want to impede development and release of USWDS components, and we aim to "catch up" to the current USWDS release eventually
+
+## Considered Options
+
+- Bump the ReactUSWDS major version when updating USWDS minor & major versions
+- Bump the ReactUSWDS version with the same release type as USWDS
```
ahobson, line 41: I vote for `Bump the ReactUSWDS version with the same release type as USWDS`. We probably should also acknowledge this means we don't anticipate maintaining multiple active versions of the library. For example, we wouldn't back port a new component or bug fix to a previous USWDS version.

brandonlenz, line 41: I'm comfortable with option 1 or option 2.

We know that `uswds` _does_ release presentational changes that are sometimes not mentioned in the release notes, which we _could_ consider breaking changes.

Reading the `uswds` release process makes me actually lean towards option 2. If our library is simply a react-wrapper for `uswds`, is there really any value to us being _more strict_ than `usdws`? Option 1 would mean that projects depending on both libraries would have minor version updates for `uswds` and a major version updates for `react-uswds` for the same changes. Specifying the package.json dependencies on a consumer project would be weird in that context. For example, if you use `^2.10.0` for `uswds`, you wouldn't have a good equivalent for `react-uswds` without also opening yourself up to other breaking changes. It kind of sounds like we're wanting consumers to manually manage the versions of USWDS and React, so maybe that concern isn't as big of a deal as I'm making it sound like.

Related note (to option 2): 

I would've thought based on 

> the following should trigger a major version increment:
> - Changing the way in which elements with .usa- class names are structured in HTML

that the 2.10.1 change (a _patch_ version change!)

> Now we're using a `p` instead of an `h3` for usa-footer\_\_logo-heading

would have constituted a **major** version change.

But ultimately for me its coming back to "What does being _more_ strict than the library we're wrapping ultimately buy us?". If consumers are expected to manually manage these versions, the stability of this library's versions isn't a big of a concern.

haworku: Great point I like how you framed that.

haworku:
Based on today's meeting I think we have consensus around option 2 🎉. Thank you @suzubara for gathering these options. 

Next steps for implementing this ADR could be:
- Document the decision and publish
- Add section on USWDS version change to our docs. Maybe something like "Guidelines for changing USWDS version in library". 
- We also said in our meeting we expect USWDS version bump to happen with a separate PR/ticket rather than being combined into the work for a specific ticket.   We should probably briefly document this point as well. Maybe we can a uwsds version label for those PRs.

haworku (approved):
Our first ADR 😭 😍 . Look at us deciding things.

haworku:
Suggestion: also add this line to the first page of our README so its front and center.

`It is suggested applications use the same version of USWDS that was used to build the version of ReactUSWDS they're using. A version mismatch may result in unexpected markup & CSS combinations.`

Issue #982, opened 2021-03-01
[feat]: Radio Button Tile Variant


brandonlenz:
**Does your feature request relate to a specific USWDS component?**

[Radio Button](https://designsystem.digital.gov/components/radio-buttons/)

**What USWDS Version is this feature present in?**

2.10.0

**Describe the solution you'd like**

Add a tile variant to the radio button component (accomplished via addition of the uswds class `usa-radio__input--tile`)

The solution should be unified in terms of implementation and interface to https://github.com/trussworks/react-uswds/issues/790

Shkeating:
just wanted to mention i have an issue on my project where I'd need this if it helps at all with prioritization!

brandonlenz:
@Shkeating this will be in the next release!

PR #1101, opened 2021-04-07
feat: Radio Button Tile Variant

mlm55:
# Summary
Adds a tile version of the radio button

## Related Issues or PRs

closes: [#982](https://github.com/trussworks/react-uswds/issues/982)

## How To Test

`yarn test`
`yarn storybook`
Viewable from Form controls > Radio buttons > Tile/Tile With Label Description

### Screenshots (optional)
Screen Shot 2021-04-07 at 12 47 49 PM
Screen Shot 2021-04-09 at 3 47 57 PM

### DIFF
--- a/src/components/forms/Radio/Radio.stories.tsx
+++ b/src/components/forms/Radio/Radio.stories.tsx
@@ -39,7 +39,7 @@ export const WithLabelDescription = (): React.ReactElement => (
     id="input-radio"
     name="input-radio"
     label="My Radio Button"
-    labelDescription="Optional label description"
+    labelDescription="This is optional text that can be used to describe the label in more detail."
   />
 )
 
@@ -52,7 +52,7 @@ export const tileWithLabelDescription = (): React.ReactElement => (
     id="input-radio"
     name="input-radio"
     label="My Radio Button"
-    labelDescription="Optional label description"
+    labelDescription="This is optional text that can be used to describe the label in more detail."
     tile
   />
 )
--- a/src/components/forms/Radio/Radio.stories.tsx
+++ b/src/components/forms/Radio/Radio.stories.tsx
@@ -33,3 +33,7 @@ export const selected = (): React.ReactElement => (
 export const disabled = (): React.ReactElement => (
   <Radio id="input-radio" name="input-radio" label="My Radio Button" disabled />
 )
+
+export const tile = (): React.ReactElement => (
+  <Radio id="input-radio" name="input-radio" label="My Radio Button" tile />
+)
--- a/src/components/forms/Radio/Radio.stories.tsx
+++ b/src/components/forms/Radio/Radio.stories.tsx
@@ -34,6 +34,25 @@ export const disabled = (): React.ReactElement => (
   <Radio id="input-radio" name="input-radio" label="My Radio Button" disabled />
 )
 
+export const WithLabelDescription = (): React.ReactElement => (
+  <Radio
+    id="input-radio"
+    name="input-radio"
+    label="My Radio Button"
+    labelDescription="Optional label description"
+  />
+)
+
 export const tile = (): React.ReactElement => (
   <Radio id="input-radio" name="input-radio" label="My Radio Button" tile />
+)
+
+export const tileWithLabelDescription = (): React.ReactElement => (
+  <Radio
+    id="input-radio"
+    name="input-radio"
+    label="My Radio Button"
+    labelDescription="Optional label description"
+    tile
+  />
 )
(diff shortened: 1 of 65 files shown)


src/components/forms/Radio/Radio.tsx
```diff
@@ -20,14 +21,18 @@ export const Radio = ({
   className,
   label,
   inputRef,
+  tile,
   ...inputProps
 }: RadioProps & JSX.IntrinsicElements['input']): React.ReactElement => {
   const classes = classnames('usa-radio', className)
+  const radioClass = classnames('usa-radio__input', {
```
brandonlenz, line 28: Super minor, but can we make this `radioClasses`? I think its just a general convention I've noticed, and it makes sense since there could me multiple CSS classes applied.

brandonlenz (changes requested):
Nicely done! Also congrats on your first `react-uswds` PR! 🎉 

This looks good! Regarding Happo, it automatically "fails" until someone manually approves if it finds a diff with any existing components and/or if new diffs are added. In this case it is just showing the added diffs the tile story. That said, upon review we can see the tile style is not applied properly there:
image

To fix that, all you'll need to do is update the `uswds` dependency under `peerDependencies` and `devDependencies` in `package.json` from `2.9.0` to `2.10.0`. (There is a separate PR to do this among other things which will merge to main, but for now to get reviews on your PR it's okay to just update it on your branch)

brandonlenz (approved):
Looks good! Adding my approval here, though we'll have to hold off on the merge until we get the 2.9.0 stuff straightened out.

brandonlenz:
Re-requested my own review because, as Marjo rightly pointed out, there is an optional description prop added as well that we'll do as a part of this PR.

src/components/forms/Radio/Radio.tsx
```diff
@@ -36,6 +49,7 @@ export const Radio = ({
       />
       <label className="usa-radio__label" htmlFor={id}>
         {label}
+        {description}
```
brandonlenz, line 52: Same feedback as the checkbox PR: https://github.com/trussworks/react-uswds/pull/1104/files#r612477199

src/components/forms/Radio/Radio.tsx
```diff
@@ -12,6 +12,8 @@ interface RadioProps {
     | React.RefObject<HTMLInputElement>
     | null
     | undefined
+  tile?: boolean
+  labelDescription?: React.ReactNode
```
brandonlenz, line 16: I think this just needs to be a `string`

src/components/forms/Radio/Radio.stories.tsx
```diff
@@ -33,3 +33,26 @@ export const selected = (): React.ReactElement => (
 export const disabled = (): React.ReactElement => (
   <Radio id="input-radio" name="input-radio" label="My Radio Button" disabled />
 )
+
+export const WithLabelDescription = (): React.ReactElement => (
+  <Radio
+    id="input-radio"
+    name="input-radio"
+    label="My Radio Button"
+    labelDescription="Optional label description"
```
brandonlenz, line 42: Same comment as the other feedback: https://github.com/trussworks/react-uswds/pull/1104#discussion_r612470944

src/components/forms/Radio/Radio.tsx
mlm55: My thinking was that the label may end up having formatting, links, or other inline html. Would that still be possible if it's a `string`?

brandonlenz: Yea, you're right. People may want to at least bold some of the text in the label and things like that. Most of my hesitancy came from [`<span>` elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/span) taking  "[phrasing content](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#phrasing_content)" only. That does in include a bunch more things than just text, so we can leave it at `ReactNode` and the onus will be on the implementer to choose the right elements.

src/components/forms/Radio/Radio.tsx
mlm55: changed

src/components/forms/Radio/Radio.stories.tsx
mlm55: changed

brandonlenz (approved):
Thanks for doing this! (once you merge main and resolve conflicts I'll likely have to approve again, so just ping me whenever and I can do that 😄 )

suzubara (dismissed):
looks great!

brandonlenz (approved):
Approving again as merge conflicts are resolved

PR #1110, opened 2021-04-11
feat: Update Table to 2.10.0 implementation 

### LINKED ISSUE #991: [feat] Update Table to USDWS 2.10.0 implementation

[Table](https://designsystem.digital.gov/components/table/)

2.10.0

**Context**
[USWDS 2.10.0](https://designsystem.digital.gov/about/releases/#version-2100) Added several new table variations for mobile and data friendly display:

- scrollable
- striped
- compact
- stacked
- stacked-header

Add new props to implement the new classes/style variations. Something like below:

```javascript
scrollable?: boolean //default false
striped?: boolean //default false
compact?: boolean //default false
stackedStyle?: 'none' | 'default' | 'headers' // default 'none'
```

I combined the stacked and stacked-header modifiers into a single prop because the two appear to be mutually exclusive. Implementers will use it as follows:

```jsx
<Table stackedStyle="default"
``` 
OR

```jsx
<Table stackedStyle="headers" ...
``` 

Add storybook stories and unit tests for the different variations and verify the existing variations continue to behave as they should.

This might be a good candidate to add a storybook controls demo story 🤔 

If there is a more effective and intuitive way to provide access to these table styles, feel free to go that route! Especially considering the stacked-header implementation. There may be a neater way to handle that.

SirenaBorracha:
# Summary

This PR updates the [USWDS Table component](https://designsystem.digital.gov/components/table) with `scrollable`, `striped`, `compact`, `stacked`, and `stacked-header` variations as added in the [USWDS 2.10.0 release](https://designsystem.digital.gov/about/releases/#version-2100).

## Related Issues or PRs

closes #991 

* Is there a better way than putting `stackedStyle = 'none'` in the function signature to assign a default value?

This PR does not add the [Sortable Table rows](https://designsystem.digital.gov/components/table/#sortable-table-rows) variant as that is future work to cover updates made by the USWDS 2.11.0 release.

Reviewers may notice some styling differences between our Storybook versus the USWDS examples, particularly with regards to fonts/text sizing. As [Hana has noted](https://github.com/trussworks/react-uswds/pull/1110#issuecomment-820733923) further down in this PR, we suspect this is due to the difference in how USWDS applies their CSS to how we apply it, and is less relevant to this work as we are simply rendering those applied styles on children, not the Table component itself. 

## How To Test

Check out this branch and run `yarn test` to execute tests and `yarn storybook` to see the new component in action. 

#### Scrollable Table
scrollable-table

#### Striped Table

Screen Shot 2021-04-15 at 3 10 05 PM

#### Compact Table

Screen Shot 2021-04-15 at 3 10 54 PM

#### Logged warning if the user attempts to apply both `compact` and `stackedStyle`

From the USWDS notes re: compact styling: "Pairs well with scrollable and striped variants, but is not suitable for use with stacked variants."

Screen Shot 2021-04-13 at 2 52 12 PM

#### Stacked Table With Custom Controls `bordered` and `stackedStyle`

stackedStyles


src/components/Table/Table.tsx
```diff
@@ -17,18 +21,39 @@ export const Table = ({
...
+            {children}
+          </table>
+        </div>
+      ) : (
+        <table className={classes}>
+          {caption && <caption>{caption}</caption>}
+          {children}
+        </table>
+      )}
+    </>
```
brandonlenz, line 56: If you extract out table into a const, then you won't need to repeat it in your ternary:

```TSX
const table = (
  <table className={classes} data-testid="table">
    {caption && <caption>{caption}</caption>}
    {children}
  </table>
)
```

```suggestion
    <>
      {scrollable ? (
        <div
          data-testid="scrollable-table"
          className="usa-table-container--scrollable">
          {table}
        </div>
      ) : (
        {table}
      )}
    </>
```

src/components/Table/Table.tsx
```diff
@@ -17,18 +21,39 @@ export const Table = ({
...
 
   return (
-    <table className={classes} data-testid="table">
-      {caption && <caption>{caption}</caption>}
-      {children}
-    </table>
+    <>
+      {scrollable ? (
+        <div
+          data-testid="scrollable-table"
```
brandonlenz, line 43: Super minor, but since this isn't the actual table, its a container div, maybe it should be called something like `scrollable-table-container`

src/components/Table/Table.stories.tsx
```diff
@@ -178,3 +192,60 @@ export const fixed = (): React.ReactElement => (
     </tbody>
   </Table>
 )
+
+export const scrollable = (): React.ReactElement => (
+  <Table scrollable={true} caption="This is a scrollable table">
+    {testContent}
```
brandonlenz, line 198: In the USWDS example, Scrollable table uses different data than the other tables, presumably because of its intended purpose:

> A scrollable table is ideal for dense, tabular data.

Can you create a set of test data for the scrollable table by copying what USWDS uses in their example to use in this story?

brandonlenz (commented):
Nice! Thanks for making these enhancements. A few small things

src/components/Table/Table.stories.tsx
```diff
@@ -178,3 +192,60 @@ export const fixed = (): React.ReactElement => (
     </tbody>
   </Table>
 )
+
+export const scrollable = (): React.ReactElement => (
+  <Table scrollable={true} caption="This is a scrollable table">
```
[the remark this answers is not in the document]
brandonlenz: Ah. I see some of our existing examples use the `booleanProp={true}` syntax already. I'd personally update those to the shorthand as well to unify everything.

haworku:
Note: some minor differences in font in this storybook from USWDS examples may be related to the fact that we don't bring in uswds scss in full to our storybook. Similar to https://github.com/trussworks/react-uswds/issues/191#issuecomment-741973845

haworku:
I think using Storybook viewport would be a great way to showcase stacked styles

package.json
```diff
@@ -62,6 +62,7 @@
   "peerDependencies": {
     "react": "^16.x || ^17.x",
     "react-dom": "^16.x || ^17.x",
+
```
brandonlenz, line 65: I think you can remove the blank lines in this file. Looks like a result of resolving merge conflicts.

src/components/Table/Table.stories.tsx
```diff
@@ -178,3 +325,70 @@ export const fixed = (): React.ReactElement => (
...
+        <td data-label="Year">1791</td>
+      </tr>
+    </tbody>
+  </Table>
+)
+stackedStylesDemo.parameters = {
+  viewport: {
+    defaultViewport: 'iphone6',
+  },
+}
```
brandonlenz, line 394: This is cool! 🎉

brandonlenz (commented):
One tiny thing and I'm good to approve. Glad you and @haworku were able to get some clarity into that cell styling. I'm good with not worrying about it too much as well.

yarn.lock
```diff
@@ -13944,10 +13944,12 @@ use@^3.1.0:
   resolved "https://registry.yarnpkg.com/use/-/use-3.1.1.tgz#d50c8cac79a19fbc20f2911f56eb973f4e10070f"
   integrity sha512-cwESVXlO3url9YWlFW/TA9cshCEhtu7IKJ/p5soJ/gGpj7vbvFrAY/eIioQ6Dw23KjZhYgiIo8HOs1nQ2vr/oQ==
 
+
```
brandonlenz, line 13947: These look like the result of a manual conflict resolution on this file. Since it's autogenerated, it's preferred to run `yarn` locally and commit + push to determine the content of this file.

brandonlenz (approved):

PR #1132, opened 2021-04-12
Add Happo as GH action

suzubara:
# Summary


haworku:
@suzubara  This works - is there a reason we wanted Happo as a Github action instead of the way we are currently running? This PR  looks like it could merge but I would want to turn off the Happo tests in our github settings so we are only running through the action.

haworku:
❤️ closing as stale. feel free to reopen

PR #1107, opened 2021-04-09
feat: New Component ProcessList MVP

SirenaBorracha:
# Summary

This PR adds the [USWDS ProcessList component](https://designsystem.digital.gov/components/process-list/) in an MVP state.

## Related Issues or PRs

#979 

An enhancement that we may want to add as part of this GitHub issue would be to create a `ProcessListItemHeader` subcomponent. This would allow us to offer an out of the box way for our users to pass a heading that would automatically be styled by the `usa-process-list__heading` class irrespective of what element type the user passed. 

This enhancement would meet the USWDS Accessibility guideline, "Use semantic heading levels," that states, "Though our default code uses an `<h4>`, use the correct heading level with the class name `usa-process-list__heading` in your own implementation."

## How To Test

Check out this branch and run yarn test to execute tests and yarn storybook to see the new component in action.

Alternatively, you can scroll to the bottom of this page and click "Show environments" on the right above the comment input box to access this component in Storybook. 

#### Default
Screen Shot 2021-04-08 at 5 46 36 PM

#### No text, custom sizing
Screen Shot 2021-04-08 at 5 48 18 PM

#### Custom Sizing
Screen Shot 2021-04-08 at 5 48 45 PM

PR #1162, opened 2021-04-20
feat: implement ProcessListHeading subcomponent

### LINKED ISSUE #1134: [feat] Implement ProcessList subcomponent

[ProcessList](https://designsystem.digital.gov/components/process-list/)

2.10.0

An enhancement for the MVP ProcessList component added by GitHub issue #979 would be to provide users an out of the box header component with USWDS classes already applied. This is [discussed in the comments](https://github.com/trussworks/react-uswds/pull/1107#pullrequestreview-633757283) of the [MVP ProcessList PR](https://github.com/trussworks/react-uswds/issues/979).

1. Create a `ProcessListItemHeader` subcomponent that will enable users to encapsulate the uswds class(es) instead of requiring consumers to manually implement them.
2. Update the storybook examples to use this subcomponent

The alternative is to not provide our users with this subcomponent, and instead leave the ProcessList as flexible as possible, leaving the implementation of their header structure and USWDS classes to the user.

SirenaBorracha:
# Summary

This PR implements the `ProcessListHeading` subcomponent so that users have an out of the box heading component and do not have to manually apply the USWDS heading className.

## Related Issues or PRs

closes #1134 

As the MVP ProcessList PR https://github.com/trussworks/react-uswds/pull/1107 has not yet been merged, I made that branch the target for this PR.

## How To Test

### Screenshots (optional)

Nothing is changed visually in this PR from the MVP ProcessList component PR https://github.com/trussworks/react-uswds/pull/1107


trussworks-infra:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 315c8f30a443252c9119f24de0ca65e5a644b36c
</p>

brandonlenz (commented):
So we talked about this earlier today. I had a comment written out for a potential refactor. So I'm going to kind of re-write that with some updates from what we learned for discussion. My original thoughts were that this seems to be a bit over engineered, IMO. Since we know that the header will be 'h1' - 'h6' or a 'p' tag', I don't think we should have to worry about custom components or anything like that. Since the implementer is going to be able to flexibly compose whatever they want into a ProcessList, if they want to use a custom component they can do so natively. This should exist as a convenience component for out of the box functionality. 

As much as I wanted to just be able to recommend a much simpler and more concise way to achieve this without any problems... well, its harder than it sounded, mostly due to trying to dynamically include JSX.IntrinsicElements for the given heading type. 

What I've stumbled into that I'm _almost_ happy with is the following:

```TSX
type ProcessListHeadingProps = {
  type: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | 'p'
  className?: string
  children?: React.ReactNode
}

export function ProcessListHeading({
  type,
  className,
  children,
  ...headingProps
}: ProcessListHeadingProps &
  (
    | React.HTMLAttributes<HTMLParagraphElement>
    | React.HTMLAttributes<HTMLHeadingElement>
  )): React.ReactElement {
  const classes = classnames('usa-process-list__heading', className)
  const Tag = type
  return (
    <Tag className={classes} {...headingProps}>
      {children}
    </Tag>
  )
}

```

The only caveat here is that it would be possible to pass `p`-only intrinsic elements to an `h1`-`h6` header (or vice-versa). I typically like to make code impossible to use incorrectly so maybe I'm overthinking it. Since React is declarative, the implementer should also be able to use their judgement to generate the desired markup following whatever convention they prefer.

brandonlenz:
> The only caveat here is that it would be possible to pass `p`-only intrinsic elements to an `h1`-`h6` header (or vice versa). I typically like to make code impossible to use incorrectly so maybe I'm overthinking it. Since React is declarative, the implementer should also be able to use their judgement to generate the desired markup following whatever convention they prefer.

So that caveat was irking me so I spent some more time trying to make this work because it *should* be possible, and I think I cracked it:

```TSX
interface BaseProcessListHeadingProps {
  type: string
  className?: string
  children?: React.ReactNode
}

interface HeadingProcessListHeadingProps extends BaseProcessListHeadingProps {
  type: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6'
}

interface ParagraphProcessListHeadingProps extends BaseProcessListHeadingProps {
  type: 'p'
}

type ProcessListHeadingProps = HeadingProcessListHeadingProps &
  React.DetailedHTMLProps<
    React.HTMLAttributes<HTMLHeadingElement>,
    HTMLHeadingElement
  >

type ProcessListParagraphHeadingProps = ParagraphProcessListHeadingProps &
  JSX.IntrinsicElements['p']

export const ProcessListHeading = ({
  type,
  className,
  children,
  ...headingProps
}:
  | ProcessListParagraphHeadingProps
  | ProcessListHeadingProps): React.ReactElement => {
  const classes = classnames('usa-process-list__heading', className)
  const Tag = type
  return (
    <Tag className={classes} {...headingProps}>
      {children}
    </Tag>
  )
}

```

Typescript is now able to recognize that if type is `'p'`, only `'p'` tag intrinsic elements are allowed. 🥳 

Even though there's some _slightly_ complex typing going on here, this would be a less cumbersome solution that matches the simplicity that a basic heading component should have, in my opinion. Thoughts?

brandonlenz:
Oh, do you want this branch targeting main?

src/components/ProcessList/ProcessList/ProcessList.stories.tsx
```diff
@@ -66,47 +69,47 @@ export const processListDefault = (): React.ReactElement => (
...
 )
 
 export const processListCustomSizing = (): React.ReactElement => (
   <ProcessList>
     <ProcessListItem className="padding-bottom-4">
-      <h4 className="usa-process-list__heading font-sans-xl line-height-sans-1">
-        Start a process.
-      </h4>
+      <ProcessListHeading type="h4" className="font-sans-xl line-height-sans-1">
+        Start a process
```
brandonlenz, line 93: Super nit-picky, but this just needs the `.` back to match USWDS

brandonlenz (approved):

PR #1163, opened 2021-04-21
refactor: Use React patterns instead of JS API in Tooltip component

### LINKED ISSUE #1027: [refactor] Minimize use of JS DOM API in the Tooltip component
The Tooltip component (added in https://github.com/trussworks/react-uswds/pull/871) uses a lot of the [USWDS vanilla JS source code](https://github.com/uswds/uswds/blob/master/src/js/components/tooltip.js). Since we're using React to build components, we can replace most (if not all) of that with React features, such as `useState, useRef, useEffect`, in order to follow React best practices and write declarative code instead of imperative.

**Suggested changes:**
- Replace the `setPositionClass` and `positionXYZ` functions with state values for storing the position and style attributes (which can be updated when the Tooltip is opened)
- Apply custom classes to the trigger and wrapper elements in the Tooltip component (instead of inside the `useTooltip` hook)

suzubara:
# Summary

I updated the Tooltip component so that it makes use of state and `useEffect` in order to render the appropriate CSS classes and styles when the Tooltip is displayed. I also added some more tests, and added some missing HTML attributes based on the [JS source](https://github.com/uswds/uswds/blob/2d8f8983d6f07427c98a112d4ee0c3c81b79ae32/src/js/components/tooltip.js).

Main code changes are:
- I consolidated some code that was repeated in the `if/else` block that handles the custom component to the top
- I handled toggling the wrap class as an effect when the `effectivePosition` changes
- I set position and inline styles as state values instead of using the JS DOM API

While working on this, I noticed the Tooltip component underwent some [more refactoring in USWDS 2.11.0](https://github.com/uswds/uswds/pull/4062/files). This PR does not include those changes, which I think should be done as a separate issue when we're working on 2.11 features.

## Related Issues or PRs

Closes #1027 

## How To Test

Test the Tooltip component, there should be no effective changes.


trussworks-infra:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 322051b2216f67bf210453814a5126b4f5beeba8
</p>

src/components/Tooltip/Tooltip.tsx
```diff
@@ -236,43 +210,50 @@ export function Tooltip<FCProps = DefaultTooltipProps>(
             break
         }
       }
-    }, [isVisible])
+    }
+  }, [isVisible])
```
brandonlenz, line 214: I _think_ this `useEffect` dependency is responsible for a small jitter that can be seen on the first hover, since the position state is not set until the tool tip is visible. Is it possible to set the position during initial render?


https://user-images.githubusercontent.com/15805554/115771319-ccf75c00-a37b-11eb-879c-a29fc01d2826.mov

brandonlenz (commented):
Looks great, and more React-y!

I noticed a couple of small things. Let me know your thoughts!

src/components/Tooltip/Tooltip.test.tsx
```diff
@@ -1,89 +1,400 @@
...
+
+      it('positions on the right', () => {
+        fireEvent.mouseEnter(screen.getByTestId('triggerElement'))
+
+        const bodyEl = screen.queryByRole('tooltip')
+        expect(bodyEl).toHaveClass('usa-tooltip__body--right')
+        expect(bodyEl).toHaveStyle('margin-bottom: 0px')
+        expect(bodyEl).toHaveStyle('margin-left: 457px')
+
+        // expect(bodyEl).toHaveStyle('bottom: ')
```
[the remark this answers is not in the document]
suzubara: good catch, it should not! :D

src/components/Tooltip/Tooltip.tsx
suzubara: It looks like USWDS gets around [this using a `setTimeout`](https://github.com/uswds/uswds/blob/v2.10.3/src/js/components/tooltip.js#L205), I updated it so that the `is-visible` class isn't turned on until _after_ the positioning values are updated (also using `useEffect`). let me know what you think!

haworku (dismissed):
Seems like an _effective_ use of `useEffect` - there's a lot in this PR but from what I can tell things are looking good. I played around with the tooltip in storybook a bit too

haworku:
I think we could merge this - running tests.

haworku (approved):

PR #1165, opened 2021-04-22
feat!: Expose FileInput component ref with a clearFiles method

### LINKED ISSUE #958: [feat] FileInput ref should expose a reset prop to clear its value

The [FileInput](https://github.com/trussworks/react-uswds/pull/900)

This is not related to a USWDS feature.

The `FileInput` exposes an `inputRef` prop in order to allow users access to the underlying `input` element. However, the file input element is unique in that its `value` [cannot be set programmatically](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#notes). This means that simply exposing a ref to the `input` element is not enough to allow a library user to reset the files selected by this input, which might be desirable if the user were implementing an async upload handler and wants to avoid rendering files in the FileInput component once they've been uploaded.

We should be able to use React's [ref forwarding](https://reactjs.org/docs/forwarding-refs.html) alongside the [useImperativeHandle](https://reactjs.org/docs/hooks-reference.html#useimperativehandle) hook to customize the ref that is returned by `FileInput` and expose a method to parent components that can be called to reset the file state to an empty array.

**Additional context**

We may want to start adopting this pattern for form control components that are more complex than a single `input` element (such as the ComboBox & DatePicker) and might also warrant exposing additional methods through a ref. This issue should help us get more understanding around this pattern, the use cases it fulfills, and how to move forward to ensure our form controls are highly adaptable for a variety of use cases.

This issue depends on https://github.com/trussworks/react-uswds/pull/900 and will likely result in a [major version update](https://reactjs.org/docs/forwarding-refs.html#note-for-component-library-maintainers).

suzubara:
# Summary

This PR modifies the `FileInput` component to use `React.forwardRef` and the `useImperativeHandle` hook to expose the underlying `<input type="file" />` element, a method to clear the files state, and the value of the `files` state.

## Related Issues or PRs

Fixes #958 

## How To Test

In an application, try passing a `ref` into the FileInput component, and then use `ref.current.files` and `ref.current.clearFiles()` to interact with the file input.

### Screenshots (optional)

Apr-23-2021 11-53-41


src/components/forms/FileInput/FileInput.tsx
```diff
@@ -12,164 +12,186 @@ interface FileInputProps {
...
+    const internalRef = useRef<HTMLInputElement>(null)
+    const [isDragging, setIsDragging] = useState(false)
+    const [showError, setShowError] = useState(false)
+    const [files, setFiles] = useState<FileList | []>([])
+
+    useImperativeHandle(
+      ref,
+      () => ({
+        input: internalRef.current,
+        clearFiles: (): void => setFiles([]),
```
haworku, line 47: 🤔 Should this also surface a value that says if the input has an error? Right now we manually duplicate code from `preventInValidFiles` for determining if the input has an error when invalid files for `accept` were dropped. Maybe there should be the option to say `fileInputRef.current.hasErrors` ? 

This matter because on those cases we don't want to clear the input or handle our async vents - we want to keep it in the invalid state and show the built-in file input errors for `showError`.

src/components/forms/FileInput/FileInput.tsx
```diff
@@ -12,164 +12,186 @@ interface FileInputProps {
...
-          file={files[parseInt(`${i}`)]}
-        />
-      )
-    }
-  }
+export type FileInputRef = {
+  clearFiles: () => void
+  input: HTMLInputElement | null
+  files: FileList | []
+}
```
haworku, line 21: A couple things:
- Make sure we export this type from `index.ts`. Its not right now.
- `files` types `FileList` or `[]` was causing me issues (type errors, not wanting to handle the empty array).  In my understanding, a `FileList` can be empty. Why do we have an empty array type here as well?

src/components/forms/FileInput/FileInput.stories.tsx
```diff
@@ -108,13 +108,15 @@ export const disabled = (): React.ReactElement => (
   </FormGroup>
 )
 
-export const withCustomHandlers = (argTypes): React.ReactElement => {
+export const withRefAndCustomHandlers = (argTypes): React.ReactElement => {
   const [files, setFiles] = useState<FileList>()
-  const fileInputRef = useRef<HTMLInputElement>()
+  const fileInputRef = useRef<FileInputRef>()
```
haworku, line 113: FYI Typescript in our project (may be a stricter setting) doesn't like this.   Had to set it up as `const fileInputRef = useRef<FileInputRef>(null)`

`Type 'FileInputRef | undefined' is not assignable to type 'FileInputRef | null'.
        Type 'undefined' is not assignable to type 'FileInputRef | null'.ts(2322)`

That conflict is with the type for `FileRefInput['input']'`

src/components/forms/FileInput/FileInput.tsx
```diff
@@ -12,164 +12,186 @@ interface FileInputProps {
...
+    }
 
-  const handleChange = (e: React.ChangeEvent<HTMLInputElement>): void => {
-    setShowError(false)
-    setFiles(e.target?.files)
-    if (onChange) onChange(e)
-  }
+    const handleChange = (e: React.ChangeEvent<HTMLInputElement>): void => {
+      setShowError(false)
+      setFiles(e.target?.files || [])
```
haworku, line 135: This is causing issues in my code because I handle only FileList which is the behavior of the normal input.

haworku: Also, there may be state update issue. On initial drop, `fileInputRef.current` consistently doesn't register the files. Only on the second drop, when a `FileList` is returned,  does it register that the files have changed.  

I'll push up my code so you can see it in our storybook (which builds by branch).

src/components/forms/FileInput/FileInput.tsx
```diff
@@ -12,164 +12,186 @@ interface FileInputProps {
...
+            ref={internalRef}
+            type="file"
+            data-testid="file-input-input"
+            name={name}
+            id={id}
+            className="usa-file-input__input"
+            disabled={disabled}
+            onChange={handleChange}
+            multiple={multiple}
+            accept={accept}
```
haworku, line 191: Getting a type error here  for `enterKeyHint` being a missing prop as soon I added the new FileInput to the application. 

Adding that prop with the value of `enter` (default) resolved the issue. It sounded [similar to this issue](https://github.com/pluralsight/design-system/commit/eb920f5a7bdd6571d578a601a629d8d411fd07f0). 🤔 Not sure what to make of this.

haworku (commented):
Took a look - lets screenshare if it seems useful

src/components/forms/FileInput/FileInput.tsx
suzubara: Ok, I'm pushing a change that modifies `FileInput` so that `files` state is an array of files, rather than a `FileList` interface. I think that will resolve the issue of how to handle an "empty" file state, given you cannot construct or modify a `FileList`.

fun fact for reference, the w3c spec says the FileList interface is "at-risk" of being replaced with an array anyways! https://w3c.github.io/FileAPI/#filelist-section

src/components/forms/FileInput/FileInput.stories.tsx
suzubara: I honestly can't remember why I used `null` instead of `undefined` (on [this line](https://github.com/trussworks/react-uswds/pull/1165/files#diff-39e22af5b23c82a7bd3b2c4dde1d5835d1ab01ab4f1173b9290c9a42b1477e0dR36)). I'm changing to `undefined`, let me know if that resolves!

src/components/forms/FileInput/FileInput.tsx
haworku: Ohh okay that's good to know.

src/components/forms/FileInput/FileInput.tsx
haworku: Oops this was an issue in our code. SMH excuse me.

haworku (approved):
This works in mc-review. ⭐ The key thing missing in our code was changing everything to use `ref.current?.input?.files` instead of ` ref.current.files`.

src/components/forms/FileInput/FileInput.tsx
haworku: the state weirdness was resolved by only referencing ` fileInputRef.current.input` instead.

brandonlenz:
I updated the title to include a "!" for the breaking change so that the squash will automatically preserve that. Feel free also when merging to add a helpful blurb about how to incorporate this change in the body of the squash commit message.

brandonlenz:
> This works in mc-review. ⭐ The key thing missing in our code was changing everything to use `ref.current?.input?.files` instead of ` ref.current.files`.

Maybe a question more for my understanding, but the FileInputRef currently looks like this:
```TSX
export type FileInputRef = {
  clearFiles: () => void
  input: HTMLInputElement | null
  files: File[]
}
```

shouldn't `files` be accessible from the ref directly as you were attempting to do?

brandonlenz:
We may want to adjust the storybook story to make sure that the ref's files prop we are exposing is what is being used. This can be a future item since the functionality of the Component will not need to change.

PR #1166, opened 2021-04-22
feat: Update Grid components to render any type of element

### LINKED ISSUE #160: Grid components should take Component prop
Right now, the Grid components (`Grid, GridContainer`) render a `div`. They should be modified to accept an optional `Component` prop (probably of type `ReactNode`) in case the user wants to render some other element (such as a `ul`). They should still default to `div` if no `Component` props is defined.

SirenaBorracha:
# Summary

This PR refactors `Grid` and `GridContainer` to be able to render any type of element with their default as a `div`

## Related Issues or PRs

closes #160 

## How To Test

### Screenshots (optional)

Nothing is changed visually in this PR from the existing Grid components.


src/components/grid/Grid.stories.tsx
```diff
@@ -37,6 +61,86 @@ export const defaultContainer = (): React.ReactElement => (
...
 )
 
+export const customContainer = (): React.ReactElement => (
+  <GridContainer<CustomGridContainerProps> asCustom={CustomGridContainer}>
+    <li>
+      <Grid row>
+        <Grid col={1}>{testContent}</Grid>
+        <Grid col={11}>{testContent}</Grid>
+      </Grid>
+    </li>
```
brandonlenz, line 71: I have some thoughts on this, but they're kinda hard to articulate, because the use case for doing this is still a bit weird to me.

There's nothing inherently _wrong_ with putting Grid components inside an `li` in this case, but it is basically a perfect opportunity to use the custom Grid component as a `li` anyways.

I see you have a story for the custom Grid item as well, so you have that covered. My thinking at least is to just combine the two into a single `customElements` story where both the container and the grid items are custom elements (container can be `ul` and grid can be `li` maybe?)

SirenaBorracha: I love this idea!

src/components/grid/Grid.stories.tsx
```diff
@@ -37,6 +61,76 @@ export const defaultContainer = (): React.ReactElement => (
...
 )
 
+export const customElements = (): React.ReactElement => (
+  <GridContainer<CustomGridContainerProps> asCustom={CustomGridContainer}>
+    <li>
+      <Grid<CustomGridProps> asCustom={CustomGrid}>
+        <Grid col={11}>{testContent}</Grid>
+        <Grid col={2}>{testContent}</Grid>
+      </Grid>
+    </li>
```
brandonlenz, line 71: Now that you've updated your custom grid to be a `li`, this is going to render `li` nested within a `li`. Does it work you you get rid of the parent `li` tags like so?:

```suggestion
    <Grid<CustomGridProps> asCustom={CustomGrid}>
      <Grid col={11}>{testContent}</Grid>
      <Grid col={2}>{testContent}</Grid>
    </Grid>
```

brandonlenz (approved):
Nicely done! looks great to me

PR #1192, opened 2021-04-27
feat: Update Search component to support i18n

### LINKED ISSUE #1103: [feat] Add i18n support to Search Component

[2.10.0 Search](https://github.com/uswds/uswds/pull/3861/files)

[2.10.0](https://github.com/uswds/uswds/releases/tag/v2.10.0)

In the 2.10.0 patch notes, a Spanish variant of the search component was added. Similarly to https://github.com/trussworks/react-uswds/issues/866, localization should be added to `Search`

The solution should be in line with the solution to https://github.com/trussworks/react-uswds/issues/866

SirenaBorracha:
# Summary

This is a quick update to the existing Search component to add support for i18n. I followed the pattern established in the [DatePicker component](https://github.com/trussworks/react-uswds/pull/990)

## Related Issues or PRs

closes #1103 

## How To Test

### Screenshots

Screen Shot 2021-04-26 at 5 52 29 PM


src/components/Search/Search.tsx
```diff
@@ -73,7 +73,7 @@ export const Search = ({
       />
       <Button type="submit">
         <span className={isSmall ? 'usa-sr-only' : 'usa-search__submit-text'}>
-          Search
```
brandonlenz, line 76: So while this works and may be perfectly acceptable, I can see a use case where someone may want to the Button text to be different from the label. 

There's also the thought of following convention for localizations which we established in #990. There we added an `i18n` object prop which contained all the localization values. Since we only have one value to worry about here, that might feel like a bit much.

_My_ overall preference with this is to follow convention, even if it feels like overkill, in the name of uniform codebase. Plus, to handle the scenario of wanting label and button text to be different we would have to introduce a new prop anyways, so that would be half way to implementing our established i18n pattern.

src/components/Search/Search.tsx
```diff
@@ -44,6 +50,8 @@ export const Search = ({
     deprecationWarning('Search property small is deprecated.  Use size')
   }
 
+  const buttonText = i18n?.buttonText || 'Search'
```
brandonlenz, line 53: If you want, you can even declare a default for `i18n` (such as the `EN_US` used in `DatePicker`). Since this is only one field within `i18n`, I'm good with it as is.

brandonlenz (approved):
Nicely done! 🎉

PR #1214, opened 2021-05-06
build(deps-dev): Update Typescript to 4.0+

### LINKED ISSUE #522: Use Typescript 4.0+
Typescript is moving on and we should too. Bump up our `package.json` to 4.0.0+ as part of our next **major** react-uswds release. This should be considered a breaking change and released with similar breaking changes.

kylehilltruss:
# Summary

Updates the React-USWDS Typescript version to 4.0 (actually 4.2.4, which is the most recent TS version, and also what [USWDS is using in their latest release](https://designsystem.digital.gov/about/releases/#version-2112)). This required updating the `@typescript-eslint` plugin and parser, which required changing some linter exception comments in a story file to solve an eslint error that arose with the plugin version change. 

The updated plugin also threw quite a few warnings around wanting `argTypes` parameters in story files to be typed, so I went ahead and added those, and fixed a few other minor type-checking complaints that existed elsewhere to get back to a warning-free baseline.

As we're updating a dependency major version, this is a BREAKING CHANGE.

## Related Issues or PRs

Closes #522.

## How To Test

* `yarn test` and `yarn lint`, and make sure everything comes back clean.
* In storybook, find a story that makes use of control values (the `Table` component has a few) and ensure that toggling control values updates the presentation of the component in the story canvas.


brandonlenz (approved):
Nice! Thanks for doing this, and for the clear testing procedure. Looks good to me as well!

Issue #1191, opened 2021-04-26
[feat] SiteAlert children enhancement


brandonlenz:
**Does your feature request relate to a specific USWDS component?**

Site Alert (2.10.x)

Currently SiteAlert requires children to be manually built and passed in. This means we are not encapsulating the (optional) `usa-alert__text` sub component.

**Describe the solution you'd like**

See the following thread for enhancement to SiteAlert (and Alert): https://github.com/trussworks/react-uswds/pull/1099#discussion_r609744998

Modifying `children` to allow `string | React.ReactElement | React.ReactElement[]` would allow us to render `string` inside the `usa-alert__text` internally, otherwise just render `children` directly.

A further enhancement would be the creation of the `AlertText` subcomponent which would allow encapsulation of the `usa-alert__text` class within a `p` tag based component, and also allow rich text via things like `<strong>` to be passed in (which would not be possible passing just a `string` as in the first enhancement).

**Describe alternatives you've considered**

There are a couple of options in the thread. Use discretion around preserving backwards compatibility, and how much time is available to

kylehilltruss:
What do folks think is the best location for a `AlertText` sub-component? This can certainly go alongside the `SiteAlert` component (since it's not a top-level part of the WDS in and of itself), but the same functionality can used in a similar way in the `Alert` component as well.

Right now, the Alert component is [tightly coupled](https://github.com/trussworks/react-uswds/blob/main/src/components/Alert/Alert.tsx#L47-L51) to _sometimes_ adding in the `usa-alert__text` class, but that could change if we wanted it to (though I'd worry about introducing a breaking change there).

brandonlenz:
> What do folks think is the best location for a AlertText sub-component? This can certainly go alongside the SiteAlert component (since it's not a top-level part of the WDS in and of itself), but the same functionality can used in a similar way in the Alert component as well.

> Right now, the Alert component is tightly coupled to sometimes adding in the usa-alert__text class, but that could change if we wanted it to (though I'd worry about introducing a breaking change there).

@kylehilltruss sorry I missed your question here from a few weeks ago.

It's a good question. I'm not aware of two separate, top-level components making use of the same sub-component in our codebase at the moment. I could see us doing something like putting `SiteAlert`, `Alert`, and `AlertText` in an `alert` directory within the project.

[...]

PR #1215, opened 2021-05-07
feat: SiteAlert child enhancement

kylehilltruss:
# Summary

Modifies the `SiteAlert` component to also accept simple text strings and multiple components as children. If the child is a simple string, the component will wrap it in `<p className="usa-alert__text">`, which aligns with what the [USWDS component](https://designsystem.digital.gov/components/site-alert/) is doing. (Single child components and multiple child components are deliberately left alone, as users may already be passing in content for the alert wrapped in a `p` tag.)

There is a desire to further enhance this functionality by creating an `AlertText` sub-component, but as I'm still trying to figure out how that could practically work (as it's possible to also use it in the `Alert` component as well but that might introduce a breaking change?), I'm making this intermediate PR since it's already a value add at this point.

## Related Issues or PRs

Addresses (but does not fully close, see above + issue discussion) #1191

## How To Test

Check out this branch, run `yarn test` to execute tests, and `yarn storybook` to see the updated component functionality in the additional stories for the `SiteAlert` component.


haworku (approved):

haworku (approved):
Thank you ⭐

PR #1236, opened 2021-05-10
feat!: Expose ComboBox component ref with a clearSelection method

### LINKED ISSUE #1137: [feat] Add a simple way to programmatically clear a ComboBox
Related to ComboBox.

react-uswds: 1.10

Unable to reset the ComboBox value.

Add a simple way to programmatically clear a ComboBox

// temporary fix to reset combobox box
`document.querySelectorAll('.usa-combo-box__clear-input').forEach((el) => (el instanceof HTMLElement ? el.click() : el));`

kylehilltruss:
# Summary

This changes the `ComboBox` component to use `forwardRef` and the `useImperativeHandle` hook. It allows users to pass in a ref to a component instance and then call a `clearSelection` method on it, which will clear any existing value.

This change is _deeply_ indebted to the pattern established in PR #1165.

As this changes an existing component to use `forwardRef`, this should be considered a BREAKING CHANGE.

## Related Issues or PRs

Closes #1137 .

## How To Test

In the application, try passing a `ref` to the ComboBox component, selecting a value for the component, and using `ref.current.clearSelection()` to reset the value.

### DIFF
--- a/src/components/forms/ComboBox/ComboBox.tsx
+++ b/src/components/forms/ComboBox/ComboBox.tsx
@@ -1,4 +1,11 @@
-import React, { KeyboardEvent, FocusEvent, useEffect, useRef } from 'react'
+import React, {
+  KeyboardEvent,
+  FocusEvent,
+  useEffect,
+  useRef,
+  forwardRef,
+  useImperativeHandle,
+} from 'react'
 import classnames from 'classnames'
 
 import { ActionTypes, Action, State, useComboBox } from './useComboBox'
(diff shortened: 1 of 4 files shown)


src/components/forms/ComboBox/ComboBox.tsx
```diff
@@ -78,372 +85,391 @@ const Input = ({
...
-
-  const containerRef = useRef<HTMLDivElement>(null)
-  const focusedItemRef = useRef<HTMLLIElement>(null)
-
-  useEffect(() => {
-    onChange && onChange(state.selectedOption?.value || undefined)
-  }, [state.selectedOption])
+export type ComboBoxRef = {
+  clearSelection: () => void
+}
```
brandonlenz, line 90: Awesome! 

Do you foresee the need to expose access to any of the child elements of combo box, similar to how the `input` for FileInput is exposed? I don't think that would be a breaking change if added in the future, so I'd be okay with a github issue for doing that as well.

src/components/forms/ComboBox/ComboBox.tsx
```diff
@@ -78,372 +85,391 @@ const Input = ({
...
         }
       }
-
-      if (!state.isOpen && state.selectedOption) {
+    }, [state.focusMode])
+
+    useImperativeHandle(
+      ref,
+      () => ({
+        clearSelection: (): void => dispatch({ type: ActionTypes.CLEAR }),
```
brandonlenz, line 185: I'm not sure if this is a bad thing or not, but it may be worth aligning on:

`ActionTypes.CLEAR` gives focus to the input. Should that be the default behavior here? I could see dragging focus to the input being problematic under certain circumstances (like a form where values are programmatically pre-filled and cleared as a user is progressing through the form. The clear action would pull them to a different section of the form potentially)

brandonlenz (commented):
The storybook example is really great. It makes testing this a breeze!

I did notice a behavior I'm not positive about on clear that I wanted to surface for discussion

src/components/forms/ComboBox/ComboBox.tsx
kylehilltruss: That's a good point, I hadn't really fully considered the implications of just glomming onto that existing action.

src/components/forms/ComboBox/ComboBox.tsx
kylehilltruss: I can't really foresee a need to expose the internal child elements? There's nothing that complicated going on under the hood, and the `clearSelection` method is just a way to make the component partially controlled. Curious what the use case might be for that.

But yeah, it's just adding the `forwardRef` that's the breaking change; additional functionality on the ref isn't breaking. I'm generally of the YAGNI opinion here; if someone asks for access to the child elements, we can just do it then.

src/components/forms/ComboBox/ComboBox.tsx
kylehilltruss: Updated to use a new ActionType.

src/components/forms/ComboBox/ComboBox.tsx
brandonlenz: Works for me!

src/components/forms/ComboBox/ComboBox.tsx
brandonlenz: Excellent. This feels in line with what I would expect as a consumer.

brandonlenz (approved):

PR #1241, opened 2021-05-14
feat: New Component Icon

### LINKED ISSUE #992: New Component: Icon
> :warning: **Note**: Creating this issue as a new "component" because it is listed by uswds as a component, _however_ there are many decisions to make around adding support internally and externally for the uswds icons. I'm going to compile some notes regarding my thoughts here, but before beginning work on this, I would suggest distilling via some discussion. The most important step in beginning this work will be to import uswds 2.10.0+ and see exactly how the icons are exposed to us, and how we can best wrap them. I don't _know_ that an Icon component is the way to go with this, but it seems like it could be good for us to use internally, and also expose for external use.

# [Icons](https://designsystem.digital.gov/components/icons/)

## USWDS Version

Icons have existed in USWDS previously to 2.10.0, but 2.10.0 added better default support and a broader selection of icons.

Not also that the icons have changed from fontawesome (also see not about docs in Other Considerations):
> **Active icons are found in** dist/img/usa-icons. The icons in this directory will be packaged into an SVG sprite. The complete set of available icons are located in two directories: dist/img/material-icons and dist/img/uswds-icons. Copy icons into a usa-icons directory within your project images directory and rebuild the SVG sprite using uswds-gulp.

## Description
Icons help communicate meaning, actions, status, or feedback.

## Requirements (proposed)
Implementation will largely depend on how the icons are externally exposed, and what importing them looks like. Below is an example thought

**Prop Interface**

```JSX
Props = {
  name: string // See bonus note below about this
  size?: 3 | 4 | 5 | 6 | 7 | 8 | 9 // Default `undefined` (scale to font size). Side note: Typescript ranges please https://github.com/Microsoft/TypeScript/issues/15480 
} & JSX.IntrinsicElements['svg']
```

The `name` prop should be used to determine the icon path. Icon names match the filename in the [uswds source code](https://github.com/uswds/uswds/tree/develop/src/img/usa-icons). By inserting `name` into the import url for the icon, we can allow the user to select an icon without needing to know anything about the path or internal implementation, so long as they use the correct name from the list of [uswds icons](https://designsystem.digital.gov/components/icons/#all-icons). Just be sure that a name that does not match an existing icon errors acceptably.

```JSX
<Icon name="accessibility_new" size={6} />
```

> ### Bonus: 
> As mentioned above, while `name` is a `string` it would be nice to also have type safety (and IDE autocomplete) where only strings matching actual icon names are allowed. 
>
> If uswds icons in 2.10.0 are exported statically in a format that we can accomplish this with, I think that would be worth pursuing.
> 
> I recommend discussion and using personal judgement here.

**More Details**

- `<svg>`root element
- Default class `usa-icon`
- 

## Other Considerations

- ~~There are several places in the code where we use uswds icons. GovBanner uses a lock icon for example. As a part of this effort, it would be a good idea to evaluate replacing our icons internally with the icon component (especially wherever we may still be using font-awesome).~~
- ~~In our [styles and assets documentation](https://github.com/trussworks/react-uswds/blob/main/docs/styles_and_assets.md) we're recommending using fontawesome per uswds' recommendation (which no longer appears to be there, in favor of their internal list of icons). Our documentation surrounding Icons should be updated alongside this PR~~
- ~~USWDS examples set some intrinsic props to defaults (e.g. `focusable` and `role`). We should set the same defaults, but allow override if the corresponding prop is passed to the `Icon` component~~

Updated some of these other considerations to be a part of their own Issue: https://github.com/trussworks/react-uswds/issues/1303

SirenaBorracha:
# Summary

This PR adds the [Icon component](https://designsystem.digital.gov/components/icon/).

## Related Issues or PRs

closes #992 

## How To Test

Check out this branch and run `yarn test` to execute tests and `yarn storybook` to see the new component in action. If you cannot see the Storybook controls dashboard, press the `D` key and it should appear on the right-hand side.

Alternatively, you can scroll to the bottom of this page and click "Show environments" on the right above the comment input box.

### Screenshots

11May-17-2021 12-01-57


package.json
```diff
@@ -133,6 +133,7 @@
   "resolutions": {
     "bl": "^4.0.3",
     "acorn-globals/acorn": "^6.4.1",
+    "@babel/preset-env": "7.13.15",
```
haworku, line 136: Can you please add something to the PR description about why are new dependency was needed? More for archive purposes if we ever need to remember in what context this was added its good to also have some description in the PR.

haworku (commented):
Looking good ⭐ . Running locally things worked fine.  I don't see the icons in Storybook in the deployment though - am I missing something?

SirenaBorracha:
> Looking good ⭐ . Running locally things worked fine. I don't see the icons in Storybook in the deployment though - am I missing something?

I'm not sure @haworku - I don't see them in the deployment either. I know it's something to do with how storybook interacts with the SVG sprite. 

The path of the `svg` `use` tag in the deployment:
 `href="static/media/sprite.483d6459.svg#accessibility_new"` 

versus locally: 
`href="static/media/node_modules/uswds/src/img/sprite.svg#accessibility_new"`

SirenaBorracha:
Drew and I spent some time pairing on this earlier today. We discovered a few things;`Icon` unexpectedly does render in all browsers in the added Happo screenshots (except in IE), we added `Icon` to the example app expecting it to work based on Happo's success but it did not, and there are rules around SVGs that differ from `image` tags which is how the rest of the components successfully render any USWDS icons as of right now.

- SVGs are stored in S3
- Icon works in Happo but does not work when imported and rendered in the example app in Chrome

haworku:
❤️ Really good work here. I feel like this component could be ready - but I'm not sure if we ready to merge this without storybook. I could see us going either way but going to close for now as stale since there hasn't been further discussion on the component for some time. @SirenaBorracha @ahobson if you have other context please share.

suzubara:
Closing in lieu of https://github.com/trussworks/react-uswds/pull/1609

PR #1266, opened 2021-05-25
fix: Use embedded images for storybook

### LINKED ISSUE #1265: Happo failures when referencing external images in storybook
We've been seeing Happo diffs when referencing external images. Try using base64 images in the stories to remove external dependencies

ahobson:
# Summary

Try embedded images to fix Happo diffs

## Related Issues or PRs

closes #1265

## How To Test

See if spurious Happo diffs go away


trussworks-infra:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 6f5897cb4dd7c0f297daa9105945ea5507b8b87b
</p>

src/components/card/Card.stories.tsx
```diff
@@ -265,7 +256,7 @@ export const cardTest = (): React.ReactElement => (
       <CardMedia imageClass="add-aspect-16x9">
         {/* eslint-disable-next-line jsx-a11y/img-redundant-alt */}
         <img
-          src="https://images.unsplash.com/photo-1583974098224-f64f67f46778?ixlib=rb-1.2.1&q=80&fm=jpg&crop=entropy&cs=tinysrgb&w=800&h=800&fit=crop&ixid=eyJhcHBfaWQiOjF9"
+          src={sparklerImage}
```
brandonlenz, line 259: Not sure if we care, but some of these images are not the original images (noticed it in Happo):
image

src/components/card/Card.stories.tsx
```diff
@@ -8,6 +8,21 @@ import { CardBody } from './CardBody/CardBody'
...
+
+const triangleImage =
```
brandonlenz, line 23: What are your thoughts on moving these to a separate file for storybook constants?

brandonlenz (commented):
Couple of questions before I approve, but I'm very excited for this to unblock dependabot updates 😄 🚀 🥳

src/components/card/Card.stories.tsx
ahobson: Yeah, I should have mentioned that! Some of the images seemed to be only used a few times and I didn't see a reason why we couldn't pare down our list.

If I'm wrong, I'm happy to change it

src/components/card/Card.stories.tsx
ahobson: I looked into that and wondered what the right approach is. We don't want these included in our final build and I think that would mean they should be excluded in `tsconfig.json`. Should we add `**/*.stories.ts`? Or put this in another `stores.tsx` file that doesn't have any components? Other ideas?

brandonlenz: Good points. Would it be worth having multiple tsconfig files like [this](https://stackoverflow.com/questions/58461649/how-to-exclude-specific-files-in-typescript-only-for-the-build)?

Our current tsconfig exclusions actually leave us with false positives for import paths for example:
image.

Moving all of those into a `tsconfig.publish.json` or something would resolve that small annoyance as well.

It starts to get funky trying to specify that specific tsconfig file for certain scripts, and that's about where my knowledge drops off.

> should we add `**/*.stories.ts`?

This is the easiest solution, and just requires some diligence with naming convention, but I like it.

brandonlenz: Hmmm you know what I didn't notice these are at least all on one line within the storybook file. I just saw the wrapping in the github UI. I'm good with it as is for getting this out the door. Its always something we can improve later.

brandonlenz (approved):

src/components/card/Card.stories.tsx
ahobson: Thanks for looking at this. I rebased it on my machine instead of doing that through the GitHub UI and now I need approval again. 😅

brandonlenz (approved):
@ahobson ✅ 

Looking forward to watching those dependabot PRs fly!

PR #1297, opened 2021-06-07
build(happo): disable edge browser because of false postives

ahobson:
# Summary

Disable edge browser in happo because it has too many false positives


brandonlenz:
@ahobson Thanks for this! I verified all the diffs in Happo for this PR are just the "deleted" edge diffs.

brandonlenz (approved):

Issue #1542, opened 2021-09-08
[feat] Add aria attributes to RangeInput


haworku:
**Does your feature request relate to a specific USWDS component?**
[Range Slider ](https://designsystem.digital.gov/components/range-slider/) - called `<RangeInput/ />` in our library

**What USWDS Version is this feature present in?**
2.12.1

**Is your feature request related to a problem? Please describe.**
Range slider didn't handle aria attributes until a recent [PR ](https://github.com/uswds/uswds/pull/4270)addressed this issue. 

Let's bring these improvements (in terms of having   `aria-valuemin`,   `aria-valuemax`,   `aria-valuenow` handled in our components, matching the `min`, `max` and `value` props) to our codebase.

PR #1560, opened 2021-09-13
feat: Add aria attributes to RangeInput

kimallen:
# Summary

Adds aria-min, aria-max, and aria-valuenow labels.  Also adds optional min and max as input props.

## Related Issues or PRs

closes [issue #1542](https://github.com/trussworks/react-uswds/issues/1542)

## How To Test

Tests should reflect added min/max props and corresponding aria values.
To test the value manually, run storybook (`yarn storybook`) and try out the different sliders with the developer tools --> Element open.  You should see `aria-min` and `aria-max` values.  When moving the slider, you should see the `aria-valuenow` changing.


trussworks-infra:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 51a83efdb7d921a5e18b99183790036974659262
</p>

src/components/forms/RangeInput/RangeInput.tsx
```diff
@@ -17,9 +19,30 @@ export const RangeInput = ({
...
   ...inputProps
 }: RangeInputProps & JSX.IntrinsicElements['input']): React.ReactElement => {
-  // Range defaults to min = 0, max = 100, step = 1, and value = (max/2) if not specified.
-
   const classes = classnames('usa-range', className)
+  // input range defaults to min = 0, max = 100, step = 1, and value = (max/2) if not specified.
+  const defaultMin = 0
+  const defaultMax = 100
+  const { min, max, defaultValue } = inputProps
+  const ariaMin = min || defaultMin
```
haworku, line 27: I think we should also handle a case where the consumer passes in their own `aria-` attributes as props.  We only want to use our handling as a fallback.

So  use here
 ```
const ariaMax = ariaMax || max || defaultMax
const ariaMin = ariaMin || max || defaultMax
```

File: src/components/forms/RangeInput/RangeInput.tsx [hunk 2]
```diff
@@ -17,9 +19,30 @@ export const RangeInput = ({
...
+    if (typeof value === 'number' || typeof value === 'string') {
+      return Number(value)
+    }
+    return undefined
+  }
+  const [ariaValue, setAriaValue] = React.useState<number | undefined>(
+    convertValueType(calculatedDefaultValue)
+  )
+  const onValueChange: React.ChangeEventHandler<HTMLInputElement> = (e) => {
+    setAriaValue(e.target.valueAsNumber)
```
haworku, line 44: I think we should also handle a case where the consumer passes in their own `aria-` attributes as props.  We only want to use our handling as a fallback.

So add conditional here
 ```
if (!ariaValueNow) setAriaValue(e.target.valueAsNumber)
```

haworku (commented):
Hey this looks good - all the elements are there and thanks for adding storybook at well.

Would you mind exploring a small code change and adding another test case for when the consumer does pass in their own `aria-` attributes? 

This would be in addition to the test already there. The way the code is written right now we would ignore those props if they passed them in. We usually want to handle cases where users are passing in props - described more [here](https://github.com/trussworks/react-uswds/blob/main/docs/adding_new_components.md#props)

Test cases I would love to see:
- "handles custom `aria-` attributes if passed in"
- "if default value is not present, uses midpoint between min and max as `aria-valuenow`"

src/components/forms/RangeInput/RangeInput.tsx
kimallen: done

src/components/forms/RangeInput/RangeInput.tsx
kimallen: done

kimallen:
I've added tests.
I'm getting this warning with tests, though not sure how to resolve (or why it's complaining):
`Warning: Invalid ARIA attribute `ariaValueNow`. Did you mean `aria-valuenow`?` at RangeInput.tsx line 21

haworku:
@kimallen Thanks for calling out those React warnings. They were caused by not using snake case `aria-valuenow` on the test case for "renders aria values" in `RangeInput.test.tsx`.

Interestingly, fixing that and causing tests to start failing.  Also started seeing type errors in `RangeInput`. This is why I love unit testing because sometimes it can reveal something we didn't see yet in the code. I ended up made changes to refer to the `aria-` attributes also in snake case (using this kind of syntax - `props['a-snake-case-html-attribute']`) to get everything working. 

Also added another step to our aria test to check that keyboard actions are updating the input.   Did that because I realized this is the way I was testing things manually.  Let me know your thoughts.

haworku (approved):
Looks good, I think we can merge!

PR #1588, opened 2021-09-24
chore(release): 2.1.0

haworku:
### Features

* SiteAlert child enhancement([#1215](https://github.com/trussworks/react-uswds/issues/1215)) ([3cd70f9 ](https://github.com/trussworks/react-uswds/commit/3cd70f95d04eab23e3ad6b1f31634d3b507b3692))
* Add aria attributes to RangeInput ([#1560](https://github.com/trussworks/react-uswds/issues/1560)) ([0034835](https://github.com/trussworks/react-uswds/commit/00348350ced514bf33c9a1795270838e2df33a24))

PR #1589, opened 2021-09-24
chore: Revert "chore(release): 2.1.0"

haworku:
Failed release - `yarn build` is is throwing an error.  We will fix and then retry release at this tag.
Reverts trussworks/react-uswds#1588


brandonlenz (approved):

brandonlenz:
Updated the PR title to make the PR lint pass. Not sure if that's what we want it to be though, so feel free to change as necessary

PR #1609, opened 2021-09-28
feat: Icon component using SVGR

### LINKED ISSUE #992: New Component: Icon

suzubara:
# Summary

## Related Issues or PRs

This PR implements the USWDS Icon component using an alternative approach, converting each icon to inline SVGs that can be rendered as React components instead of a single `sprite.svg` file with the `<use>` element. I worked off the existing PR (https://github.com/trussworks/react-uswds/pull/1241/files) and used most of the existing `Icon` component code in an HOC that wraps each React SVG component with the supported props (ie, `size`) and adds the USWDS classes.

Closes #992 

## How To Test

* View the Icons in the deployed Storybook for this branch and verify they match the icons listed at [USWDS](https://designsystem.digital.gov/components/icon/). Note that the thumbs down icon changed in the [2.11.2 release](https://github.com/uswds/uswds/releases/tag/v2.11.2), so that asset will appear differently until we've updated to that version of USWDS.
* Check out the branch locally and start the Example app, and view the Icons page. It should display all of the new icons.
* If you have another app that uses this library, try installing this branch and verify you can import and render icon components in your app.

### Screenshots (optional)

image


src/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,46 @@
...
+      role,
+      ...iconProps,
+    }
+
+    return <Component {...finalProps} />
+  }
+}
+
+export const AccessibilityNew: React.ComponentType<IconProps> =
+  makeUSWDSIcon(AccessibilityNewSvg)
```
ahobson, line 46: And so the plan would be to create one of these for each one of the icons from uswds, we at one point we had in the `icons.ts` file? Or did you want to try to automate the generation of this somehow?

ahobson (commented):
Fabulous work!

example/src/pages/Icons.tsx
```diff
@@ -1,19 +1,21 @@
...
   <section>
     <h1>Icons</h1>
-
     <p className="usa-intro">
-      We recommend using icons from <a href="https://github.com/FortAwesome/react-fontawesome">react-fontawesome</a> for projects that need icons.
+      We recommend using icons from{' '}
+      <a href="https://github.com/FortAwesome/react-fontawesome">
+        react-fontawesome
+      </a>{' '}
+      for projects that need icons.
```
brandonlenz, line 15: I assume we want to update this to remove references to fontawesome and keep things exclusive to uswds icons?

src/components/Icon/Icon.tsx
suzubara: as much as I would like to automate this, it's taking a whole new effort to figure out and I'm inclined to just do the cheap & easy thing for now (knowing new icons won't be added unless the USWDS version changes). I will create a followup issue for automating this, possibly using a webpack plugin or script??

ahobson: I'm totally on board with a manual process!

src/components/Icon/Icons.ts
```diff
@@ -0,0 +1,722 @@
+import AccessibilityNewSvg from 'uswds/dist/img/usa-icons/accessibility_new.svg?react'
```
suzubara, line 1: Note the query param `?react` at the end of each import. That is what's telling webpack to load these files using `svgr`, and all other SVGs are still loaded using the asset loader (so SVGs imported as images or in CSS should not break)

ahobson: That's neat! Total bikeshed, but would using `?svgr` maybe make it more obvious about what is going on? Feel free to ignore this comment :)

ahobson (dismissed):
This is amazing. Thank you

src/components/Icon/Icons.ts
suzubara: OH I like that idea! tbh I got this from following an example elsewhere and at first didn't understand the use of `?react`. I will change it to `?svgr` as I think that is much clearer.

brandonlenz (dismissed):
Seconding what Drew said. WOW! 

Happo Diff looked good too!

brandonlenz (approved):
Changes in d87838740e0d85aa53ad9166ec9524a3fbcfc0ed look good. Re-approved.

(Storybook deployment still looks great as well 🥳)

PR #1622, opened 2021-10-01
[WIP] feat: Modal component

suzubara:
# Summary


brandonlenz:
Should this PR be targeting `sr-1304-uswds-2.11` for now, given that we'll be merging that one first, so that the diff shows only the changes for this PR?

src/components/Modal/Modal.stories.tsx
```diff
@@ -0,0 +1,218 @@
...
+        aria-labelledby="modal-1-heading"
+        aria-describedby="modal-1-description">
+        <ModalHeading id="modal-1-heading">
+          Are you sure you want to continue?
+        </ModalHeading>
+        <div className="usa-prose">
+          <p id="modal-1-description">
+            You have unsaved changes that will be lost.
+          </p>
+        </div>
```
brandonlenz, line 53: I'm assuming this wasn't encapsulated via `header` and `text` props (or similar) to make the `Modal` body more flexible for consumers?

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,160 @@
...
+}: ModalProps): React.ReactElement => {
+  const [mounted, setMounted] = useState(false)
+  const initialPaddingRef = useRef<string>()
+  const tempPaddingRef = useRef<string>()
+  const modalEl = useRef<HTMLDivElement>(null)
+
+  const modalRootSelector = modalRoot || '.usa-modal-wrapper'
+
+  const NON_MODALS = `body > *:not(${modalRootSelector}):not([aria-hidden])`
+  const NON_MODALS_HIDDEN = `[data-modal-hidden]`
```
brandonlenz, line 43: Nice!

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,160 @@
...
+        el.removeAttribute('data-modal-hidden')
+      })
+    }
+  }, [isOpen])
+
+  const ariaLabelledBy = divProps['aria-labelledby']
+  const ariaDescribedBy = divProps['aria-describedby']
+
+  delete divProps['aria-labelledby']
+  delete divProps['aria-describedby']
```
brandonlenz, line 114: My understanding of `delete` leads me to believe that `ariaLabelledBy` and `ariaDescribedBy` would always be undefined here, and therefore also when passed in to the `ModalWrapper`. Is that not the case?

brandonlenz: When I run storybook examples, they are indeed passed along appropriately. So this works, and I just don't understand why. I can trace through that in my own time to understand it 👍

src/components/Modal/Modal.stories.tsx
```diff
@@ -0,0 +1,218 @@
...
+export const defaultModal = (): React.ReactElement => {
+  const { isOpen, openModal, closeModal } = useModal()
+
+  return (
+    <>
+      <ModalOpenButton
+        handleOpen={openModal}
+        href="#example-modal-1"
+        aria-controls="example-modal-1"
+        onClick={(e) => e.preventDefault()}>
```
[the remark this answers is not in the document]
suzubara: whoops, yes, I added one after the other and forgot to remove it 🤦‍♀️

src/components/Modal/Modal.stories.tsx
suzubara: > I'm assuming this wasn't encapsulated via `header` and `text` props (or similar) to make the `Modal` body more flexible for consumers?

Yup correct, ModalHeading & ModalFooter just apply the markup in the USWDS examples and they are totally optional to use if someone needs to do something more different

src/components/Modal/Modal.tsx
suzubara: this is because these 2 props get applied to the `ModalWrapper` component explicitly, with the rest of the props being passed into the `ModalWindow`

brandonlenz: Yea, I just assumed that `ariaLabelledBy` would become `undefined` when `delete divProps['aria-labelledby']` runs, assuming that `ariaLabelledBy` would hold a reference to `divProps['aria-labelledby']`, but that's just me not knowing exactly how JS behaves 😅

src/components/Modal/ModalOpenLink.tsx
```diff
@@ -0,0 +1,52 @@
...
+    'aria-controls': modalRef?.current?.modalId,
+    'data-open-modal': true,
+    onClick: handleClick,
+  } as DefaultLinkProps | CustomLinkProps<FCProps>
+
+  if (isCustomProps(linkProps)) {
+    return <Link<FCProps> {...linkProps} />
+  }
+
+  const definitelyLinkProps = linkProps as DefaultLinkProps
```
suzubara, line 49: This seems to fix the Typescript error that we started seeing in [this update](https://github.com/trussworks/react-uswds/pull/1548) - I'm not sure why the overload isn't correctly narrowing this type after the `isCustomProps` check but I think that is a separate issue to resolve.

src/components/Modal/utils.ts
```diff
@@ -0,0 +1,74 @@
...
+          return false
+        }
+      }
+    }
+
+    return true
+  }
+
+  const toggleModal = (e?: React.MouseEvent, open?: boolean): boolean => {
+    // console.log('TOGGLE MODAL', e, open)
```
brandonlenz, line 34: nit: should this be removed?

src/components/Modal/Modal.stories.tsx
```diff
@@ -0,0 +1,213 @@
...
+- Use the forceAction prop on the modal component if the user should be forced to take an action before closing the modal.
+
+You can also use the provided ModalToggleButton and/or ModalOpenLink components, which will adhere to the above guidelines for convenience.
+`,
+      },
+    },
+  },
+}
+
+export const defaultModal = (): React.ReactElement => {
```
brandonlenz, line 47: I was looking at the Happo diff and I'm curious if you think it would be valuable to have a story for a modal that is in open state by default. That way Happo is able to detect changes to the modal content rather than just the toggle button, which is all we can currently see.

brandonlenz (approved):
Couple of small things, but looks great and works nicely from my testing!

src/components/Modal/utils.ts
suzubara: oh yes 🤣

src/components/Modal/Modal.stories.tsx
suzubara: oh yes good call! I can add stories just for the ModalWindow component. I'll do that!

PR #1587, opened 2021-09-23
ci: Don't request all codeowners review on dependabot PRs

brandonlenz:
# Summary

Prevents PRs that include changes to ONLY the `package.json` and/or `yarn.lock` files from adding all CODEOWNERS as reviewers. This PR introduces some granularity in who is added to what PRs.

There were two approaches here:
1. At first (see commit 83861963d7eee33b4afdee08a11981055bcb44aa) I thought it would be convenient to add a separate ownership rule for `package.json` and `yarn.lock` files specifically, so that dependabot PRs (or PRs touching only one or both of those files) would only be assigned to those users.
    - Pros:
      - Simple, and gets the job done. 
    - Cons:
      - No ability to specify *only* dependabot PRs.
2. Still exclude `package.json` and `yarn.lock` from default CODEOWNERS, but use `dependabot.yml` to configure *specifically* who is added to dependabot PRs 2998738c4fc58c53b6cb58ca9174d936066b5872:
    - Pros:
      - More specifically targets *only* dependabot PRs.
      - Uses dependabot config as designed.
    - Cons:
      - Cannot specifically target non-dependabot PRs touching only `package.json` and/or `yarn.lock`. If you are a Codeowner for those files, you will be added to dependabot and non-dependabot PRs alike. If you are not a codeowner for those files, but are configured as a dependabot reviewer, you will not be auto-assigned to review non-dependabot PRs that only update dependencies without other code changes.

I have chosen approach number **2** for now.

## Related Issues or PRs

Closes #769

## How To Test

According to [CODEOWNERS docs](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners), the PR base branch CODEOWNERS file is the one that github uses to assign reviewers. The best way to test will be to

## References

- [CODEOWNERS docs](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners)
- [Other nice explanation of syntax and how the file is parsed](https://tech.people-doc.com/using-github-codeowners-file.html)
- [Dependabot `reviewers` docs](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#reviewers)

PR #1653, opened 2021-10-05
Add dependabot reviewers for example app deps

brandonlenz:
# Summary

Previous PR missed the example app dependabot configuration (I didn't scroll down to see it).

I wanted to use a yaml anchor to be more DRY, but for whatever reason, [dependabot does not support anchors and aliases in yaml config](https://github.com/dependabot/dependabot-core/issues/2127)

## Related Issues or PRs

Previous PR: https://github.com/trussworks/react-uswds/pull/1587

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -51,6 +51,11 @@ updates:
     pull-request-branch-name:
       separator: '-'
     open-pull-requests-limit: 10
+    reviewers:
+      - "suzubara"
+      - "haworku"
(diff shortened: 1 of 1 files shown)


brandonlenz:
Since merging the [previous PR](https://github.com/trussworks/react-uswds/pull/1587), dependbot PRs have only the desired maintainers on them, however I'm noticing codeowners is no longer automatically adding everyone as I would expect. Investigating, as I can probably fix that here.

PR #1731, opened 2021-11-01
ci: build artifact using CI

### LINKED ISSUE #1725: [ci] Build the library package in CI instead of on local machines
We've been tripped up a couple times now when publishing a new version by building the library on local dev machines. For example, not having local packages up-to-date or including artifacts that shouldn't be a part of the build. We are blocked on being able to publish a package to NPM from CI since we require 2FA, however we should still be able to run the test & build workflow to generate the `.tar.gz` artifact used in a release.

Maybe we can add a Github CI job that creates & stores that artifact, and simplify our release workflow to publish that artifact ([npm docs](https://docs.npmjs.com/cli/v6/commands/npm-publish) say you can `npm publish` "A url or file path to a gzipped tar archive containing a single folder with a package.json file inside.")

brandonlenz:
# Summary

Adds CI to build the tgz artifact so that local developer machines are not used, which can cause sporadic problems if things are not _pristine_ (see #1725)

## Related Issues or PRs

Closes #1725 

## How To Test

Since this action runs when a PR titled `release-*` is merged, it's certainly hard to test at this stage. I allowed the workflow to run on this PR by commenting that logic, to verify the meat of the functionality actually works. Workflow link here: https://github.com/trussworks/react-uswds/actions/runs/1412820829

Sadly, [there is no way to get upload-artifact to not zip the already zipped `.tgz`](https://github.com/actions/upload-artifact#zipped-artifact-downloads), so we have this annoying manual step to unzip it 🤷 

### Screenshots

Workflow page:
image

Download `Artifact.zip` and unzip to get the `.tgz`:
image


.github/workflows/publish-npm.yml
```diff
@@ -0,0 +1,37 @@
...
+      - closed
+    branches: 
+      - main
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    if: |
+      startsWith(github.head_ref, 'release-'')
+      && github.event.pull_request.merged
```
brandonlenz, line 15: So AFAIK there's no concise "on:" syntax for "on merge of a pr to main from a branch like feature-*", but the above should be what we need. 

'startsWith` is an [expression](https://docs.github.com/en/actions/learn-github-actions/expressions) available in Actions.

'github.head_ref' is a property available in the [github context](https://docs.github.com/en/actions/learn-github-actions/contexts)

This ends up looking somewhat complex, but that's effectively what's going on here.

.github/workflows/publish-npm.yml
```diff
@@ -0,0 +1,37 @@
...
+      
+      # The yarn cache is not node_modules
+      - name: Install Dependencies
+        run: yarn --prefer-offline
+      
+      - name: Run tests
+        run: yarn test
+      
+      - name: Publish to npm
+        uses: JS-DevTools/npm-publish@v1
```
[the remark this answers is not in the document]
brandonlenz: We can't use this without implementing a strict key rotation policy. Updated docs to capture why, so that infra doesn't get the same question again in the future (hopefully)

.github/workflows/publish-npm.yml
```diff
@@ -0,0 +1,37 @@
...
+        with:
+          token: ${{ secrets.NPM_AUTOMATION_TOKEN }}
```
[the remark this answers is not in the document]
brandonlenz: https://github.com/trussworks/react-uswds/pull/1731#discussion_r741274532

.github/workflows/package-release.yml
```diff
@@ -0,0 +1,42 @@
...
+      
+      # The yarn cache is not node_modules
+      - name: Install Dependencies
+        run: yarn --prefer-offline
+      
+      - name: Run tests
+        run: yarn test
+
+      - name: Package library
+        run: yarn pack
```
brandonlenz, line 35: After @suzubara's demo today, I have questions. Do I need to explicitly run `yarn build` instead of just install for this?

`yarn install` and `yarn build` (which just runs `webpack --progress`) both create the `lib` directory, and I'm having trouble understanding and finding info about what the difference between the two is besides the --progress flag.

.github/workflows/package-release.yml
```diff
@@ -0,0 +1,42 @@
...
+      && github.event.pull_request.merged
+      
+    steps: 
+      - name: Checkout out repository
+        uses: actions/checkout@v2
+
+      - name: Set up Node
+        uses: actions/setup-node@v2
+        with:
+          node-version: '12.x'
```
jim, line 24: Should we be using node `16.x` here? It's LTS and [12 stopped getting active support in 2020](https://endoflife.date/nodejs).

jim (approved):
This looks good to me. I checked that the artifact was built correctly, but I don't have NPM credentials and I didn't attempt to push the package.

We can maybe wrap a script around downloading the asset and running `npm publish` as a possible future optimization if we need to.

.github/workflows/package-release.yml
jim: Just found https://github.com/trussworks/react-uswds/pull/1743 😄

brandonlenz: I added https://github.com/trussworks/react-uswds/issues/1749 so that we don't have to worry about merging one PR before the other and choosing which one to update the node version of our CI jobs in.

That way we can merge #1743 and this PR at any time, while also making sure to update the node version that this CI and the rest of our existing workflows use

PR #1780, opened 2021-11-18
Combobox fires custom onChange events when option is selected

### LINKED ISSUE #1730: [fix] ComboBox doesn't fire change events on input

ReactUSWDS: 2.4.1
USWDS: 2.11.2 (or all)

**Describe the bug**

Changing the value of the ComboBox component does not fire the `onChange` event for the text input or the select element. If you look at the [USWDS source code](https://github.com/uswds/uswds/blob/v2.11.2/src/js/components/combo-box.js#L43) (look for the `changeElementValue` function) these events are fired when:
- mount if there is a selected option
- when an option is selected
- when the input is cleared or reset

**To Reproduce**

1. Add an input `onChange` handler prop to an instance of the ComboBox
2. This handler will be called when typing into the text input, but _not_ when an option is selected

**Expected behavior**

I would expect onChange events to be called consistently with the USWDS implementation

Igarfinkle:
# Summary

If a custom onChange prop is passed to the input or the select in ComboBox, that function will be called whenever a ComboBox option is selected. 

## Related Issues or PRs

Closes #1730

### DIFF
--- a/src/components/forms/ComboBox/ComboBox.stories.tsx
+++ b/src/components/forms/ComboBox/ComboBox.stories.tsx
@@ -40,6 +40,8 @@ export const defaultComboBoxWithPropOptions = (): React.ReactElement => {
         name="input-ComboBox"
         options={fruitList}
         onChange={noop}
+        inputProps={{ onChange: (e) => console.log('input') }}
+        selectProps={{ onChange: (e) => console.log('select') }}
       />
     </Form>
   )
--- a/src/components/forms/ComboBox/ComboBox.tsx
+++ b/src/components/forms/ComboBox/ComboBox.tsx
@@ -59,13 +59,14 @@ interface ComboBoxProps {
 
 interface InputProps {
   focused: boolean
+  inputRef: React.RefObject<HTMLInputElement>
 }
 
 const Input = ({
   focused,
+  inputRef,
   ...inputProps
 }: InputProps & JSX.IntrinsicElements['input']): React.ReactElement => {
-  const inputRef = useRef<HTMLInputElement>(null)
   useEffect(() => {
     if (focused && inputRef.current) {
       inputRef.current.focus()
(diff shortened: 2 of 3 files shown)


src/components/forms/ComboBox/ComboBox.tsx
```diff
@@ -399,6 +423,7 @@ export const ComboBox = forwardRef(
           onKeyDown={handleInputKeyDown}
           value={state.inputValue}
           focused={state.focusMode === FocusMode.Input}
+          inputRef={inputRef}
```
brandonlenz, line 426: Should this be?:

```suggestion
          ref={inputRef}
```

brandonlenz:
I should note here that while debugging some discrepancies with Isaac yesterday, I was trying to better understand the default behavior or the USDWS combo box. 

By opening the [USWDS component example](https://designsystem.digital.gov/components/combo-box/), opening the console, and running the following:

```javascript
const input = document.querySelector("#fruit");
input.addEventListener('change', (event) => {
    console.log("input changed...")
});
```

I was able to listen for `change` events on the `input` to see when it actually fired. As the github issue states, change events are only fired if the input selection, or reset happens:

https://user-images.githubusercontent.com/15805554/142665768-e0dde24e-fad6-4194-b5ef-4f7126f550ae.mov

Posting this on the ticket as a reference, mostly, as it took me a bit to confirm what the behavior should be.

src/components/forms/ComboBox/ComboBox.stories.tsx
```diff
@@ -40,6 +40,8 @@ export const defaultComboBoxWithPropOptions = (): React.ReactElement => {
         name="input-ComboBox"
         options={fruitList}
         onChange={noop}
+        inputProps={{ onChange: (e) => console.log('input') }}
```
brandonlenz, line 43: @suzubara am I understanding correctly that we want to be able to pass an "onChange" to the input like this, and we need to not only make sure it triggers on select/reset/mount, but also override the default behavior of triggering on when the user types within the input?

I don't see where USWDS overrides the "typing" change events. Does a JSX/React created `input` element have a different default "onChange" behavior?

Igarfinkle:
I think I'm looking for a little more concrete guidance on the intended functionality here, if you have any.

rpdelaney:
I'm closing this as stale. Please re-open if you disagree!

PR #1865, opened 2022-01-07
feat: accept prop for heading level (#1751)

### LINKED ISSUE #1751: [fix] StepIndicator should handle custom heading levels to allow semantic html
```
@trussworks/react-uswds 2.4.1
uswds 2.11.2
```

**Describe the bug**

**To Reproduce**

1. Add` <StepIndicator>`
2. Inspect HTML 
3. See that heading always uses `<h4>` tags and there are no  props on the component to adjust this. This is an accessibility issue.

**Expected behavior**

Consumer can adjust the heading level to allow semantic HTML. his could be a `type` prop on the component similar to the proposed `IconList` interface.

FYI this type of functionality is also mentioned in the [USWDS docs for Step indicator](https://designsystem.digital.gov/components/step-indicator/)

> Use semantic heading levels. Though our default code uses an `<h4>`, use the correct heading level in your own implementation.

rswerve:
# Summary

The heading level in the StepIndicator currently defaults to `h4`, and the inability to change this causes an accessibility issue when used in a document that doesn't include `h1`, `h2`, and `h3`.  We should be able to specify the desired heading level.

This PR introduces a `headingLevel` prop to the StepIndicator, so that any given instance can have any valid heading level. 

In order to make this a non-breaking change, if no heading level is specified, the component maintains the current behavior of defaulting to `h4`.

## Related Issues or PRs

Closes #1751 

## How To Test

In Storybook, select the Heading variant of the StepIndicator, and inspect the markup to see that it accords with the screenshot below.

Review new unit tests.

### Screenshots (optional)

image

### DIFF
--- a/src/components/stepindicator/StepIndicator/StepIndicator.stories.tsx
+++ b/src/components/stepindicator/StepIndicator/StepIndicator.stories.tsx
@@ -65,7 +65,7 @@ export const smallCounters = (): React.ReactElement => (
   </StepIndicator>
 )
 
-export const heading = (): React.ReactElement => (
+export const customHeadingLevel = (): React.ReactElement => (
   <StepIndicator headingLevel="h2">
     <StepIndicatorStep label="Personal information" status="complete" />
     <StepIndicatorStep label="Household status" status="complete" />
--- a/src/components/stepindicator/StepIndicator/StepIndicator.stories.tsx
+++ b/src/components/stepindicator/StepIndicator/StepIndicator.stories.tsx
@@ -64,3 +64,13 @@ export const smallCounters = (): React.ReactElement => (
     <StepIndicatorStep label="Review and submit" />
   </StepIndicator>
 )
+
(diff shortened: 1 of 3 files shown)


src/components/stepindicator/StepIndicator/StepIndicator.stories.tsx
```diff
@@ -64,3 +64,13 @@ export const smallCounters = (): React.ReactElement => (
     <StepIndicatorStep label="Review and submit" />
   </StepIndicator>
 )
+
+export const heading = (): React.ReactElement => (
```
haworku, line 68:
```suggestion
export const customHeadingLevel = (): React.ReactElement => (
```

Naming the story after what its demonstrating.

haworku (dismissed):
LGTM

suzubara (approved):

haworku (approved):

PR #1878, opened 2022-01-11
fix: Fix Tooltip positioning

### LINKED ISSUE #1877: [fix] Tooltip top/bottom positioning is broken

ReactUSWDS 2.6.0
USWDS 2.11.2

**Describe the bug**

The Tooltip stories (https://trussworks.github.io/react-uswds/?path=/story/components-tooltip--tooltip-top) with top/bottom positioning is off.

**To Reproduce**

Visit the Tooltip stories for top/bottom and display the tooltip.
- Top: https://trussworks.github.io/react-uswds/?path=/story/components-tooltip--tooltip-top
- Bottom: https://trussworks.github.io/react-uswds/?path=/story/components-tooltip--tooltip-bottom

**Expected behavior**

The top/bottom tooltips are positioned correctly relative to the toggle element. See examples: https://designsystem.digital.gov/components/tooltip/

**Screenshots**

image

image

Mac OS 12, Chrome

suzubara:
# Summary

Updates the code used to position the Tooltip component to reflect [USWDS 2.11.2 Tooltip source code](https://github.com/uswds/uswds/blob/v2.11.2/src/js/components/tooltip.js). This seems to resolve the positioning bug.

## Related Issues or PRs

Fixes #1877 

## How To Test

View the Tooltip component in Storybook at various screen sizes and verify the tooltip is always visible and positioned as expected.

### Screenshots (optional)

image

image

### DIFF
--- a/src/styles/index.scss
+++ b/src/styles/index.scss
@@ -4,6 +4,7 @@
 @import 'type';
 
 // This should be set by USWDS but isn't
+// It can be removed when https://github.com/uswds/uswds/issues/4458 is fixed
 .usa-tooltip__body {
   opacity: 0;
 }
--- a/src/styles/index.scss
+++ b/src/styles/index.scss
@@ -2,3 +2,8 @@
 
 @import 'backgrounds';
 @import 'type';
+
+// This should be set by USWDS but isn't
+.usa-tooltip__body {
+  opacity: 0;
+}
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,19 +21,6 @@ jobs:
             - ~/.cache/yarn
           key: v1-yarn-{{ checksum "yarn.lock" }}
 
-      - run: yarn test:coverage -w 1
-
-      - run: yarn test:serverside
-
-      - run:
-          name: 'Happo'
-          command: 'yarn happo-ci'
-          environment:
-            HAPPO_IS_ASYNC: 'true'
-            BASE_BRANCH: 'origin/main'
-
-      - run: yarn lint
-      - run: yarn danger ci --failOnErrors
       - run:
           name: Create GitHub Deployment
           command: ./scripts/deployment_start.sh > deployment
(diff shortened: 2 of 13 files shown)


brandonlenz (approved):

suzubara:
While trying this branch out on my project, I realized that there were more issues with positioning - the code to loop through the various positions wasn’t working as expected. I think it was because the code to set the new CSS and check if the element was visible or not were all happening in the same `useEffect` hook, but to accurately detect whether or not the element is visible it needs to wait for the re-render. So I did some more refactoring and pushed up an update that I think solves this once and for all!

The new way this works is: when the tooltip gets shown, it sets the position to whatever was intended. in a second `useEffect` hook that specifically looks for changes in the position, it checks whether the tooltip is visible. If it is, it turns on the `is-visible` class and that’s that. if it’s not, it starts the loop to go through each position (and also try wrapping text if needed).

It is possible that it won’t find a position that works… this might happen if the tooltip contents is very long and/or the window is SO small that there’s no room. I checked and this can come up in the USWDS implementation as well. I put a `console.warn` in this case to surface the issue

Finally… while working on this I also discovered what I think is a USWDS bug so I filed https://github.com/uswds/uswds/issues/4458
but to fix in the meantime, I just added that line of CSS to our lib.. hopefully it can be removed when the bug is fixed.

src/styles/index.scss
```diff
@@ -2,3 +2,9 @@
 
 @import 'backgrounds';
 @import 'type';
+
+// This should be set by USWDS but isn't
+// It can be removed when https://github.com/uswds/uswds/issues/4458 is fixed
+.usa-tooltip__body {
+  opacity: 0;
+}
```
brandonlenz, line 10: Nice!

brandonlenz:
Something I just noticed (not for us to fix here): Looks like the tooltip takes the component's margin into account causing top not to be centered on the element, for example: 
Screen Shot 2022-01-12 at 4 11 26 PM


But its the same way in the USWDS example, so we're good!!

brandonlenz (approved):

PR #1890, opened 2022-01-25
fix: Render modal outside app root by default.

### LINKED ISSUE #1886: [fix] Modal doesn't work with keyboard using JAWS

Latest version of ReactUSWDS. Does not occur with USWDS.

**Describe the bug**

When a user is using Chrome and JAWS, the content within modals (including buttons) can't be interacted with using the keyboard.

**To Reproduce**

1. Start Chrome and JAWS
2. Go to [the modal page in Storybook](https://trussworks.github.io/react-uswds/?path=/story/components-modal--default-modal)
3. Open the modal
4. Attempt to close the modal using the buttons within it

**Expected behavior**

The buttons should work just as they do when JAWS isn't being used.

**Additional context**

I think the issue is that we're adding `aria-hidden` to parent elements of the modal, and that this removes the modal's content from the accessibility tree. The original USWDS JS assumes that modals exist outside the normal flow of the page, so this technique isn't an issue there.

jim:
# Summary

## Related Issues or PRs

This fixes #1886 by using a Portal to render modals outside the application root by default.

The root cause is that when we mark elements in the page with `aria-hidden` as a part of opening a modal, it prevents the modal and its contents from appearing in the accessibility tree since they appear as children of elements that are now hidden.

This means that users with JAWS can’t interact with the modal using the keyboard. Without JAWS running, the keyboard works properly. This is also how we missed it for so long

I was able to reproduce the issue both in Managed Care’s app and react-uswds storybook.

The original USWDS modal’s JS assumes that the modal will appear outside the rest of the app, usually at the top level of the document body. This PR uses a Portal to accomplish something similar, and that is also [the approach used by react-modal](https://github.com/reactjs/react-modal/blob/master/src/components/Modal.js#L21-L24).

## How To Test

Visit the modal in storybook. Verify that you can interact with it and its contents using JAWS.

### DIFF
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -30,6 +30,12 @@ export type ModalRef = {
   toggleModal: (event?: React.MouseEvent, open?: boolean) => boolean
 }
 
+// Modals are rendered into the document body default. If an element exists with the id
+// `modal-root`, that element will be used as the parent instead.
+// 
+// If you wish to override this behavior, `renderToPortal` to `false` and the modal
+// will render in its normal location in the document. Note that this may cause the modal to
+// be inaccessible due to no longer being in the document's accessbility tree.
 export const Modal = forwardRef(
   (
     {
@@ -188,11 +194,13 @@ export const Modal = forwardRef(
           </ModalWindow>
         </ModalWrapper>
       </FocusTrap>
-   
+  
     if (renderToPortal) {
+      const modalRoot = document.getElementById("modal-root")
+      const target = modalRoot || document.body
       return ReactDOM.createPortal(
         modal,
-        document.body
+        target
       )
     } else {
       return modal
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -10,6 +10,7 @@ import FocusTrap from 'focus-trap-react'
 import { useModal, getScrollbarWidth } from './utils'
 import { ModalWindow } from './ModalWindow/ModalWindow'
 import { ModalWrapper } from './ModalWrapper/ModalWrapper'
+import ReactDOM from 'react-dom'
 
 interface ModalComponentProps {
   id: string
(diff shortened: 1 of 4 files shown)


src/components/Modal/Modal.tsx
```diff
@@ -162,7 +165,7 @@ export const Modal = forwardRef(
       },
     }
 
-    return (
+    const modal = 
       <FocusTrap active={isOpen} focusTrapOptions={focusTrapOptions}>
         <ModalWrapper
           role="dialog"
```
haworku, line 171: Unit tests are looking for role dialog. Do we need to move that attribute onto something wrapping the `modal` used for the portal as well? 

If we do, I like the pattern in [here ](https://assortment.io/posts/accessible-modal-component-react-portals-part-2)where its a default but can be overridden. This could be a good chance to surface props for that role attribute which might change depending on use case.  In the new code, seems like the role goes on the modal wrapper when `renderToPortal` is false;  otherwise the role goes on `modal` itself ... _I think_

jim: I think the issue was that because the `Modal` was being directly rendered into `document.body`, it wasn't within the `div` that React Testing Library was looking inside of when `screen.getByRole` was called. I fixed this in the next commit and modified the behavior to use a div with an `id` of `modal-root` to render modals if it exists.

jim:
I am realizing that something similar to this could be accomplished by careful use of the `modalRoot` property by providing a selector that locates a top-level `div` defined outside the app root. I think it's going to be better to default to something that will always work, and then let folks opt out of the portal usage if they need to for some reason.

src/components/Modal/Modal.tsx
haworku: ah makes sense! Thanks for explaining

haworku (approved):

haworku:
My only concern here is should this be a breaking change. Any thoughts? @suzubara @jim just pinging you all here to weigh in.

haworku:
We decided this seems fine and the `fix` minor release works.

lpsinger:
This broke our application because we are using server-side rendering, and `document` is not defined by Node.

haworku:
@lpsinger Thanks for comment here - could you please file an issue?

lpsinger:
> @lpsinger Thanks for comment here - could you please file an issue?

See 1892.

PR #1895, opened 2022-02-03
feat: Add ability to pass through heading props

brandonlenz:
# Summary

Adds the ability to pass through props to the heading element.
Also restructures classNames internally so that we don't destructively assign classNames to elements if their passthrough props include a `className`.

This is important for being able to set accessible identifiers on the various elements (for instance using `aria-labelledby` and referencing the `id` of the heading element)

## How To Test

Tested by locally pulling in dependency into my Truss project.
Also added unit tests.

No user facing changes. Backwards compatible, unless users were clobbering the default classNames with the various passthrough props we made available, but that feels more like fixing a bug than anything to me.


src/components/stepindicator/StepIndicator/StepIndicator.tsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react'
 import classnames from 'classnames'
 import { StepIndicatorStepProps } from '../StepIndicatorStep/StepIndicatorStep'
+import classNames from 'classnames'
```
Igarfinkle, line 4: duplicate import?

Igarfinkle (approved):

src/components/stepindicator/StepIndicator/StepIndicator.tsx
brandonlenz: Good catch [3921f25](https://github.com/trussworks/react-uswds/pull/1895/commits/3921f25f76143152c68cba23fed7f913f5a97753)

haworku (approved):

PR #1899, opened 2022-02-11
fix: Fix modal causing body padding to increase when unmounted

### LINKED ISSUE #1898: [fix] Modal element adds padding to the body element on page navigation

```
"@trussworks/react-uswds": "2.7.1",
"uswds": "2.11.2"
```

**Describe the bug**

In the Modal component, the unmount effect toggles the `body.style.paddingRight` value regardless of whether the modal was open or not when unmounted. On my project, the unmount happens on each page navigation, resulting in the body padding right increasing by 15px each time. This effect should only toggle the padding right if the modal was open.

**Expected behavior**

The body padding right should only be altered when the modal is opened or closed.

**Screenshots**

image

suzubara:
# Summary

I fixed this issue by consolidating some of the code that handles DOM changes when the modal is opened & closed, and explicitly resetting body padding when its closed instead of toggling between the two values.

I also added to the docs about how to use `yarn link` to test the library locally with a NextJS app.

## Related Issues or PRs

Fixes #1898 

## How To Test

* You can test this on an application that uses the Modal component by causing the modal to mount & unmount (typically by navigating to a new page). Without this fix, the padding-right on the body element would continue to increase by 15px (or the width of your scrollbar)
* Tested & verified on [Space Force](https://github.com/USSF-ORBIT/ussf-portal-client)
* I also added a test case for this that was failing without the fix.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -128,10 +128,44 @@ You can then run `yarn build:watch` in ReactUSWDS and your application's compile
 
 > **Warning:** Make sure to **unlink** the package once you're done making changes! It can be easy to forget you're using a linked package in your application, and commit changes without switching back to a released version of ReactUSWDS.
 
+> If you get errors about invalid hook calls, you may also need to link the `react` & `react-dom` packages:
+>
+> ```
+> ➜ cd /path/to/react-uswds
+> ➜ yarn link
+> ➜ yarn install
+>
+> ➜ cd node_modules/react
+> ➜ yarn link
+>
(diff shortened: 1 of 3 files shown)


brandonlenz (approved):
Nice! I did some smoke testing in deployed storybook, but I trust your local testing with your project code (plus the test coverage here for the issue is great!)

Also glad to see what looks like prettier catching some formatting quirks that snuck through 🥳 

The `handleOpen/CloseEffect` make things nice and clear too 🎉

PR #1905, opened 2022-02-17
feat: Accept Accordion headingLevel in items prop 

### LINKED ISSUE #1904: [feat] Allow Accordion component to accept different heading levels

[Accordion](https://github.com/trussworks/react-uswds/blob/main/src/components/Accordion/Accordion.tsx)

The heading level is fixed at `h4`.  In order to meet accessibility guidelines, we can't skip heading levels, so we need the flexibility to choose the heading level that makes sense in the context of where it is used.

I'd like to add a prop `headlingLevel` for Accordion, with a default (h4? not sure what would be the best default) that would allow the options of h2, h3, h4, h5, h6

**Additional context**

I am using this currently on my project, and the heading level of these accordions will need to be an h2 or h3.

kimallen:
# Summary

This allows the Accordion component to accept any heading level.  Previously, it was fixed at `h4`.  For backwards compatibility, the default is `h4`.
 
## Related Issues or PRs

closes #1904 

## How To Test

To test, one can try running storybook locally and adding a `headingLevel` prop to an item in `items` Accordion prop of anything other than the default `h4`.

### DIFF
--- a/src/components/Accordion/Accordion.test.tsx
+++ b/src/components/Accordion/Accordion.test.tsx
@@ -379,4 +379,119 @@ describe('Accordion component', () => {
       )
     })
   })
+  describe('custom headingLevel for AccordionItems', () => {
+    it('passes on the headingLevel', () => {
+      const customTestItems = [
+        {
+          title: 'First Amendment',
+          content: (
+            <p>
+              Congress shall make no law respecting an establishment of
+              religion, or prohibiting the free exercise thereof; or abridging
+              the freedom of speech, or of the press; or the right of the people
+              peaceably to assemble, and to petition the Government for a
+              redress of grievances.
+            </p>
+          ),
+          expanded: false,
+          headingLevel: 'h2',
+          id: '123',
+          className: 'myCustomAccordionItem',
+        },
+        {
+          title: 'Second Amendment',
+          content: (
+            <>
+              <p>
+                A well regulated Militia, being necessary to the security of a
+                free State, the right of the people to keep and bear Arms, shall
+                not be infringed.
+              </p>{' '}
+              <ul>
+                <li>This is a list item</li>
+                <li>Another list item</li>
+              </ul>
+            </>
+          ),
+          expanded: false,
+          headingLevel: 'h2',
+
+          id: 'abc',
+        },
+        {
+          title: 'Third Amendment',
+          content: (
+            <p>
+              No Soldier shall, in time of peace be quartered in any house,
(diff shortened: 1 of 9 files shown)


src/components/Accordion/Accordion.test.tsx
```diff
@@ -379,4 +379,119 @@ describe('Accordion component', () => {
       )
     })
   })
+  describe('custom headingLevel for AccordionItems', () => {
+    it('passes on the headingLevel', () => {
+      const customTestItems: AccordionItemProps[] = [
```
brandonlenz, line 384: For other potential reviewers:

AccordionItemProps needed to be renamed (from AccordianItem) so that it could be exported without conflict and used here so that the type inference on `headingLevel` would match instead of just being `string` which caused type errors

src/components/Accordion/Accordion.test.tsx
```diff
@@ -379,4 +379,119 @@ describe('Accordion component', () => {
...
+      ]
+
+      const { getAllByRole } = render(<Accordion items={customTestItems} />)
+
+      const headings = getAllByRole('heading', { level: 2 })
+      expect(headings.length).toEqual(customTestItems.length)
+    })
+
+    it('uses headingLevel h4 by default', () => {
+      const testItems = [
```
brandonlenz, line 443: I think here you can just use the `testItems` defined at the top of the file which the first few tests use. That should help with what you were saying about repeating this data a lot

brandonlenz: I would also be 100% okay with adding the expects around heading level to the default `renders without errors` test as well, but that's not blocking for me. There's pros and cons to going each way 🤷

brandonlenz (approved):
LGTM, though I would definitely recommend at least using the existing `testItems` in the default h4 test

brandonlenz:
Looking at Happo diffs, it looks like they're all IE, on all components. Something about the margin or screen size:
image

Happo usually issues a warning when something like that happens, but i don't see one. Still I'm inclined to approve that since it appears to be an IE issue or some update with happo runners on IE, not related to these changes.

src/components/Accordion/Accordion.test.tsx
kimallen: I went ahead and removed the test-specific test item.  Left it in its own test since it seemed to make sense next to the non-default test.

brandonlenz (approved):

PR #1928, opened 2022-03-23
feat: Adds headingLevel prop for Alert component

### LINKED ISSUE #1917: [feat] Allow flexible heading for Alert component

[Alert](https://github.com/trussworks/react-uswds/blob/main/src/components/Alert/Alert.tsx)

The heading in this component is fixed at `h4`.  In order to make sure heading levels are not skipped, particularly for accessibility, one needs to be able to choose which heading level to use.

Create a `headingLevel` prop for Alert to allow for h1-h6

Here is an [example](https://github.com/trussworks/react-uswds/pull/1905/files#diff-a31ce76dd91c59f1f5ac8d8adbe78ee1ca5b3e195f546b8419f6f80554f6d626) of a solution to a similar problem in the Accordion component.

**Additional context**

kimallen:
# Summary

Adds an optional `headingLevel` prop with options h1-h6.  Default is `h3` for backwards compatibility.

## Related Issues or PRs

closes #1917 

## How To Test

pass in a prop `headingLevel` with a heading that is not the default h3

### Screenshots (optional)

Storybook example:
image


src/components/Alert/Alert.tsx
```diff
@@ -39,10 +41,11 @@ export const Alert = ({
     className
   )
 
+  const Heading = headingLevel
   return (
     <div className={classes} data-testid="alert" {...props}>
       <div className="usa-alert__body">
-        {heading && <h4 className="usa-alert__heading">{heading}</h4>}
+        {heading && <Heading className="usa-alert__heading">{heading}</Heading>}
```
kylehilltruss, line 48: The original heading level for this component was an `h4`. I'd be concerned about changing the default heading level to be an `h3` -- it might well break any existing uses of this component if they aren't updated to include the new prop.

kylehilltruss (changes requested):
Everything else looks good, but the change in default heading level inside this component might constitute a breaking change.

src/components/Alert/Alert.tsx
kimallen: oops, I thought it was an h3- that must have been the SummaryBox I was looking at.  Will change that.

brandonlenz (approved):
Looks like you've addressed Kyle's feedback. Everything looks good now!

kylehilltruss (approved):

PR #1929, opened 2022-03-24
feat!: Make Summary Box more flexible

### LINKED ISSUE #1916: [!feat] Make summary box components more flexible

[Summary Box](https://designsystem.digital.gov/components/summary-box/)

Right now it is not possible to compose a summary box flexibly, IE, if someone wanted to place a logo above the header within the summary box. 
Additionally, this should include in scope the ability to specify the heading level on a summary box (currently uses the default of `h3`, but like many components containing headers, the heading level is important to be customizable)

Build SummaryBoxHeading and SummaryBoxText components to allow consumers to compose a summary box themselves.

> ⚠️ Note: This is likely a breaking change as it would alter the existing summary box api

It would be possible to let the current implementation live in parallel with a more flexible implementation, but I don't think we want to be maintaining multiple versions of the same components or extensive conditional rendering logic within a component for the sake of backwards compatibility alone

**Additional context**

Lets say I wanted to do something like this:
image

Igarfinkle:
> ### ⚠️ Set the following as the commit body when merging this PR: `BREAKING CHANGE: SummaryBox now exposes sub-components (SummaryBoxHeading and SummaryBoxContent) for a more compositional API. Consumers will need to update their implementation to match.`

# Summary

Makes the SummaryBox component more flexible, separating out SummaryBoxHeading and SummaryBoxHeading into their own components.
Is NOT backwards compatible. 

## Related Issues or PRs

Closes #1916 

## How To Test

Try the SummaryBox Storybook story to ensure it still displays properly


src/components/SummaryBox/SummaryBoxHeading/SummaryBoxHeading.tsx
```diff
@@ -0,0 +1,24 @@
...
+interface SummaryBoxHeadingProps {
+  heading: string
+  className?: string
+  headingLevel?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6'
+}
+
+export const SummaryBoxHeading = ({
+  heading,
+  className,
+  headingLevel,
```
kimallen, line 13: You could give `headingLevel` a default here rather than handle that below with `headingLevel = h3`.  That might be more explicit for someone reading this code.

src/components/SummaryBox/SummaryBox/SummaryBox.test.tsx
```diff
@@ -47,34 +47,32 @@ const testSummaryBoxContent = (
...
 describe('SummaryBox component', () => {
   it('renders without errors', () => {
-    const { getByRole } = render(
-      <SummaryBox heading="Key information">{testSummaryBoxContent}</SummaryBox>
+    const { getByTestId } = render(
+      <SummaryBox>{testSummaryBoxContent}</SummaryBox>
     )
 
-    expect(getByRole('heading')).toBeInTheDocument()
+    expect(getByTestId('summary-box')).toBeInTheDocument()
```
brandonlenz, line 58: This test _was_ (although the name didn't reflect it explicitly) the only test that validated that a heading as being rendered. Perhaps `testSummaryBoxContent` should be updated to include the heading now?

src/components/SummaryBox/SummaryBoxHeading/SummaryBoxHeading.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
+import classnames from 'classnames'
+
+interface SummaryBoxHeadingProps {
+  heading: string
```
brandonlenz, line 5: instead of a heading prop, can we use `children` here? That way the api can look like `<SummaryBoxHeading>My Heading Content</SummaryBoxHeading>`

That'll also allow more flexibility over just `string` in case someone wanted to otherwise stylize or pass in custom tags.

src/components/SummaryBox/SummaryBoxHeading/SummaryBoxHeading.tsx
```diff
@@ -0,0 +1,24 @@
...
+  className?: string
+  headingLevel?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6'
+}
+
+export const SummaryBoxHeading = ({
+  heading,
+  className,
+  headingLevel = 'h3',
```
brandonlenz, line 13: @suzubara @haworku @kylehilltruss What are your thoughts on having a default value here? In the past we've done it as a way to preserve backwards compatibility (non-breaking changes) where we've introduced the ability to customize the heading level. 

In this case, this is a breaking change anyways since the SummaryBox API is changing to be more compositional in nature. Should we just rip off the bandaid and make it a mandatory prop? IMO I think we should, since there's no real significance to `h3` other than it is what the USWDS example happens to use.

kimallen:
I'm wondering if it would be worth breaking out the `headingLevel` change into its own issue/PR from the breaking change so that the `headingLevel` flexibility can be incorporated as a minor non-breaking change before the breaking change.

brandonlenz (commented):
@Igarfinkle thanks so much for picking this up!

A couple of suggestions and one question to some other maintainers in my review here. 

One nit: We don't seem to have a unified pattern for how we name folders when double nesting things, (e.g. in this case `SummaryBox/SummaryBox`, or `Footer/Footer` or `card/Card`). JS, React, and TS are pretty unopinionated about this, so none are wrong.

Personally, if a folder does not directly contain a component file, I think it should be lower-cased (ie. `summarybox/SummaryBox`) 

@kylehilltruss @suzubara @haworku thoughts? Some people prefer snakeCase or kebab-case for multi-word directories. If we come to a consensus, I'll draft a PR + ADR to set them all straight 😉

src/components/SummaryBox/SummaryBoxHeading/SummaryBoxHeading.tsx
brandonlenz: @kimallen also makes a good point [here](https://github.com/trussworks/react-uswds/pull/1929#issuecomment-1081003074) about multi-staging this process.

src/components/SummaryBox/SummaryBox/SummaryBox.test.tsx
Igarfinkle: I don't think that's needed: now the heading is just... part of the children of SummaryBox.

src/components/SummaryBox/SummaryBoxHeading/SummaryBoxHeading.tsx
Igarfinkle: As long as we're not taking the USWDS example as the source of truth, then yes, I think we should remove the default value

src/components/SummaryBox/SummaryBox/SummaryBox.test.tsx
brandonlenz: Right, but shouldn't the tests somewhat reflect how we expect the component to be used? In which case, the `render` should include those tags with `testSummaryBoxContent` rendered inside the `SummaryBoxContent`.

And then I think there's still value in checking for the header to be rendered within. What we could probably do here is combine the second test (renders passed in children) with the first test, make sure we do a check for the header, and call it a day

src/components/SummaryBox/SummaryBox/SummaryBox.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+    expect(getByTestId('summary-box')).toBeInTheDocument()
+  })
+
+  it('renders passed in children', () => {
+    const { getAllByRole } = render(
+      <SummaryBox>{testSummaryBoxContent}</SummaryBox>
+    )
+
+    expect(getAllByRole('listitem')).toHaveLength(4)
```
brandonlenz, line 73: I see you've updated `testSummaryBoxContent` to include the header and content tags. Can we put the check for the header here to assuage my concerns?

src/components/SummaryBox/SummaryBoxHeading/SummaryBoxHeading.tsx
haworku: Yes to removing default value-  this is a breaking change we should put things in the ideal state we would like it to be in going forward.

src/components/SummaryBox/SummaryBoxHeading/SummaryBoxHeading.tsx
```diff
@@ -0,0 +1,24 @@
...
+  children,
+  className,
+  headingLevel,
+  ...h3Props
+}: SummaryBoxHeadingProps &
+  JSX.IntrinsicElements['h3']): React.ReactElement => {
+  const classes = classnames('usa-summary-box__heading', className)
+  const Heading = headingLevel
+  return (
+    <Heading className={classes} {...h3Props} data-testid="summaryBoxHeading">
```
brandonlenz, line 20: nit: I prefer to save the data-testid for when something really can't easily be queried from the dom by conventional means.

Before this PR the test used the following: `expect(getByRole('heading')).toBeInTheDocument()`. I think it could also work with `getByText`, knowing what the heading's content is.

Igarfinkle: ah, I forgot about getByRole!

src/components/SummaryBox/SummaryBoxHeading/SummaryBoxHeading.test.tsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react'
+import { getByRole, render } from '@testing-library/react'
```
brandonlenz, line 2: Let's get rid of this unused import

src/components/SummaryBox/SummaryBox/SummaryBox.test.tsx
```diff
@@ -0,0 +1,87 @@
+/* eslint-disable jsx-a11y/anchor-is-valid */
+import React from 'react'
+import { getByRole, render } from '@testing-library/react'
```
brandonlenz, line 3: This one also

brandonlenz (approved):
Approving but with a couple of tiny things to fix listed below.

We'll likely want to time the release of this (since it's breaking), so lets work that out in our slack

src/components/SummaryBox/SummaryBoxHeading/SummaryBoxHeading.tsx
kimallen: @Igarfinkle Did you decide to use `heading` prop anyway instead of `children`? I liked @brandonlenz 's suggestion

kimallen: either way, a comment here to explain why it was resolved without changes would be helpful to future viewers.

src/components/SummaryBox/SummaryBox/SummaryBox.test.tsx
Igarfinkle: Done. I'm surprised eslint doesn't find this problematic enough to fail the CI check.

haworku: Good point. I think that unused direction might be a warning not error? Perhaps we need to adjust to have CI checks fail on warnings too - I think its just adding `--max-warnings=0` to our eslint call.

haworku (approved):
LGTM. I made https://github.com/trussworks/react-uswds/issues/2000 to address the issue brandon brought up.

brandonlenz (approved):
Approving now since I think we've done all the communicating around an intentionally breaking release

brandonlenz:
@Igarfinkle when merging this, please make sure that the commit body of the merge commit contains some verbiage like 

```
BREAKING CHANGE: SummaryBox now exposes sub-components (SummaryBoxHeading and SummaryBoxContent) for a more compositional API. Consumers will need to update their implementation to match.
```

Issue #1946, opened 2022-03-31
Deprecate headingLevel default values and make the prop mandatory


brandonlenz:
**Which USWDS component(s) does this relate to?**

All components taking a `headingLevel` prop with a default value (typically with `= 'h3'` in the props argument to a functional component).

Since we have been recently moving to making all components with headers more flexible, it makes sense to me to require that consumers specify which headingLevel they would like to use, since the defaults are just what the USWDS examples happen to use.

I recommend doing this in 2 stages:

1. Add a deprecation warning when headingLevel is _not_ passed into components. We could probably achieve this by removing the default value from the props extraction, and checking if it is undefined within the functional component body to issue a deprecation warning, and _then_ set the default value.
2. Add a breaking change that makes the prop mandatory, and remove the deprecation warning.

PR #1947, opened 2022-03-31
feat: Deprecate default heading levels

Igarfinkle:
# Summary

- Deprecates default headingLevels on a few components. 
- A future PR will mandate the specification of a heading level.
## Related Issues or PRs

towards #1946 

## How To Test

In Storybook, change a component to not specify a heading level. See that a deprecation warning appears in the console.


src/components/Accordion/Accordion.test.tsx
```diff
@@ -446,4 +464,62 @@ describe('Accordion component', () => {
...
+        {
+          title: 'First Amendment',
+          content: (
+            <p>
+              Congress shall make no law respecting an establishment of
+              religion, or prohibiting the free exercise thereof; or abridging
+              the freedom of speech, or of the press; or the right of the people
+              peaceably to assemble, and to petition the Government for a
+              redress of grievances.
+            </p>
```
brandonlenz, line 480: Maybe worth defining as a `const firstAmendmentContent` or something (same with the other accordion items), just to keep the test a little more brief. Not a huge deal since DAMP tests are good, but I think that'd be more than acceptable to factor out.

src/components/Accordion/Accordion.tsx
```diff
@@ -33,7 +34,13 @@ export const AccordionItem = ({
     'usa-prose',
     className
   )
-  const Heading = headingLevel
+  let Heading = headingLevel
```
brandonlenz, line 37: Similar comment. I would try to take the approach of keeping this const

src/components/Alert/Alert.tsx
```diff
@@ -41,7 +42,14 @@ export const Alert = ({
...
   )
 
-  const Heading = headingLevel
+  let Heading = headingLevel
+  if (!Heading) {
+    deprecationWarning(
+      'Default headingLevel h4 has been deprecated. If you are including a heading, specify a headingLevel.'
+    )
+    Heading = 'h4'
+  }
```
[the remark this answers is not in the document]
Igarfinkle: good idea

src/components/Accordion/Accordion.test.tsx
Igarfinkle: smart

kimallen (commented):
Do you want to also include a deprecation warning for `SummaryBox`?

All the changes LGTM.

haworku (approved):

brandonlenz:
I think @kimallen's suggestion of adding the prop for SummaryBox is good, if you're up for that, then adjusting the follow-up breaking PR (as I'm sure it will introduce conflicts).

Also good with just releasing if it feels like overhead, since I do expect us to make the breaking release shortly after the one we're planning for today

PR #1967, opened 2022-04-06
fix(docs): Fix typo in Storybook documentation template #1914 (#1949)

brandonlenz:
# Summary

Internal PR for CI checks

## Related Issues or PRs

Merge to main of OSS contirbution: https://github.com/trussworks/react-uswds/pull/1949

PR #1949, opened 2022-04-03
fix(docs): Fix typo in Storybook documentation template #1914

fitiskin:
# Summary

## Related Issues or PRs

closes https://github.com/trussworks/react-uswds/issues/1914

Issue #1235, opened 2021-05-10
[feat] New Component: Input prefix or suffix


SirenaBorracha:
# [Input prefix or suffix](https://designsystem.digital.gov/components/input-prefix-suffix/)

## USWDS Version

[2.11.0](https://github.com/uswds/uswds/releases/tag/v2.11.0)

## Description

Use input prefixes and suffixes to show symbols or abbreviations that help users enter the right type of information in a form’s text input.

## Requirements (proposed)

**Prop Interface**

```JSX
interface InputPrefixSuffixProps = {
 type: ‘prefix’ | ‘suffix’
 className?: string
 children: React.ReactNode
} & JSX.IntrinsicElements['div']
```

## Example Implementation (proposed)

```JSX
return (
  <FormGroup>
    <label class="usa-label" for="example-input-prefix">Credit card number</label>
    <InputPrefixSuffix type="prefix">
      <Icon name="credit_card"/>
    </InputPrefixSuffix>
  </FormGroup>

  <FormGroup error={true}>
    <label class="usa-label" for="example-input-prefix-error">Credit card number (Error)</label>
    <InputPrefixSuffix type="prefix">
      <Icon name="credit_card"/>
    </InputPrefixSuffix>
  </FormGroup>

   <FormGroup>
     <label class="usa-label" for="example-input-suffix">Weight, in pounds</label>
    <InputPrefixSuffix type="suffix">
      lbs.
    </InputPrefixSuffix> 
  </FormGroup>
)
```

**More Details**

- `<div>`root element

## Other Considerations

-

jenbutongit:
👋 Hey there, I'm an engineer currently consulting in uk gov, are you still looking for contributors for this? Happy to look into it 🤓

@SirenaBorracha - I think for your proposed prop interface you meant to use `type` instead of `interface`. Using interface  with `& JSX.xx` will give you this namespace error TS2708: Cannot use namespace 'JSX' as a value.

I'd also suggest two components here, but happy to concede (since we don't have this component ourselves so I'm not sure how your developers would rather use this component), I'd have one component for `InputPrefix` and one for `InputSuffix`, and remove the need for `type: ‘prefix’ | ‘suffix’`

Is InputPrefix|Suffix only applicable to textInputs? Could I suggest an API similar to this? https://github.com/NHSDigital/nhsuk-react-components/tree/main/src/components/error-summary

so in this case the implementation would be 
```tsx
 <FormGroup>
[...]

PR #1911, opened 2022-02-25
docs(Storybook): 404 page template added 

svalexander:
# Summary
This pr adds a 404 page not found example to the ReactUSWDS Storybook like the one seen in USWDS' documentation. As many of the elements of the template already existed in the documentation template and the authentication templates I borrowed chunks of code from those examples and updated with the appropriate changes. 

## Related Issues or PRs

https://github.com/trussworks/react-uswds/issues/1619

## How To Test

* Run "yarn storybook" to start Storybook locally. 
* Can verify addition of template by navigating to "page templates > not found page". 
* Can then compare template addition with the USWDS 404 example found [here](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/develop/components/preview/layout--404.html).

### Screenshots (optional)
Screen Shot 2022-02-25 at 5 18 05 PM
Screen Shot 2022-02-25 at 5 18 15 PM

PR #1969, opened 2022-04-07
docs: Callout outside contributors process

rogeruiz:
# Summary

Outside contributors can submit PRs against but can't have our status checks run on their forked branches due to security concerns. We mitigate these limitations by having documentation around outside contributions. This patch series helps by adding documentation and messaging to outside contributors so that both maintainers and contributors know what next steps are. We can continue fleshing this out going forward as well.

## Related Issues or PRs

- https://github.com/trussworks/react-uswds/pull/1967
- https://github.com/trussworks/react-uswds/pull/1949#issuecomment-1090316632
- https://github.com/trussworks/react-uswds/issues/1235#issuecomment-941145975


Igarfinkle:
I'm looking at [this PR](https://github.com/trussworks/react-uswds/pull/1911) where it looks like a forked branch _is_ having its checks run. Is that supposed to be happening?

rogeruiz:
> I'm looking at [this PR](https://github.com/trussworks/react-uswds/pull/1911) where it looks like a forked branch _is_ having its checks run. Is that supposed to be happening?

@Igarfinkle 
While checks are run, not all checks are run from outside contributors. See my screenshot below for another reference. But the build and danger checks are not run when there are changes from outside contributors.

image

brandonlenz (approved):

brandonlenz (commented):
Not sure if this PR is still draft or not, but the verbiage looks good to me.

suzubara (approved):

PR #1970, opened 2022-04-07
ci: Improve Happo Consistency

brandonlenz:
# Summary

Hopefully resolves some spurious diffs we see on almost every PR with tooltips and occasionally with default value DatePicker.

## How To Test

First test will be making sure that happo completes correctly on this branch/PR. It is somewhat unclear how beforeScreenshot and waitFor play together if used simultaneously.
There may also still be happo diffs on this PR (I expect it in fact). I don't think the issue will be resolved until the base branch _and_ PR branches both have this code. 

## References
https://docs.happo.io/docs/storybook#waiting-for-content


Igarfinkle (approved):
Let's merge this! I'm curious to see if it works, and it only affects stories so it's very low risk

PR #1980, opened 2022-04-12
ci: Fix flaky Happo tests

Igarfinkle:
# Summary

Attempts to fix remaining flaky happo tests, especially in Tooltip.

### DIFF
--- a/src/components/Tooltip/Tooltip.stories.tsx
+++ b/src/components/Tooltip/Tooltip.stories.tsx
@@ -119,7 +119,10 @@ export const tooltipTopLeftWrap = (): React.ReactElement => (
 )
 tooltipTopLeftWrap.parameters = {
   happo: {
-    delay: 1000,
+    waitFor: () =>
+      document.querySelector(
+        '.usa-tooltip__body.usa-tooltip__body--right.usa-tooltip__body--wrap.is-visible.is-set'
+      ),
   },
 }
 
@@ -132,7 +135,10 @@ export const tooltipBottomLeftWrap = (): React.ReactElement => (
 )
 tooltipBottomLeftWrap.parameters = {
   happo: {
-    delay: 1000,
+    waitFor: () =>
+      document.querySelector(
+        '.usa-tooltip__body.usa-tooltip__body--right.usa-tooltip__body--wrap.is-visible.is-set'
+      ),
   },
 }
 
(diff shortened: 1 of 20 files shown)


src/components/forms/DatePicker/DatePicker.stories.tsx
```diff
@@ -92,7 +92,7 @@ export const withDefaultInvalidValue = (): React.ReactElement => (
 )
 withDefaultValue.parameters = {
   happo: {
-    waitForCOntent: '1988-05-16',
+    waitForContent: '1988-05-16',
```
brandonlenz, line 95: Oh dang. Nice catch!

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -126,6 +133,14 @@ export const tooltipBottomLeftWrap = (): React.ReactElement => (
     </Tooltip>
   </div>
 )
+tooltipBottomLeftWrap.parameters = {
+  happo: {
+    waitFor: () =>
+      document.querySelector(
+        '.usa-tooltip__body.usa-tooltip__body--right.usa-tooltip__body--wrap.is-visible.is-set'
```
brandonlenz, line 140: Depending on the screen size that happo is taking a screenshot for, the tooltip position may use a different default position, which would mean different different classes than what you expect to find. That probably limits us to doing something like:

```suggestion
        '.usa-tooltip__body.usa-tooltip__body--wrap.is-visible.is-set'
```

Or more generically using waitFor at the top level with `'.usa-tooltip__body.is-visible.is-set'`.

A lot of the flaky diffs _are_ on the ios-safari screenshots, which usually forces the alternate position of the tooltip.

brandonlenz (changes requested):

src/components/Tooltip/Tooltip.stories.tsx
Igarfinkle: Fixed!

brandonlenz (approved):

brandonlenz:
I've permanently ignored the ios-safari diffs for tooltip. Happo seems to crop the screenshots differently every time, despite now waiting for the tooltip to be visible before taking them. We should still merge this as it *should* in theory make these diffs more stable

PR #1983, opened 2022-04-14
ci: exclude ios-safari from happo on Tooltip

Igarfinkle:
# Summary

Exclude Happo from checking Tooltip on ios-safari, which has frequently caused problems with ci checks


brandonlenz (approved):
These can be ignored in the Happo interface (which I started doing because, yea, ios-safari tooltip diffs have been a nightmare for too long), but I think controlling via the code provides more easily accessible developer documentation, so I'm good with this 👍

PR #1987, opened 2022-04-15
ci: change expected date format for DatePicker in Happo

Igarfinkle:
# Summary

Expects to find `MM/DD/YYYY` in Happo screenshots instead of `YYYY-MM-DD`

Even though the component returns the selected date in `yyyy-mm-dd` format, it displays as `mm/dd/yyyy`, so I believe we should use that format in the `waitforcontent` preceding the happo check.

### Screenshots (optional)
image


brandonlenz (approved):
Great catch. Turns out my entire PR trying to help Happo was wrong 🤦

PR #1988, opened 2022-04-15
refactor!: Remove 1.x.x and 2.x.x deprecated properties

brandonlenz:
# Summary

## Related Issues or PRs

closes: https://github.com/trussworks/react-uswds/issues/1946

## How To Test

The following components that used to have deprecated props do not issue any deprecation warnings:
- Accordion
- Alert
- Button
- CollectionHeading
- Footer/Address
- Footer/Footer
- Footer/FooterNav
- Footer/Logo
- Search
- Fieldset
- TextInput
- header/NavList
- StepIndicator

The storybook components that previously relied on default heading levels render:
- Accordion
- Alert
- CollectionHeading
- StepIndicator


src/components/forms/TextInput/TextInput.tsx
```diff
@@ -18,23 +17,7 @@ interface RequiredTextInputProps {
...
   validationStatus?: 'error' | 'success'
-  /**
-   * @deprecated since 1.6.0, use validationStatus
-   */
-  error?: boolean
-  /**
-   * @deprecated since 1.6.0, use validationStatus
-   */
-  success?: boolean
   inputSize?: 'small' | 'medium'
```
brandonlenz, line 20: We actually don't use this prop at all in storybook, and USWDS doesn't show it in their examples, though it comes from the following:

image

Wondering if we should create a `type` that enumerates those valid sizes, or just leave it up to consumers to pass in the desired `className`.

If the latter is chosen, I can remove the prop here and call it out as part of the breaking release.

Thoughts?

brandonlenz: I _have_ been meaning to see what it would look like to create a nice interface for `2xs`...`2xl` passed in as a `size` prop on a couple of components, which would just append `size` to the `className` like so:

```TSX
`usa-input--${size}`: size
```

I think its a nice convenience to offer, and probably doesn't involve much overhead on our part besides combing through the code. So that's what the alternative might look like.

brandonlenz:
> Jest: "global" coverage threshold for branches (87%) not met: 86.99%

😅

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -16,7 +16,7 @@ export default {
         })
         document.querySelector('.usa-tooltip__trigger').dispatchEvent(event)
       },
-      waitFor: () =>
+      waitFor: <E extends Element>(): E =>
```
brandonlenz, line 19: This was just an eslint warning

src/components/forms/TextInput/TextInput.stories.tsx
```diff
@@ -65,16 +65,6 @@ export const readonly = (): React.ReactElement => (
...
 
-/*
-export const small = (): React.ReactElement => (
-  <TextInput id="input-type-text" name="input-type-text" type="text" inputSize="small" />
-)
-
-export const medium = (): React.ReactElement => (
-  <TextInput id="input-type-text" name="input-type-text" type="text" inputSize="medium" />
-)
-*/
```
brandonlenz, line 76: Deletion of these commented stories corresponds to https://github.com/trussworks/react-uswds/pull/1988#discussion_r851441877

src/components/forms/TextInput/TextInput.tsx
haworku: Oh, I like that. I think my only 🤔 was if that would make maintenance harder because we would be using dynamically calculated class names instead of having to list out all the potential options ourselves in a `classnames` object and having it readable in the code.  But thinking about it more, that's probably what storybook is for - we can show all the variants of things there.

src/components/Footer/Logo/Logo.test.tsx
```diff
@@ -29,28 +26,4 @@ describe('Logo component', () => {
     const { getByText } = render(<Logo image={logoImage} heading={heading} />)
     expect(getByText('Swoosh Branding')).toBeInTheDocument()
   })
```
haworku, line 28: we should still keep some kind of test of the `size` prop influencing classes for Logo

tsconfig.json
```diff
@@ -20,8 +20,6 @@
   "include": ["src/**/*", "custom.d.ts"],
   "exclude": [
     "node_modules",
-    "**/*.test.ts",
-    "**/*.test.tsx",
```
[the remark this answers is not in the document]
haworku: My haphazard guess is _maybe_ we included this to not have type issues in tests just break typescript compile for the rest of the lib?

haworku (approved):

src/components/Footer/Logo/Logo.test.tsx
brandonlenz: Oh nice catch, yea this test file doesn't have tests for just size by itself, so instead of deleting that code I'll repurpose it to be similar to what we have for other components

brandonlenz: [dffb9a5](https://github.com/trussworks/react-uswds/pull/1988/commits/dffb9a5c255796173d5c79e6d8e0ef88e1956357)

tsconfig.json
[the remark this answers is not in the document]
brandonlenz: Yea that sounds right. I could see it being an annoying if you're just trying to write some tests quickly and focus on the typing where it really matters. Given that it works when I've removed these exclusions, it seems like we're in pretty good shape. I definitely don't want to harm anyone's developer experience though

Igarfinkle (dismissed):

Igarfinkle (approved):

Igarfinkle (approved):

PR #1998, opened 2022-04-19
fix: Happo updates

trotzig:
Hi there! Henric from Happo here. 

@Igarfinkle reached out to me to inform me that your Happo jobs were timing out. I've had a look at some of the recent failures and have found them to be caused by unstable and slow execution times in Internet Explorer. With IE going away soon (June 15th is the official date from Microsoft), we've found it increasingly hard to keep IE up and perform well. Recent Windows updates push Edge pretty hard, and it looks like they've even made IE deliberately slower (although I have no concrete numbers to back this up...). 

In this PR, I'm taking care of a few things: 
- Upgrade to latest happo.io library
- Re-enable Edge (should be more stable at this point)
- Split up IE jobs in three chunks

I initially came here with the intention to remove IE from `.happo.js`, but I decided to try a workaround first to see if we can keep it up for just a short while longer. Eventually however, you are going to have to remove it. We've set a tentative shutdown date to June 1st, but we'll work with the few customers still on IE to adjust that date if necessary.

### DIFF
--- a/.happo.js
+++ b/.happo.js
@@ -18,6 +18,7 @@ module.exports = {
     }),
     'internet explorer': new RemoteBrowserTarget('internet explorer', {
       viewport: '1024x768',
+      chunks: 3,
     }),
     safari: new RemoteBrowserTarget('safari', {
       viewport: '1024x768',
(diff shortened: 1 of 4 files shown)


trotzig:
I'm not 100% sure about this, but I think the IE diffs we are seeing in this PR could be caused by a previously rendered story affecting the width of the body. Now that we have three separate IE runs, some stories are no longer co-rendered in the same session. The "after" version has the right image dimensions (1024 wide) so I'd say this is a positive diff.

brandonlenz:
@trotzig Thank you so much for taking the time to diagnose and make a PR on this repository!

PR #2002, opened 2022-04-19
chore(tests): Use await on all userEvents

brandonlenz:
# Summary

Noticed some Combobox tests were throwing warnings that there were overlapping act events. Likely related to `userEvent` upgrade since `userEvent`s are wrapped in `act`, and they're all supposed to be async/awaited in the newest version. Found a few that were not, so I updated those here and the warnings have been silenced :)

## How To Test

`yarn test` and verify no output for Combox relate to overlapping `act` events exists


haworku (approved):
🏅

Issue #2053, opened 2022-04-25
[fix] react-uswds no longer ships a .d.ts file, cannot be used with TypeScript any more


lpsinger:
**ReactUSWDS Version & USWDS Version:**

`@trussworks/react-uswds@3.0.0`, `uswds@2.13.1`

**Describe the bug**

react-uswds 3.0.0 no longer ships a .d.ts file, so it cannot be used in strict TypeScript projects.

**To Reproduce**

See https://github.com/tachgsfc/gcn.nasa.gov/runs/6163692157?check_suite_focus=true. Error message:

```
Error: app/components/CopyableCode.tsx(9,41): error TS7016: Could not find a declaration file for module '@trussworks/react-uswds'. '/home/runner/work/gcn.nasa.gov/gcn.nasa.gov/node_modules/@trussworks/react-uswds/lib/index.js' implicitly has an 'any' type.
[5](https://github.com/tachgsfc/gcn.nasa.gov/runs/6163692157?check_suite_focus=true#step:6:5)
  Try `npm i --save-dev @types/trussworks__react-uswds` if it exists or add a new declaration (.d.ts) file containing `declare module '@trussworks/react-uswds';`
[6](https://github.com/tachgsfc/gcn.nasa.gov/runs/6163692157?check_suite_focus=true#step:6:6)
````

**Expected behavior**

react-uswds should ship an index.d.ts file.

**Screenshots**

**Additional context**

**Device and Browser Information (please complete the following information if describing a UI bug):**

Joshua-Flores:
Experiencing the same issue

brandonlenz:
I'm looking in to this at the moment. Will update as a fixed version becomes available.

@lpsinger For now I would recommend staying on the 2.8.0 release.

brandonlenz:
This will be resolved in the next release. Likely 3.0.2

PR #2076, opened 2022-05-09
ci: Fix build not including typescript types

brandonlenz:
# Summary

Looks like when the TS version was bumped, the refactor to remove `awesome-typescript-loader` and rely purely on babel didn't backfilll for type generation which babel does not do on its own. In this PR we manually call `tsc` in the build step and let it emit the type declarations. This is the recommended approach for projects using babel alone and not relying on ts-loader.

Alternative considered: use ts loader on ts|x files before running babel. ts-loader is a bit slow, and felt like a more repetitive solution, using two loaders on all of our ts/tsx files when `tsc` will directly create just the type exports for us without any additional config.

Resources: 
- https://www.typescriptlang.org/docs/handbook/babel-with-typescript.html

## Related Issues or PRs

closes: https://github.com/trussworks/react-uswds/issues/2053
stems from: https://github.com/trussworks/react-uswds/pull/1548

## How To Test

Checkout the code, clean your local files (delete node_modules and lib), freshly install packages with yarn. running `yarn` will also run `yarn build` which will create the lib directory where you can confirm that .d.ts files are now output correctly again. Running `yarn pack` will generate the tarball which you can also verify includes the types.

```
> git checkout bl-fix-types
> rm -rf node_modules
> rm -rf lib
> yarn
> yarn pack
```

### Screenshots (optional):
image


src/index.ts
```diff
@@ -93,11 +98,12 @@ export { Logo } from './components/Footer/Logo/Logo'
 export { SocialLinks } from './components/Footer/SocialLinks/SocialLinks'
 
 /** Modal components */
-export { Modal, ModalProps, ModalRef } from './components/Modal/Modal'
+export { Modal } from './components/Modal/Modal'
 export { ModalToggleButton } from './components/Modal/ModalToggleButton'
 export { ModalOpenLink } from './components/Modal/ModalOpenLink'
 export { ModalHeading } from './components/Modal/ModalHeading/ModalHeading'
 export { ModalFooter } from './components/Modal/ModalFooter/ModalFooter'
+export type { ModalProps, ModalRef } from './components/Modal/Modal'
```
brandonlenz, line 106: These `export type` changes silence some TS warnings we were getting on build

tsconfig.json
```diff
@@ -11,11 +11,12 @@
     "module": "esnext",
     "moduleResolution": "node",
     "resolveJsonModule": true,
-    "noEmit": true,
+    "emitDeclarationOnly": true,
```
brandonlenz, line 14: Previously, `awesome-typescript-loader` was exporting the types. Now we rely on tsc to do it, so we emit those type declarations only, and let babel emit the rest of the transpiled source code

tsconfig.json
```diff
@@ -11,11 +11,12 @@
...
     "jsx": "react",
     "noImplicitAny": true,
     "declaration": true,
-    "outDir": "./lib"
+    "outDir": "./lib",
+    "isolatedModules": true,
```
brandonlenz, line 19: Recommended setting from https://www.typescriptlang.org/docs/handbook/babel-with-typescript.html

https://www.typescriptlang.org/tsconfig#isolatedModules

package.json
```diff
@@ -27,9 +27,9 @@
     "test:serverside": "yarn build && node src/serverSideTest.js",
     "storybook": "start-storybook -p 9009",
     "storybook:deploy": "storybook-to-ghpages",
-    "build": "webpack --progress",
-    "build:watch": "webpack --watch",
-    "lint": "tsc --noEmit && eslint --ext js,jsx,ts,tsx src && stylelint \"src/**/*.{css,scss}\"",
+    "build": "tsc && webpack --progress",
+    "build:watch": "tsc && webpack --watch",
```
brandonlenz, line 31: Manually call `tsc` here so that types are built to `lib`

package.json
```diff
@@ -27,9 +27,9 @@
...
+    "lint": "tsc --noEmit --emitDeclarationOnly false && eslint --ext js,jsx,ts,tsx src && stylelint \"src/**/*.{css,scss}\"",
```
brandonlenz, line 32: `noEmit` cannot be used alongside `emitDeclarationOnly` (as of ts 3.8), which is now in tsconfig. So we have to set it to false here, which is the recommended workaround

Igarfinkle (approved):
This tentatively looks good to me... how can we check that this fixes the problem?

brandonlenz:
> This tentatively looks good to me... how can we check that this fixes the problem?

The `How to test` steps involve running the commands that our release CI runs to package the library. Following those steps results in properly packaged type definitions.

PR #2097, opened 2022-05-17
Update README.md for 3.0

wpears:
Now that 3.x has been released, the README should match.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@
 
 **ReactUSWDS Component Library**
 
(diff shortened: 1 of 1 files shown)


haworku:
Hey! Thanks for the PR. We actually don't yet support USWDS 3.0, so we don't want to make this change yet.
.
FYI the current level of uwsds we support is dictated by the `package.json` - you can always see the most recent supported version in the badges at the top of our README. Each time a new uswds release goes out we look through the release notes and code changes and determine if any of our components need to change before we bump up.

 Let me know if you have any questions here.

PR #2113, opened 2022-05-20
ci: Support Node@18

Igarfinkle:
# Summary

Node@12 has reached end of life and is no longer maintained. Node@18 was just released as the current version of Node. This PR replaces Node@12 with Node@18 in our build tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "lib"
   ],
   "engines": {
-    "node": "10.x - 16.x"
+    "node": "10.x - 18.x"
   },
   "scripts": {
     "test": "jest",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -17,7 +17,7 @@
     "lib"
   ],
   "engines": {
-    "node": "10.x - 16.x"
+    "node": "10.x - 18.x"
```
brandonlenz, line 20: This is where I'm not sure. Should this be 14.x-18.x if we only test for those versions?

I think we've previously said that that is a *breaking* release for us. so perhaps that should be reserved for a second, breaking PR.

brandonlenz: https://github.com/trussworks/react-uswds/issues/2129

PR #2133, opened 2022-05-27
fix: export HeadingLevel type

lpsinger:
# Summary

## Related Issues or PRs

Fixes #2115.

PR #2134, opened 2022-05-27
fix: Emit headingLevel type and do not emit type declarations for tests

brandonlenz:
# Summary

Accomplishes https://github.com/trussworks/react-uswds/pull/2133 while also making sure that tsconfig checks compilation of test and story files, without emitting them on build (so they will not be included in `lib`). Previously, empty *.test.d.ts files were being emitted when I removed tests from the `tsconfig.json` `exclude`.

Here We use a standard `tsconfig.json` for development which _does_ look at tests and stories for typescript compilation (big win for us because we were not doing this previously, so our storybook examples had typescript errors in them). 

We also use a new `tsconfig.build.json` which extends the development `tsconfig.json` to emit types (which we need to do), as well as excludes tests and stories, since we do not need to be concerned with either of those types of files for release packaging.

- [ ] `@reviewers` There's an issue I could use some help with debugging. There's a _massive_ .png file (larger than all of the code exported combined) being emitted into `lib` on build. Any idea where it is coming from and how to stop that from happening?:
image 

- [ ] Also, do we need all those woff and ttf font files in there?

## How to test

1. clear your local environment of transient files, and rebuild the project (`yarn` or `yarn install` runs `yarn build` after the install)
   ```
   rm --rf lib
   rm --rf node_modules
   yarn
   ```
1. check the `lib` folder that gets generated to verify that no test or story type files are present and that `types/headingLevel.d.ts` is included

## Review notes:

You might note a lot of props `interface`s changing to `type`s. This is to fix the following TS error that shows up in storybook files: https://stackoverflow.com/questions/64703203/typescript-problem-with-storybook-migration-to-csf

## Related Issues or PRs

replaces:  https://github.com/trussworks/react-uswds/pull/2133

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,9 +27,9 @@
     "test:serverside": "yarn build && node src/serverSideTest.js",
     "storybook": "start-storybook -p 9009",
     "storybook:deploy": "storybook-to-ghpages",
-    "build": "tsc && webpack --progress",
+    "build": "tsc -p tsconfig.build.json && webpack --progress",
     "build:watch": "tsc && webpack --watch",
-    "lint": "tsc --noEmit --emitDeclarationOnly false && eslint --ext js,jsx,ts,tsx src && stylelint \"src/**/*.{css,scss}\"",
+    "lint": "tsc && eslint --ext js,jsx,ts,tsx src && stylelint \"src/**/*.{css,scss}\"",
     "release": "standard-version -t ''",
     "prepare": "yarn build",
     "prepublishOnly": "yarn test && yarn lint",
@@ -127,7 +127,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "tsc --noEmit --emitDeclarationOnly false && lint-staged",
+      "pre-commit": "tsc && lint-staged",
       "pre-push": "yarn danger local -b main --failOnErrors"
     }
   },
--- a/src/components/Accordion/Accordion.test.tsx
+++ b/src/components/Accordion/Accordion.test.tsx
@@ -2,6 +2,7 @@ import React from 'react'
 import { render, fireEvent } from '@testing-library/react'
 
 import { Accordion, AccordionItemProps } from './Accordion'
+import { HeadingLevel } from '../../types/headingLevel'
 
 const firstAmendment = (
   <p>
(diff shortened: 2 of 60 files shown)


package.json
```diff
@@ -27,9 +27,9 @@
     "test:serverside": "yarn build && node src/serverSideTest.js",
     "storybook": "start-storybook -p 9009",
     "storybook:deploy": "storybook-to-ghpages",
-    "build": "tsc && webpack --progress",
```
brandonlenz, line 30: for building, use the build tsconfig

package.json
```diff
@@ -27,9 +27,9 @@
...
+    "build": "tsc -p tsconfig.build.json && webpack --progress",
     "build:watch": "tsc && webpack --watch",
-    "lint": "tsc --noEmit --emitDeclarationOnly false && eslint --ext js,jsx,ts,tsx src && stylelint \"src/**/*.{css,scss}\"",
```
brandonlenz, line 32: return to how this used to be, since the base tsconfig that is used by default has `noEmit: true` already set

package.json
```diff
@@ -127,7 +127,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "tsc --noEmit --emitDeclarationOnly false && lint-staged",
```
brandonlenz, line 130: return to how this used to be, since the base tsconfig that is used by default has noEmit: true already set

lpsinger:
I confirm that this fixes #2133. Thank you! Can't wait for this to go into a release so that I can update to 3.x!

src/index.ts
```diff
@@ -134,3 +134,6 @@ export { ProcessListItem } from './components/ProcessList/ProcessListItem/Proces
 export { ProcessListHeading } from './components/ProcessList/ProcessListHeading/ProcessListHeading'
 
 export { SiteAlert } from './components/SiteAlert/SiteAlert'
+
+/** Types */
+export type { HeadingLevel } from './types/headingLevel'
```
brandonlenz, line 139: Rather than forcing prop structs like `items` on `Accordion` to use `const items: ComponentProps<typeof Accordion>['items'] = {...}`, which is a bit gross, this export was added to allow the `headingLevel` prop to be declared as `headingLevel: "h4" as HeadingLevel` if consumers want to go that route.

PR #2167, opened 2022-06-14
fix: Danger contributors check

rogeruiz:
# Summary

This patch series adds better checking for Danger verifying if contributors was modified by only checking the JSON Diff. If there's not contributor object, then there is no change and move on. This eliminates all the false positives we've seen in recent PRs around modifying the package.json file near the word contributors.

## How To Test

You can create a new branch locally and modify the package.json with a contributors stanza. Then commit it and run the following Danger command to trigger it.

```
yarn danger local -b main --failOnErrors
```

### DIFF
--- a/dangerfile.ts
+++ b/dangerfile.ts
@@ -1,6 +1,7 @@
 import * as child from 'child_process'
 
-import { danger, fail, warn } from 'danger'
+import { danger, fail, schedule, warn } from 'danger'
+
 const shouldRun = !danger.github || (danger.github && danger.github.pr.user.type !== 'Bot');
 
 // Load all modified and new files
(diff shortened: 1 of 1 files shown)


dangerfile.ts
```diff
@@ -126,43 +128,28 @@ const checkDependencyChanges: () => void = () => {
...
-      })
+// Check for any changes to the contributors section of package.json
+schedule(async () => {
+  if (!shouldRun) {
+    return;
   }
-}
+  const pd = await danger.git.JSONDiffForFile('package.json')
+
+  if (pd.contributors) {
```
rogeruiz, line 138: If contributors exists in the diff, then we fail. This is a better check than seeing if contributors exists in the text diff.

dangerfile.ts
```diff
@@ -1,6 +1,8 @@
 import * as child from 'child_process'
 
-import { danger, fail, warn } from 'danger'
+import { danger, fail, schedule, warn } from 'danger'
+
+const shouldRun = !danger.github || (danger.github && danger.github.pr.user.type !== 'Bot');
```
rogeruiz, line 5: Add this up here as we're checking if we should run asynchronously now as well as functionally.

dangerfile.ts
```diff
@@ -43,7 +45,7 @@ const checkYarnAudit: () => void = () => {
       })
       fail(
         `${issuesFound}${summary.data.vulnerabilities.high} high vulnerabilities and ` +
-          `${summary.data.vulnerabilities.critical} critical vulnerabilities found`
+        `${summary.data.vulnerabilities.critical} critical vulnerabilities found`
```
rogeruiz, line 48: This was a formatting fix based setup.

rogeruiz (commented):
This comments should help with the review.

brandonlenz (approved):

Issue #2183, opened 2022-08-02
[feat] Make GovBanner more modular


brandonlenz:
**Does your feature request relate to a specific USWDS component?**

[Banner](https://designsystem.digital.gov/components/banner/)

**What USWDS Version is this feature present in?**

N/A

**Is your feature request related to a problem? Please describe.**

Kind of. There's been a desire to update the gov banner copy for both [react-uswds](https://github.com/trussworks/react-uswds/issues/1836) and [uswds](https://github.com/uswds/uswds/issues/4419) for some time. 

Many of our components are moving to a more modular approach where we enable the library consumer to add their own content, however the GovBanner is implemented pretty rigidly, only exposing control over the TLD and the language.

By making the component more modular, the onus of passing in content is on the consumer, therefore they may use their own internationalization implementation, and can update the copy/content to reflect the purpose of their application. While the uswds examples use specific content, uswds components can be composed, so ReactUSWDS components should offer similar compositional convenience.

closes #1836 

**Describe the solution you'd like**

Allow consumers to structure the govbanner compositionally, and pass in their own content/elements. That likely includes making smaller components (BannerHeader, BannerHeaderText, BannerHeaderAction, BannerContent, BannerGuidance, etc) as needed

Optional, but probably worth it to rename `GovBanner` to `Banner` to match the uswds naming convention, so long as there are no conflicts with other components.

Ideally, expose a way for consumers who want the 'default' text to be able to easily use it, as the component is relatively complex when the sub components are not encapsulated.

This will be a breaking change, so consider the workaround/action that will need to be taken by existing consumers.  

**Describe alternatives you've considered**

- Leaving things as-is is an option, and certainly less work, but limits the usefulness of this component considerably.

**Additional context**

Using this component at a _state_ government is an example of when it is useful to be able to update the copy/content.

Issue #1836, opened 2021-12-10
[feat] Update Gov banner copy


suzubara:
**Does your feature request relate to a specific USWDS component?**

Gov Banner

**What USWDS Version is this feature present in?**

TBD

**Is your feature request related to a problem? Please describe.**

https://github.com/uswds/uswds/issues/4419

**Describe the solution you'd like**

When new copy has been implemented in the above issue we should reflect it in ours.

PR #2184, opened 2022-08-08
feat: Add compositional Banner components

brandonlenz:
# Summary

Allow consumers to build custom Banners (say if a consumer needed a banner for a state government website)

This change is inspired by the DOL project which is not open source so I cannot link it here, but in short, we needed a custom Banner which GovBanner didn't satisfy the needs of.

I made the choice to leave the existing GovBanner in place because it's a nice convenience for consumers that want the default banner, and there's no easy way to introduce the breaking change workaround of removing it and then telling consumers to build a lengthy compositional component (see the [example](https://github.com/trussworks/react-uswds/blob/0e5368f9528f58fc0e431d1b8dc209b11ec722d5/src/components/banner/CustomBanner.stories.tsx#L18)) just to replace what they had. This way, there's no breaking changes, consumers are happy, and they have more options to boot.

## Related Issues or PRs

closes: #2183
closes: #1836 

## How To Test

1. Verify `GovBanner` has not changed
   - There is a snapshot jest test that does this for us. Note the only change is the use of the inline svg, which actually matches the USWDS banner component code _more closely_ than GovBanner previously did. 
   - Note that this means that the exported uswds lock svg isn't used. That is intentional, so that props can be passed in to the inline svg (title, description, so that consumers can pass in localized strings)
   - The lack of a Happo diff on GovBanner is also a good sign that this refactor breaks nothing
2. The new components are largely tested by GovBanner making use of them and GovBanner being tested, however each includes its own tests as well
3. Run storybook and view the `Components/Banner/Custom Banner` story

## Screenshots:

Custom banner storybook example (this one just changes some of the header text, but you can change _anything_! On the DOL project, we use the compositional structure to also allow us to internationalize _all_ of the text and aria content with i18n!:

image


src/components/banner/GovBanner/GovBanner.tsx
```diff
@@ -0,0 +1,161 @@
...
+  header: string
+  headerAction: string
+  tldSectionHeader: string
+  tldSectionContent: JSX.Element
+  httpsSectionHeader: string
+  httpsSectionContent: JSX.Element
+}
+
+const getCopy = (language: Language, tld: TLD): GovBannerCopy => {
+  const lock = <BannerLockImage title="Lock" description="A locked padlock" />
```
gidjin, line 31: thought: I'm wondering if we need to think about internationalization support differently as we add more languages. This isn't something I'm thinking for this PR or anything just starting to wonder as I review.

gidjin (approved):
LGTM! Thanks for updating this and the little refactoring for `MediaBlockBody`

src/components/banner/GovBanner/GovBanner.tsx
haworku: Strongly agreed. Added a i18n label to capture what issues have occurred so far related to this topic https://github.com/trussworks/react-uswds/labels/i18n for reference.

Issue #2129, opened 2022-05-23
[feat!]: Drop support for unsupported Node versions (10.x -12.x at time of creation)


brandonlenz:
**Is your feature request related to a problem? Please describe.**

We only test for supported versions (14.x - 18.x at the time of writing), since older versions of Node are not supported. We should explicitly drop that support under `engines` in `package.json`. This is, by our definition, a _breaking_ release (for consumers who are not on up-to-date node versions.

relevant PR comment: https://github.com/trussworks/react-uswds/pull/2113#discussion_r879840894

**Describe the solution you'd like**

Explicitly show which range of node versions are supported, and make sure those are the versions that are tested.

**Describe alternatives you've considered**

We could leave it in place, but I imagine we don't want to leave any doubt in what versions we maintain the library for.

PR #2200, opened 2022-10-11
build(node): Upgrade to nodejs 18

gidjin:
# Summary

Upgrades node from 16.x to 18.x since 16 is EOL 2023-09-11 See [nodejs release schedule](https://github.com/nodejs/release#release-schedule) and [update about the EOL of node 16.x](https://nodejs.org/en/blog/announcements/nodejs16-eol/)

Also drops support for unsupported versions of node 10.x - 12.x

## Related Issues or PRs

#2129 

## How To Test

Ensure you can install dependencies, run various commands like `yarn lint`, `yarn test`, and `yarn storybook`.


brandonlenz (approved):
@gidjin there's one more reference to node 16.13.0 in .circleci.config.yml that I think we might as well update too.

Approving to not be a blocker on it

gidjin:
> @gidjin there's one more reference to node 16.13.0 in .circleci.config.yml that I think we might as well update too.
> 
> Approving to not be a blocker on it

Thanks for catching that! Fun fact our configuration requires re-review of approved PRs that are updated.

.circleci/config.yml
```diff
@@ -5,7 +5,7 @@ jobs:
       - auth:
           password: $DOCKER_PASSWORD
           username: $DOCKER_USERNAME
-        image: circleci/node:16.13.0
+        image: circleci/node:18.10.0
```
brandonlenz, line 8: Looks like the build failed. I'm wondering if we need to use the latest? Looks like we're attempting to use a deprecated image https://circleci.com/developer/images/image/cimg/node

brandonlenz: It looks like the deprecated circleci/node:* images don't have one for this node version based on the build failure we got https://app.circleci.com/pipelines/github/trussworks/react-uswds/9313/workflows/5d04dc51-0c15-4d6c-ae43-45911c6963d3/jobs/9348/steps

brandonlenz:
```suggestion
        image: cimg/node:18.10.0
```

gidjin: Just to note, this new image is the one to use, but it has also removed `python` version 2.x so I had to update the deploy scripts to use `python3` instead.

brandonlenz (approved):

PR #2207, opened 2022-10-20
fix: Add defaultValue prop to Search component

### LINKED ISSUE #2206: [fix] Search component should support default value

**Describe the bug**

Cannot add default value to the search component.

**To Reproduce**

1. Go to Search component story 
2. Add `defaultValue` prop and value
3. View in Storybook

**Expected behavior**

See a default value appear in the search bar 

**Screenshots**

**Additional context**

1Maria:
# Summary

## Related Issues or PRs

closes #2206

### DIFF
--- a/src/components/Search/Search.test.tsx
+++ b/src/components/Search/Search.test.tsx
@@ -36,6 +36,18 @@ describe('Search component', () => {
     )
   })
 
+  it('renders a default value', () => {
+    const mockSubmit = jest.fn()
+    const defaultValue = 'Search here'
+    const { queryByTestId } = render(
+      <Search onSubmit={mockSubmit} defaultValue={defaultValue}></Search>
+    )
+    expect(queryByTestId('textInput')).toHaveAttribute(
+      'value',
+      defaultValue
+    )
+  })
+
   it('renders a label', () => {
     const mockSubmit = jest.fn()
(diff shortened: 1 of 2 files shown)


haworku:
@1Maria Can you share more about the use case you envision for the "default value" on the search component? Right reading the test it looks like what you would want to use is the `placeholder` prop to say something like "Search here" in the input before the user types something. Is that what you are looking for?

1Maria:
@haworku the use case I had in mind was for rehydrating state - for example, after a page reload. Is there a different way to accomplish that?

haworku:
@1Maria Ah, gotcha. Thanks for clarifying, I agree, we need a way to latch into the input for that. I can definitely see the use case where you would want to load a search component with a value already present (like loading a search page based on url with query params or something).

rogeruiz (approved):
lgtm 🌈

Great discussion on this @hannahkanetruss and @1Maria. Also great work on this @1Maria on getting issue #2206 resolved.

gidjin:
In accordance with our notes on [contributions](https://github.com/trussworks/react-uswds/blob/main/docs/contributing.md#%EF%B8%8F-note-to-outside-contributors) I opened a new PR #2211 that will allow the full build suite to run.

PR #2231, opened 2022-12-15
chore(deps): update various dependencies

gidjin:
# Summary

This PR fixes a typo in the `CODEOWNERS` file and upgrades various deps listed below

- Storybook releated deps
- Danger
- Happo.io
- Husky
- webpack-cli

## How To Test

Ensure build works `yarn`
Ensure tests run `yarn test`
Run storybook `yarn storybook`


CODEOWNERS
```diff
@@ -1,6 +1,6 @@
 # Default owners for all files
 
-- @gidjin @haworku @ahobson @brandonlenz @sirenaborracha @kylehilltruss @rogeruiz @jcbcapps
+* @gidjin @haworku @ahobson @brandonlenz @sirenaborracha @kylehilltruss @rogeruiz @jcbcapps
```
rogeruiz, line 3: What prompted this change? Was it our pre-commit configuration around Markdown linting?

package.json
```diff
@@ -47,20 +47,21 @@
   },
   "homepage": "https://github.com/trussworks/react-uswds#readme",
   "peerDependencies": {
+    "@uswds/uswds": "3.1.0",
     "react": "^16.x || ^17.x || ^18.x",
-    "react-dom": "^16.x || ^17.x || ^18.x",
-    "@uswds/uswds": "3.1.0"
```
rogeruiz, line 52: Are we removing the `@uswds/uswds` peer dependency here?

rogeruiz (approved):
lgtm 🌈

Approving this to help prevent blocking. But I have a question around peer-dependencies. 

Also since most of these were minor releases, what do we think about relaxing the `^` dependency versioning to have the last value be a `0` so that we don't have to update the dependencies at such a granular level?

We could replace some of our dependencies with a `~` rather than a `^` which would allow us to have minor updates with ease. 

[=> SO answer around clarity about carat versus tilde in package.json](https://stackoverflow.com/a/22345808/2096221)

Thanks for getting to this work @gidjin 🏆

jcbcapps (approved):

CODEOWNERS
gidjin: I'm not sure, I think someone's editor _helpfully_ converted markdown bullet points from `*` to `-` but turns out `CODEOWNERS` isn't a markdown file and `*` is referring to all files and `-` means nothing. This happened on my current project too, I suspect VSCode or something similar. But I have no actual proof of such things but this is the fix for why our PRs are not requiring owners at the moment.

package.json
gidjin: Not removed, it was alphabetized by my editor and is now the first one, which seems reasonable.

gidjin:
> lgtm 🌈
> 
> Approving this to help prevent blocking. But I have a question around peer-dependencies.
> 
> Also since most of these were minor releases, what do we think about relaxing the `^` dependency versioning to have the last value be a `0` so that we don't have to update the dependencies at such a granular level?
> 
> We could replace some of our dependencies with a `~` rather than a `^` which would allow us to have minor updates with ease.
> 
> [=> SO answer around clarity about carat versus tilde in package.json](https://stackoverflow.com/a/22345808/2096221)
> 
> Right now because our dependency versions don't include a left-most zero digit, we might as well be hard-coding the version and not use the `^` nor the `~` in our dependencies.
> 
> Thanks for getting to this work @gidjin 🏆

If I understand correctly from the link using `~` would make it only update patch versions vs minor and patch versions like `^`. So that feels more strict to me, so I'm not sure I am following your suggestion. Or are you saying only allowing patch versions would be easier to manage since minor upgrades would take manual intervention vs be caught by dependabot?

PR #2238, opened 2023-01-26
feat: Allow custom onClick function to be passed to ModalToggleButton

brandonlenz:
# Summary

ModalToggleButton could be enhanced by allowing users to pass a custom onClick function through to it.

I chose to run the custom onClick before actually toggling the modal somewhat arbitrarily, but it felt like the right move. Feel free to persuade me otherwise if you have a strong opinion.

## How To Test

Unit test covers the onClick being passed through


gidjin (approved):
LGTM, I think before the toggle is good, I can see an argument for both but I'm fine adding a second one that is after if/when the need for that arises.

brandonlenz:
> LGTM, I think before the toggle is good, I can see an argument for both but I'm fine adding a second one that is after if/when the need for that arises.

Yea good idea. An `onAfterToggle` or similar could be nice for that

PR #2243, opened 2023-02-16
chore: Remove CircleCI cruft from old deployment workflow

rpdelaney:
# Summary

- Remove CircleCI configuration and ./scripts/*

CircleCI is blocking dependabot, and it appears that these failing workflows are already replicated in GHA.


brandonlenz (approved):
It's becoming clearer to me that this repo is needing some devoted attention. But I'm with you, let's migrate away from CircleCI and build back up incrementally

PR #2256, opened 2023-02-21
chore: Update dependency reviewers

rpdelaney:
# Summary


.github/dependabot.yml
```diff
@@ -8,8 +8,7 @@ updates:
       separator: '-'
     open-pull-requests-limit: 10
     reviewers:
-      - "haworku"
-      - "ahobson"
```
gidjin, line 12: Question: Curious why you removed @haworku and @ahobson ? Maybe we can keep one of them at least in addition to me? Also I'm not familiar with dependabot configuration, is this just who is allowed to approve dependabot PRs? If so I'd add all the people listed as code owners in the `CODEOWNERS` file.

rpdelaney: I don't know how dependabot works either anymore. We plan to migrate to renovatebot soon anyway. With that in mind this PR is probably premature.

PR #2269, opened 2023-02-25
feat: Add className prop to Table component

sawyerh:
# Summary

Add a `className` prop to the `Table` component so that projects can add utility classes, for instance.

## How To Test

- Unit test added


haworku (approved):
LGTM

Issue #2327, opened 2023-04-03
[fix] storybook deploy library is deprecated


gidjin:
We use the [storybook deployer](https://github.com/storybookjs/storybook-deployer) which is now deprecated as of Jan 16, 2023. We should replace it with [Deploy To Github Pages](https://github.com/marketplace/actions/deploy-to-github-pages) github action action. For reference I implemented this change in another repo as an example. https://github.com/USSF-ORBIT/ussf-portal-client/pull/962

PR #2328, opened 2023-04-03
build: Update to use new storybook deployer

gidjin:
# Summary

We use the [storybook deployer](https://github.com/storybookjs/storybook-deployer) which is now deprecated as of Jan 16, 2023. We should replace it with [Deploy To Github Pages](https://github.com/marketplace/actions/deploy-to-github-pages) github action action. For reference I implemented this change in another repo as an example. https://github.com/USSF-ORBIT/ussf-portal-client/pull/962
 
## Related Issues or PRs

#2327

## How To Test

No way to test this as it is only triggered upon release.


.github/workflows/deploy-storybook.yml
```diff
@@ -34,7 +37,13 @@ jobs:
...
-          GH_TOKEN: trussworks:${{ secrets.GITHUB_TOKEN }}
+      - name: Build Storybook
+        run: yarn build-storybook --modern
+
+      - name: Deploy 🚀
+        uses: JamesIves/github-pages-deploy-action@v4
+        with:
+          clean: true
+          clean-exclude: |
+            CNAME
```
rogeruiz, line 48: I read in the commit that this is _if_ it exists, but where is this meant to be deployed today? Do we have a URL for this now? Should we make a subdomain for the `truss.dev` domain?

rogeruiz (commented):
Looks okay. Thanks @gidjin. Leaving a comment on this to discuss it further.

.github/workflows/deploy-storybook.yml
gidjin: I put it in since `clean: true` would remove it, in case we decided to have a custom domain. But maybe it's better to leave it out until we make that choice? We don't have a domain for it now but I'm not opposed to a custom one. Likely there are links in the repo we'd have to adjust.

rogeruiz (approved):
lgtm 🌈

Issue #234, opened 2020-06-05
[feat] Add type enum prop to NavList


haworku:
**Does your feature request relate to a specific USWDS feature?**
[Header component](https://designsystem.digital.gov/components/header/) - specifically the `<NavList/>` subcomponent. 

**Is your feature request related to a problem? Please describe.**
Discussion of problem is in #187.

**Describe the solution you'd like**
- Add new `type` prop.
- Deprecate `primary`, `secondary`, `subnav`, `megamenu`, `footerSecondary`, props in favor of new prop. 
- Add new tests and adjust stories as needed.

This work should look very similar to #229.

**Describe alternatives you've considered**
n/a, we converged on a solution for handling these cases in the linked issue.

PR #2350, opened 2023-04-21
feat: Add logic to handle DatePicker value prop

### LINKED ISSUE #2348: [fix] DatePicker component value not properly updated when modified by other component

"@trussworks/react-uswds": "4.2.0",
"@uswds/uswds": "3.1.0",

**Describe the bug**

When the DatePicker `value` prop is set to some state value, and that state is updated by some other component, e.g. an input, the new value is not reflected in the DatePicker component. I tried assigning both the `defaultValue` and `value` props, but got the same results. I've also confirmed in the React Devtools that the prop value(s) are correctly updated, but the internal DatePicker state is not updated. Am I perhaps overlooking something?

I've pasted a minimal reproduction below.

```js
const Test = () => {
  const [data, setData] = useState({
    statusId: 0,
    endDate: null,
  });

  return (
    <div className="grid-row">
      <div className="grid-col">
        <label htmlFor="end-date">
          End Date
        </label>
        <DatePicker
          id="end-date"
          name="end-date"
          minDate={new Date().toISOString().split('T')[0]}
          defaultValue={data.endDate}
          value={data.endDate}
        />
      </div>
      <div className="grid-col">
        <input
          id="active"
          type="radio"
          value={0}
          checked={data.statusId === 0}
          onChange={(e) =>
            setData((prevState) => ({
              ...prevState,
              endDate: null,
              statusId: +e.target.value,
            }))
          }
        />
        <label htmlFor="active">Active</label>
      </div>
      <div className="grid-col">
        <input
          id="disabled"
          type="radio"
          value={1}
          checked={data.statusId === 1}
          onChange={(e) =>
            setData((prevState) => ({
              ...prevState,
              endDate: new Date().toISOString().split('T')[0],
              statusId: +e.target.value,
            }))
          }
        />
        <label htmlFor="disabled">Disabled</label>
      </div>
    </div>
  );
};
```

**Expected behavior**
Updating the value prop from somewhere else should properly update the DatePicker value.

adil-ahorger:
# Summary

## Related Issues or PRs

Closes #2348 

## How To Test

- Create a component with a DatePicker and other input, e.g. radio button, and some state.
- Set DatePicker `value` prop to state value.
- Update state value from other input.


gidjin:
Thanks for your contribution. I want to get some other code owners to review this as well. In the mean time can you add a unit test for the changes you are proposing?

adil-ahorger:
I added unit tests as requested. I followed the same test cases as the `defaultValue` prop test suite, which I hope should be fine

haworku:
@adil-ahorger Thank you for https://github.com/trussworks/react-uswds/issues/234 and this PR. This is a valid use case for DatePicker that we aren't handling properly in the library right now. 

As written we cannot merge this initial PR for a couple reasons (outlined below). Closing for now,  **please re-open if you want to implement again and also feel free to discuss with us on the issue itself.** We recognize there aren't good docs on controlled v uncontrolled component behaviors  in our codebase. We have now made an [issue](https://github.com/trussworks/react-uswds/issues/2448)  to document related patterns better.

Notes on the code change:
1. With this change , the `onChange` function  would start to be called outside of any direct user interaction with DatePicker. We think this goes against HTML defaults. For example it would be called even when the component initially loads. This can also be seen in the test case `clears the input values if the value prop is falsy`.
2.  There is concern that surfacing  `value`  as an additional prop could lead to confusion with `defaultValue` already in use in props. Instead, we want to see this use case handled by wrapped the component in [`forwardRef`](https://react.dev/reference/react/forwardRef) to take over control and achieve a similar result. Here are [two](https://github.com/trussworks/react-uswds/pull/1165) [examples](https://github.com/trussworks/react-uswds/pull/1236)) from previous PRs.

PR #2378, opened 2023-05-02
docs: Add documentation for how we use All Contributors

brandonlenz:
# Summary

We didn't document how to use All Contributors, and I honestly have to look it up every time. Now it's documented!

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -21,6 +21,7 @@ We welcome contributions in the form of comments, issues, or pull requests with 
       - [`scope`:](#scope)
       - [`body`:](#body)
       - [`footer`:](#footer)
+  - [All-contributors](#all-contributors)
 
 ## Environment setup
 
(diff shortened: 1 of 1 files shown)


docs/contributing.md
```diff
@@ -209,10 +210,6 @@ When your branch is ready for review, open a new pull request into `main` and re
 
 When opening the pull request, it's important to understand the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) specification. **All pull requests opened into `main` must have a title that follows the [conventional commits spec](https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional)**. This generates an automated changelog entry and is later used to determine versioning for release. It is required to merge.
 
-#### ⚠️ Note to outside contributors
-
-Currently our CI cannot run directly on external PRs (work from outside the Truss organization) and prevents merge. To manage this, we pull these PRs into a separate branch that a CODEOWNER creates, run automation, and merge from there. Your initial PR will be closed with a comment and your work will be merged instead from the related PR. We will always ensure that attribution for your work will be associated with your account.
-
```
brandonlenz, line 215: Removed because this is no longer true. I believe thanks to @rpdelaney and @gidjin

brandonlenz: DangerJS was previously the culprit, if that jogs memories

rpdelaney:
> should _not_ be manually changed.

I wonder if a pre-commit hook could assist with enforcing this? 🤔

jcbcapps (approved):
LGTM! Thanks for documenting this!

brandonlenz:
> > should _not_ be manually changed.
> 
> I wonder if a pre-commit hook could assist with enforcing this? 🤔
> 
> I mean, we can make a hook that fails if a named file has changed at all. My open question is how that would interact with the service that _is_ allowed to change the file -- if the hook blocks it too, that's a fail!

It'd have to be smart enough to identify certain parts of each file being changed. 

For example, only [this section](https://github.com/trussworks/react-uswds/blob/main/.all-contributorsrc#L12) of .all-contributorsrc should not be manually edited, which you can only really identify if it is updated without corresponding changes to [this section](https://github.com/trussworks/react-uswds/blob/91697014151bd312a77a5d48c4fa6ee424012d09/README.md?plain=1#L133) of the README

jcbcapps (approved):

brandonlenz:
@all-contributors 
please add @brandonlenz for maintenance,bug
please add @rpdelaney  for maintenance,infra

brandonlenz:
> @all-contributors please add @brandonlenz for maintenance,bug please add @rpdelaney for maintenance,infra

Hmm that was mostly to test that allcontributors bot now uses conventional commits. It _does_!

Looks like it has a wait time between creating PRs, might be a rate limit safeguard, since it was almost a minute between the PR for me and for @rpdelaney.

PR #2383, opened 2023-05-03
feat: InputGroup component

### LINKED ISSUE #1692: New Component: InputGroup
# Input Group

## USWDS Version

Not sure when it was introduced, but some other components (e.g. [Input Prefix and Suffix](https://designsystem.digital.gov/components/input-prefix-suffix/)) make use of this convenience component, and it seems like we should include it as well. It is very similar to `FormGroup`. 

## Description

A component that USWDS uses to group components (linked above), but not mentioned as its own exclusive component. Providing this convenience component may be nice

## Requirements (proposed)

Should be able to toggle the error state with a prop

**Prop Interface**

```JSX
interface InputGroupProps {
  children: React.ReactNode
  className?: string
  error?: boolean
}
```

```JSX
export const InputGroup = ({
  children,
  className,
  error,
}: InputGroupProps): React.ReactElement => {
  const classes = classnames(
    'usa-input-group',
    { 'usa-input-group--error': error },
    className
  )

  return (
    <div data-testid="inputGroup" className={classes}>
      {children}
    </div>
  )
}
```

jcbcapps:
# Summary

## Related Issues or PRs
Resolves #1692 

## How To Test
Check unit tests

I did not add a Storybook component for this, but rather updated the `InputPrefix` and `InputSuffix` components to use `InputGroup` instead of a `div`. Happy to come back and add a Storybook component, but wasn't sure how to approach that in this context. Let me know if I need to make that change!

### Screenshots (optional)

### DIFF
--- a/src/components/forms/InputGroup/InputGroup.test.tsx
+++ b/src/components/forms/InputGroup/InputGroup.test.tsx
@@ -1,0 +1,16 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+
+import { InputGroup } from './InputGroup'
+
+describe('InputGroup component', () => {
+  it('renders without errors', () => {
+    const { queryByTestId } = render(<InputGroup>My Input Group</InputGroup>)
+    expect(queryByTestId('inputGroup')).toBeInTheDocument()
+  })
+
+  it('renders its children', () => {
+    const { queryByText } = render(<InputGroup>My Input Group</InputGroup>)
+    expect(queryByText('My Input Group')).toBeInTheDocument()
+  })
+})
--- a/src/components/forms/InputGroup/InputGroup.tsx
+++ b/src/components/forms/InputGroup/InputGroup.tsx
@@ -1,7 +1,7 @@
 import React from 'react'
 import classnames from 'classnames'
 
-interface InputGroupProps {
+export interface InputGroupProps {
   children: React.ReactNode
   className?: string
   error?: boolean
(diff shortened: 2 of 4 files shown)


src/components/forms/InputGroup/InputGroup.test.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+
+import { InputGroup } from './InputGroup'
+
+describe('InputGroup component', () => {
```
brandonlenz, line 6: Maybe add a test for checking that the right className is applied?

brandonlenz: + with the error state

brandonlenz:
Love it! The more we create components for these classes, the more reacty and intuitive our library becomes to use! I might be preemptively commenting things since I saw another commit come in for something I was gonna mention. So far so good! I'll wait on the CI pipeline before approving!

src/components/forms/InputGroup/InputGroup.test.tsx
jcbcapps: Good catch! I was just thinking the tests could probably be more thorough. Thanks!

brandonlenz: @jcbcapps sorry, I shoulda been more clear. Can you add a test that verifies that the component has the default class name applied? (`usa-input-group`)

It can go in the `renders without errors` test, imo, since it's jsut verifying that the component built with the right uswds class

brandonlenz (commented):
One final thing, added a comment to thread about tests

Issue #2241, opened 2023-02-13
[fix] CharacterCount's TextArea uses a `defaultValue` default, which triggers react console errors 


brandonlenz:
**ReactUSWDS Version & USWDS Version:**

"@trussworks/react-uswds": "4.0.0",
"@uswds/uswds": "^3.3.0",

**Describe the bug**

<img width="710" alt="image" src="https://user-images.githubusercontent.com/15805554/218567540-df7731f8-29d8-463b-b77b-ced23835c306.png">

**To Reproduce**

1. Use any project that uses a form library like formik to set the `value` prop of a CharacterCount's TextArea 
2. Note the error console error on load

**Expected behavior**

Consumers should be able to pass `value` instead of `defaultValue`

PR #2397, opened 2023-05-09
fix: allow CharacterCount value ^ defaultValue

werdnanoslen:
# Summary

Allows either `value` or `defaultValue` on CharacterCount component

## Related Issues or PRs

#2241 

## How To Test

This branch is simply the fix, but since it is based on an interaction with something like Formik, do two things:
1. Run this branch as-is to show nothing bad was introduced
2. Check out [this temporary test branch](https://github.com/trussworks/react-uswds/tree/an-defaultvalue-2241__test)
3. `yarn`, or just install `formik`
4. `yarn storybook`
5. Open the ['test' story](http://localhost:9009/?path=/docs/components-charactercount--test) to see the error is no longer appearing given the conditions reported in #2241.

Once merged, please also delete the [test branch
](https://github.com/trussworks/react-uswds/tree/an-defaultvalue-2241__test)

### Screenshots (optional)
image

### DIFF
--- a/src/components/forms/CharacterCount/CharacterCount.tsx
+++ b/src/components/forms/CharacterCount/CharacterCount.tsx
@@ -37,6 +37,7 @@ type BaseCharacterCountProps = {
   id: string
   name: string
   maxLength: number
+  value?: string
   defaultValue?: string
   className?: string
   isTextArea?: boolean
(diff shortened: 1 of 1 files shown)


gidjin (commented):
LGTM, could you add a unit test to check that value is used properly if passed in?

werdnanoslen:
> LGTM, could you add a unit test to check that value is used properly if passed in?

done

brandonlenz (approved):

PR #2411, opened 2023-05-22
fix!: Unify icon usage

### LINKED ISSUE #1303: [feat] Unify icon usage

Dependent on new react-uswds Icon component. PR: https://github.com/trussworks/react-uswds/pull/1241
More context: https://github.com/trussworks/react-uswds/issues/992

USWDS 2.11 +

USWDS has moved away from Font Awesome towards Google Material Icons, as well as using SVG icon sprite. Some of the changes are discussed in this [video](https://youtu.be/OhtiTfKAIRw?t=2843). We will follow suit and update this codebase as we move to 2.11. This will require making some documentation edits as well as using our new [Icon ](https://designsystem.digital.gov/components/icon/) component.

Several steps to take here:
- [x] [Make sure react-uswds is on 2.11+](https://github.com/trussworks/react-uswds/issues/1304)
- [ ] Double check our components are using the correct uswds icons. `GovBanner` uses a lock icon for example.
- [ ] Replace usage of icons in our components with our react-uswds `<Icon>`. 
- [x] In our styles and assets documentation we're recommend use fontawesome per uswds' previous recommendation. Our documentation needs to be updated here to remove reference to fontawesome in favor of Material icons.
- [x] Remove usage of @fortawesome fontawesome in example app entirely
- [ ] USWDS examples set some intrinsic props on `<svg/>` tags to defaults (e.g. focusable and role). We should set the same defaults in our code,  but allow override these html attributes if the corresponding prop is passed to the Icon component

werdnanoslen:
# Summary

Addresses these two ACs in the original issue:
- Double check our components are using the correct uswds icons. GovBanner uses a lock icon for example.
- Replace usage of icons in our components with our react-uswds <Icon>.

## Related Issues or PRs

Resolves #1303

## How To Test

Confirm that all SVGs still in use are for images that aren't icons

### Screenshots (optional)
N/A, looks the same visually, and behaves the same with some attribute changes


werdnanoslen:
Social links I wasn't sure if it needed to change per se, given the styles referenced there turn it black based on its parent div and USWDS doesn't define their style directly. So since our implementation uses their styles, they are sort of 'supposed' to be blue out of their context, since they're links. I _think_ 🤷 

I'll fix the search though, could've sworn they were the same on my local. 🤔

werdnanoslen:
Ok yeah search definitely looks normal on my local in both Firefox & Chrome, 24px icon with 12px margin
image
image

werdnanoslen:
@brandonlenz are Happo diffs approvable?

src/components/Footer/SocialLinks/SocialLinks.test.tsx
```diff
@@ -40,7 +40,9 @@ describe('SocialLinks component', () => {
 describe('SocialLink component', () => {
   it('renders without errors', () => {
     const { container } = render(<SocialLink name="Instagram" />)
-    expect(container.querySelector('.usa-social-link')).toBeInTheDocument()
-    expect(screen.getByRole('img')).toHaveAttribute('alt', 'Instagram')
```
gidjin, line 44: Question: Should these have alt tags too?

gidjin (approved):
LGTM, left a question but it's not blocking.

src/components/Footer/SocialLinks/SocialLinks.test.tsx
werdnanoslen: no, the `alt` prop only applies to ``. Now that it's an `<svg>`, I have a line at the bottom of that block to check the `name` prop instead, which serves the same purpose.

PR #2415, opened 2023-05-25
fix: Creates new Select component to eventually replace Dropdown

werdnanoslen:
# Summary

USWDS 3.0 renamed Dropdown to Select. This PR duplicates the Dropdown component to a new Select component, and adds a deprecation warning to the story and component via console.warn

## Related Issues or PRs

Resolves #2308 

## How To Test

* Confirm Select behaves exactly the same as Dropdown
* Open Dropdown story to see deprecation warning on page and in console

### Screenshots (optional)
image

Screenshot 2023-05-25 at 11 37 19

PR #2434, opened 2023-06-05
Select Deprecation edits

brandonlenz:
# Summary

Batched PR feedback/suggestions for #2415 

## How To Test

Unit tests, storybook

### Screenshots (optional)

Dropdown still works, and displays deprecation warning:
image

Select of course also works, does not display deprecation warning:
image

### DIFF
--- a/src/components/forms/Dropdown/Dropdown.test.tsx
+++ b/src/components/forms/Dropdown/Dropdown.test.tsx
@@ -1,18 +1,24 @@
 import React from 'react'
 import { render } from '@testing-library/react'
+jest.mock('../../../deprecation')
+import { deprecationWarning } from '../../../deprecation'
 
-import { Select } from '../Select/Select'
+import { Dropdown } from './Dropdown'
 
 describe('Dropdown component', () => {
-  it('renders without errors', () => {
+  it('renders without errors, displaying a deprecation warning', () => {
     const { queryByTestId } = render(
-      <Select id="input-type-text" name="input-type-text">
+      <Dropdown id="input-type-text" name="input-type-text">
         <option>- Select - </option>
         <option value="value1">Option A</option>
         <option value="value2">Option B</option>
         <option value="value3">Option C</option>
-      </Select>
+      </Dropdown>
     )
     expect(queryByTestId('Select')).toBeInTheDocument()
+    expect(deprecationWarning).toHaveBeenCalledTimes(1)
+    expect(deprecationWarning).toHaveBeenCalledWith(
+      'Dropdown is deprecated and will be removed in the future. Please use the Select component instead.'
+    )
   })
 })
--- a/src/components/forms/Dropdown/Dropdown.tsx
+++ b/src/components/forms/Dropdown/Dropdown.tsx
@@ -1,52 +1,11 @@
 /*
-  @deprecated Make updates to the Select component instead
-  @todo Remove this component
+  TODO: Remove this component
 */
 
-import React from 'react'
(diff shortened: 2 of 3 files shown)


src/components/forms/Dropdown/Dropdown.test.tsx
```diff
@@ -1,18 +1,24 @@
...
 
-import { Select } from '../Select/Select'
+import { Dropdown } from './Dropdown'
 
 describe('Dropdown component', () => {
-  it('renders without errors', () => {
+  it('renders without errors, displaying a deprecation warning', () => {
     const { queryByTestId } = render(
-      <Select id="input-type-text" name="input-type-text">
+      <Dropdown id="input-type-text" name="input-type-text">
```
brandonlenz, line 11: This test should still be for the Dropdown component. Might've accidentally gotten updated by IDE in #2415

src/components/forms/Dropdown/Dropdown.test.tsx
```diff
@@ -1,18 +1,24 @@
...
         <option value="value2">Option B</option>
         <option value="value3">Option C</option>
-      </Select>
+      </Dropdown>
     )
     expect(queryByTestId('Select')).toBeInTheDocument()
+    expect(deprecationWarning).toHaveBeenCalledTimes(1)
+    expect(deprecationWarning).toHaveBeenCalledWith(
+      'Dropdown is deprecated and will be removed in the future. Please use the Select component instead.'
+    )
```
brandonlenz, line 22: Not really necessary, since deprecationWarning and withDeprecationWarning have tests, but it was peace of mind for me making sure that my suggestions actually work because this was just a theory until confirming it 😉

src/components/forms/Dropdown/Dropdown.tsx
```diff
@@ -1,52 +1,11 @@
...
-      {children}
-    </DeprecatedDropdown>
-  )
-}
-
-export default Dropdown
+export const Dropdown = withDeprecationWarning(
+  Select,
+  'Dropdown is deprecated and will be removed in the future. Please use the Select component instead.'
+)
```
brandonlenz, line 11: @werdnanoslen this was the main thing I wanted to suggest. `withDeprecationWarning` allows you to wrap any component, and it just returns a _new_ component with the name you give it, which renders with a `useEffect` that calls `deprecationWarning` on initial render, and also renders the component you pass in.

So no need to define and pass around props, it basically inherits all of that from Select. `Dropdown` is now just `Select` with the old name, that also renders a warning.

🪄 Magic 🪄

src/components/forms/Select/Select.tsx
```diff
@@ -48,5 +48,3 @@ export const Select = ({
     </select>
   )
 }
-
-export default Select
```
brandonlenz, line 52: I noticed this leading to some new(?) eslint warnings. Since the default export had no usages, and we seem to prefer the named export as a convention, this can be removed 😄

src/components/forms/Dropdown/Dropdown.tsx
werdnanoslen: ahh got it, haven't used a pattern like this (intentionally) before, but that makes sense!

src/components/forms/Dropdown/Dropdown.test.tsx
werdnanoslen: bah i tried adding this but couldn't get it to work

PR #2467, opened 2023-06-28
fix: a11y: adds css to hide search clear button that causes accessibility issues

shkeating:
# Summary

## Related Issues or PRs

**USWDS issue:** https://github.com/uswds/uswds/issues/5277

**Chromium issue:** https://bugs.chromium.org/p/chromium/issues/detail?id=1147314

## How To Test

`yarn storybook`

http://localhost:9009/?path=/story/components-search-search--default-search 

1. type in the search box in Chrome
2. make sure the little X icon does not show up

### Screenshots (optional)
before:
image

after:
image


brandonlenz (approved):

brandonlenz:
Since the SCSS implementation of this library is always hard (to me) I'll be curious to know if you have to do anything special to make sure to use react-uswds' index.scss when setting up react uswds + uswds on consuming projects

PR #2474, opened 2023-07-03
chore(ci): Perform linting and formatting in CI

### LINKED ISSUE #2377: [feat] Run prettier and elsint in CI

Upgrading husky stopped our pre-commit checks from running. 

The new version of husky requires setup that is unfriendly for consumers for a component library like this. We implemented it before but had to revert it #926 

While figuring out what we should do as an alternative to husky for pre-commit checks, at least run/lint in CI so that we don't merge unformatted or un linted code.

Go ahead and fix pre-commits, but we likely want this CI sanity check anyways, and this requires little to no decision energy

brandonlenz:
# Summary

- Removes Husky and lint-staged which are not used (pre-commit is not configured, and Husky did not have a solution suitable for our needs as a component library)
  - If we want to set up precommit checks again with these tools, no problem, we can bring them back at that time. 
- Updates our actions to newer versions (particularly checkout and setup-node)
  - Also updates setup-node to use the `node-version-file` property rather than explicitly re-declaring the node version across files. Should make our node upgrade process easier
- The code changes are from running `yarn format:fix` to get us caught up in this PR.

## Related issues or PRs

Resolves #2377

## How To Test

It ran on this PR. Before updating the format of noncompliant files:
image

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -38,6 +38,3 @@ updates:
         versions:
           - '>=7.0.0'
       - dependency-name: 'uswds'
-      - dependency-name: 'husky'
-        versions:
-          - '>=5.0.0'
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -19,10 +19,10 @@ jobs:
 
     steps:
       - name: Check out repository
-        uses: actions/checkout@v2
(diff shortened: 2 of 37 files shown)


jcbcapps (approved):
LGTM!

jcbcapps (approved):

PR #2494, opened 2023-07-14
build: disable include-v-in-tag option

gidjin:
# Summary

Forgot to turn off the option that adds a v to the version tag name. Disabling it because we haven't been having one in the past and didn't want to change the pattern.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -20,6 +20,7 @@ jobs:
           release-type: node
           package-name: '@trussworks/react-uswds'
           release-labels: 'type: release'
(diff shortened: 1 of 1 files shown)


.github/workflows/release-please.yml
```diff
@@ -20,6 +20,7 @@ jobs:
           release-type: node
           package-name: '@trussworks/react-uswds'
           release-labels: 'type: release'
+          include-v-in-tag: false
```
brandonlenz, line 23: I love that this is an option. Very specific flag

brandonlenz (approved):

PR #2505, opened 2023-07-21
build(tools): Add documenation for supporting tool version managers like asdf and mise-en-place

brandonlenz:
# Summary

Nodenv is a great tool and many folks in the JS ecosystem use it. We support and recommend it today, and include a `.node-version` to help with that.

mise and asdf are also very popular, but typically more for developers working in a mixed tech stack (JS & Go, JS & Java, etc).

asdf and mise users likely need those tools for other projects, and should not be expected to switch to nodenv

It is also not recommended to use multiple node version managers, so either only asdf or only nodenv/n/nvm is recommended. Otherwise the tools can compete, causing unexpected shimming of Node to your PATH.

This PR preserves support for Nodenv for the folks who use that tool, but describes how to use asdf and mise for folks who use those tools. It is up to contributors to install, use, and manage the tool of their preference.

## How To Test

As an asdf user, I tested by adding `legacy_version_file`, attempting to install the node version specified by the project. It then recommended I install the version from the .node-version file, which I did. Then I verified the installed version came up when within the react-uswds directory:

```
% node --version
> v18.15.0
```


docs/contributing.md
```diff
@@ -25,7 +25,9 @@ We welcome contributions in the form of comments, issues, or pull requests with
 
 ## Environment setup
 
-1. Use the node environment manager of your choice, but make sure you have the required version specified in `.node-version`. We recommend using [nodenv](https://github.com/nodenv/nodenv) to manage your node versions, but you can also use [homebrew](https://brew.sh/). More info can be found here: [how to install Node.js](https://nodejs.dev/how-to-install-nodejs)
```
brandonlenz, line 28: The nodejs link was a 404, so I removed it.

Additionally, it is not recommended to use brew for Node, as brew prefers to use `latest` whenever possible. Contributors using brew will likely continue to use it (and hopefully know how to deal with those quirks), so I think it best to recommend a version manager and leave it at that.

gidjin: Something I've wondered for a while about asdf is that it supports `.node-version` file, is it better to have two files or just let asdf users know how to enable `legacy_version_file` for asdf? 

https://github.com/asdf-vm/asdf-nodejs#nvmrc-and-node-version-support

gidjin (commented):
Left a note that I'd like to chat about before approving, otherwise this is good.

docs/contributing.md
werdnanoslen: @gidjin did y'all ever chat about this?

brandonlenz: While I'm still on asdf, I do plan on switching to Mise which [supports .tool-versions](https://mise.jdx.dev/configuration.html#tool-versions), [and .node-version](https://mise.jdx.dev/configuration.html#idiomatic-version-files), so this should be fine for that tool as well.

I just updated the PR for both cases

gidjin (approved):

PR #2526, opened 2023-07-25
feat: changed template links to buttons

werdnanoslen:
# Summary

Changed links to buttons in signin templates.

## Related Issues or PRs

Resolves #2508 

## How To Test

Open the [signin 'form' template](http://localhost:9009/?path=/story/components-form-templates--sign-in-form) and [signin 'page' template](http://localhost:9009/?path=/story/page-templates-authentication-pages--sign-in)

PR #2531, opened 2023-07-26
feat: added language selector

### LINKED ISSUE #2516: New Component: Language Selector
# [Language Selector](https://designsystem.digital.gov/components/language-selector/)

## USWDS Version

Released in 3.3.0 via https://github.com/uswds/uswds/pull/4937

## Description

Basically is a button component that either...
* (if there are only 2 languages) toggles its content/function between the two languages ([example](https://www.usa.gov/))
   * Note that the demo on the USWDS site is broken (only shows Español on click)
* (if >2 languages) shows a [header](https://designsystem.digital.gov/components/header/)-like dropdown

## Requirements (proposed)

**Prop Interface**

```JSX
Props = {
}
```

```JSX
return (
)
```

**More Details**

- `<div>`root element

## Other Considerations

-

werdnanoslen:
# Summary

1. Adds [Language Selector](https://designsystem.digital.gov/components/language-selector/) component
2. Adds a pass-through prop to Menu for showing the right kind of NavList (using subnav as default if not given, as is currently hard-coded)
3. Adds language type to NavList

## Related Issues or PRs

Resolves #2516

## How To Test

Open the stories for [LanguageSelector](http://localhost:9009/?path=/story/components-languageselector--two-languages-as-a-link)

### Screenshots (optional)
Two languages just toggles:
two

More than two language shows a dropdown:
link


werdnanoslen:
This uses links per [USWDS' implementation](https://designsystem.digital.gov/components/language-selector/#language-selector--three-or-more-preview), but I think it's also likely that people will want to implement in-page with buttons using `onClick` (similar to their [guidance here](https://github.com/uswds/uswds/pull/4847), resulting in our [other PR](https://github.com/trussworks/react-uswds/pull/2526)). This is also how my current project expects to use it, so I'll look into a button option as a follow-up commit here or future PR.

werdnanoslen:
> This uses links per [USWDS' implementation](https://designsystem.digital.gov/components/language-selector/#language-selector--three-or-more-preview), but I think it's also likely that people will want to implement in-page with buttons using `onClick` (similar to their [guidance here](https://github.com/uswds/uswds/pull/4847), resulting in our [other PR](https://github.com/trussworks/react-uswds/pull/2526)). This is also how my current project expects to use it, so I'll look into a button option as a follow-up commit here or future PR.

Latest commit adds this, so ready for review (pending styles ⚠️ )!

werdnanoslen:
> > This uses links per [USWDS' implementation](https://designsystem.digital.gov/components/language-selector/#language-selector--three-or-more-preview), but I think it's also likely that people will want to implement in-page with buttons using `onClick` (similar to their [guidance here](https://github.com/uswds/uswds/pull/4847), resulting in our [other PR](https://github.com/trussworks/react-uswds/pull/2526)). This is also how my current project expects to use it, so I'll look into a button option as a follow-up commit here or future PR.
> 
> Latest commit adds this, so ready for review (pending styles ⚠️ )!

Fixed styling now that 3.3 styles are merged, and [filed an issue](https://github.com/uswds/uswds/issues/5409) to let us remove that later if merged. This is fully ready for review now

werdnanoslen:
Hm, not sure I'm interpreting the build rule correctly, but given my [`@extend`](https://github.com/trussworks/react-uswds/pull/2531/files#diff-6d1ac1e7072d002fd68fa47aec09fccdbdd03bfc6b3366e3a7bd5fd96239abf3R22) is the simplest way I know of to reference code from USWDS and not copy-paste it into our code, could [this error](https://github.com/trussworks/react-uswds/actions/runs/5720183598/job/15499617241?pr=2531#step:5:9) be overridden?

src/components/LanguageSelector/LanguageSelector.tsx
```diff
@@ -0,0 +1,122 @@
...
+    )
+  } else {
+    if (label)
+      console.warn(
+        "LanguageSelector's label is not used when only two languages are available."
+      )
+    const [langIndex, setLangIndex] = useState(false)
+    const curLang = langs[Number(langIndex)]
+    const curLangNotEn =
+      curLang.attr && curLang.attr !== 'en' ? curLang.attr : undefined
```
jcbcapps, line 97: Bit of a nitpick: I'm wondering if you could write a test to cover when `curLangNotEn` is `undefined`?

src/styles/index.scss
```diff
@@ -15,3 +15,9 @@
 .usa-search [type='search']::-webkit-search-cancel-button {
   display: none;
 }
+
+// This just allows buttons as well as links in the LanguageSelector
+// It can be removed when https://github.com/uswds/uswds/issues/5409 is fixed
+.usa-language__submenu .usa-language__submenu-item button {
+  @extend a;
+}
```
jcbcapps, line 23: I'm wondering if the failing lint check is related to this?

jcbcapps (commented):
This looks great! Just left a couple of minor questions. Didn't feel like they were big enough to request changes.

src/styles/index.scss
werdnanoslen: it is, I [commented above](https://github.com/trussworks/react-uswds/pull/2531#issuecomment-1659272663)

jcbcapps: Sorry I missed that! Yeah that is a weird one. It may need to be overridden like you said, but not sure.

src/components/LanguageSelector/LanguageSelector.tsx
werdnanoslen: I'm actually going to slightly refactor this to be more language agnostic, will request a review on the changes...

werdnanoslen: Ok, i've removed this bit so that the attr always shows (adds a few extraneous bytes to a page, but make this more language-flexible), and I renamed the input object to be called "lang_local" so that we don't assume this is used on a primarily `en` site

src/components/LanguageSelector/LanguageSelectorButton.tsx
```diff
@@ -0,0 +1,36 @@
...
+  label,
+  labelAttr,
+  isOpen,
+  onToggle,
+  className,
+  ...buttonProps
+}: LanguageSelectorButtonProps &
+  JSX.IntrinsicElements['button']): React.ReactElement => {
+  const classes = classnames('usa-button', 'usa-language__link', className)
+  const labelNotEn = labelAttr && <span lang={labelAttr}>{label}</span>
```
jcbcapps, line 21: Since you removed the `curLangNotEn` variable in `LanguageSelector` and are no longer using the `labelAttr` prop from there, do we still need this `labelNotEn` variable? Looking at the test coverage, the `<span>` here is not checked, and it just made me wonder if it is still needed.

src/components/LanguageSelector/LanguageSelectorButton.tsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react'
+import classnames from 'classnames'
+
+type LanguageSelectorButtonProps = {
+  label: string
+  labelAttr?: string
```
jcbcapps, line 6: Do we still need the `labelAttr` prop? It seems that in order for this prop to be utilized, `LanguageSelectButton` would need to be used outside of the `LanguageSelector` component.

src/styles/index.scss
werdnanoslen: @shkeating any ideas?

src/components/LanguageSelector/LanguageSelectorButton.tsx
werdnanoslen: I'd still keep it on the button so that the button contents get the same lang attr treatment. I just updated it to handle that better if you want to see those changes

src/components/LanguageSelector/LanguageSelectorButton.tsx
werdnanoslen: it's still useful for the 2 language variant, where the language's text should be tagged. I've updated it a bit to reflect that

shkeating:
so what I have gathered so far is the lint tool is mad at how you used @extend because it _can_ cause specificity issues. I personally don't feel this particular instance carries much of that risk, especially as a temporary solution parked in the index file. I'm seeing if there is a way to push the link styles from uswds dynamically that doesn't use @extend. 

I'm pretty sure those styles are stored in a mixin already which we can call with @include for a similarly elegant solution that does not make us have to override lint rules

shkeating:
> so what I have gathered so far is the lint tool is mad at how you used @extend because it _can_ cause specificity issues. I personally don't feel this particular instance carries much of that risk, especially as a temporary solution parked in the index file. I'm seeing if there is a way to push the link styles from uswds dynamically that doesn't use @extend.
> 
> I'm pretty sure those styles are stored in a mixin already which we can call with @include for a similarly elegant solution that does not make us have to override lint rules

im now seeing that the styles for this particular link instance are pretty custom. maybe our quickest path forward is just grabbing the scss and pasting it where @extend would have output it. not as sleek but it is a temp solution. im trying that now to see if it works

src/styles/index.scss
shkeating, line 1: `@use 'uswds-core' as *;` would need to be added up top for the suggestion to work

src/styles/index.scss
```diff
@@ -15,3 +15,9 @@
 .usa-search [type='search']::-webkit-search-cancel-button {
   display: none;
 }
+
+// This just allows buttons as well as links in the LanguageSelector
+// It can be removed when https://github.com/uswds/uswds/issues/5409 is fixed
+.usa-language__submenu .usa-language__submenu-item button {
+  @extend a;
```
shkeating, line 22:
```suggestion
    color: color('white');
  display: block;
  line-height: line-height($theme-navigation-font-family, 3);
  padding: 0;
  padding: units(1);
  text-decoration: none;
  width: 100%;  /* this is something that was actually missing when we were doing the extend because button widths hve different default widths than block element links */

  &:focus {
    outline-offset: units('neg-05');
  }

  &:hover {
    color: color('white');
    text-decoration: underline;
  }
```

shkeating: this just grips the scss [right out of the package](https://github.com/uswds/uswds/blob/develop/packages/usa-language-selector/src/styles/_usa-language-selector.scss), replicating the @extend output.

only advantage of this over using @extend is it stops us from overriding lint rules, and it makes sure the focus indicator behaves the same with with width assignment I added, but you could also add that using the @extend.

I would defer to the app eng folks on how they feel about overriding lint rules, but if you are looking to get this going this is how I'd get around it.

src/components/LanguageSelector/LanguageSelector.tsx
```diff
@@ -0,0 +1,121 @@
...
+  const classes = classnames(
+    'usa-language-container',
+    {
+      [`usa-language--small`]: small !== undefined,
+    },
+    className
+  )
+
+  if (langs.length > 2) {
+    const [isOpen, setIsOpen] = useState(false)
```
shimonacarvalho, line 38: Thought i'd jump the gun a bit and try this in IRS project. Getting the following eslint error there:
```
React Hook "useState" is called conditionally. React Hooks must be called in the exact same order in every component render.eslint[react-hooks/rules-of-hooks](https://reactjs.org/docs/hooks-rules.html)
```

src/components/LanguageSelector/LanguageSelector.tsx
```diff
@@ -0,0 +1,121 @@
...
+      </div>
+    )
+  } else {
+    if (label) {
+      console.warn(
+        "LanguageSelector's label is not used when only two languages are available."
+      )
+    }
+    const [langIndex, setLangIndex] = useState(false)
+    const curLang = langs[Number(langIndex)]
```
shimonacarvalho, line 96: Getting the same eslint error about this instance of`useState` as well.

src/styles/index.scss
werdnanoslen: thanks! committed that

src/components/LanguageSelector/LanguageSelector.tsx
werdnanoslen: fixed

src/components/LanguageSelector/LanguageSelector.tsx
werdnanoslen: fixed

shkeating (dismissed):

src/styles/index.scss
werdnanoslen: hm, we don't seem to be using utilities _anywhere_, and I couldn't get it to work, so I've just copied the compiled styles over here for now.

werdnanoslen:
Happo all looks as intended, except for the ios-mobile ones:
Screenshot 2023-08-09 at 2 22 28 PM


However, I can't reproduce this myself, so maybe it's a Happo issue?
Screenshot 2023-08-09 at 2 21 09 PM

I'm going to self-approve Happo unless someone is able to reproduce that issue.

src/styles/index.scss
werdnanoslen: i tried, but it kept giving me node-related loading issues that are too numerous and complicated to share async

werdnanoslen: I tried, but i kept getting [build errors](https://github.com/trussworks/react-uswds/actions/runs/5812487162)

shkeating (approved):
also cannot reproduce the safari issue

PR #2538, opened 2023-07-31
feat: input mask

### LINKED ISSUE #2517: New Component: Input Mask
# [Input Mask](https://designsystem.digital.gov/components/input-mask/)

## USWDS Version

Released in 3.3.0 via https://github.com/uswds/uswds/pull/4884

## Description

Adds visual styling and auto content formatting to [text input](https://designsystem.digital.gov/components/text-input/) components

## Requirements (proposed)

**Prop Interface**

```JSX
Props = {
}
```

```JSX
return (
)
```

**More Details**

- `<div>`root element

## Other Considerations

- Consider extending the text input field component with this functionality instead of a whole new component

werdnanoslen:
# Summary

Adds [input masks](https://designsystem.digital.gov/components/input-mask/) to text inputs

## Related Issues or PRs

Resolves #2517 

## How To Test

Open the last story for Text input: [Input Mask](http://localhost:9009/?path=/story/components-text-input--input-mask)

### Screenshots (optional)
This is me just pianoing 1-9, showing that it autoformats:
Aug-02-2023 13-53-06


shkeating:
looks good to me~

shkeating (dismissed):

werdnanoslen:
> ~looks good to me~
> 
> whoops, I overlooked [an ANDI warning](https://www.ssa.gov/accessibility/andi/help/alerts.html?dby_alone) we might want investigate
> image
> 
> I'm seeing it suggest using `aria-labelledby` instead of `aria-describedby`

It already has a label though (associated with `for=`), and describedby is supposed to be for [extra info](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby) on top of a label. Also [USWDS uses](https://designsystem.digital.gov/components/input-mask/#component-code) describedby

src/components/forms/TextInput/TextInput.test.tsx
```diff
@@ -11,6 +11,47 @@ describe('TextInput component', () => {
...
+      fireEvent.change(input, { target: { value: 'A1B2C3' } })
+      expect((input as HTMLInputElement).value).toBe('A1B 2C3')
+    })
+
+    it('rejects entry based on charset', () => {
+      const { input } = setup()
+      fireEvent.change(input, { target: { value: 'A1B 2CC' } })
+      expect((input as HTMLInputElement).value).toBe('A1B 2C')
+    })
+  })
```
brandonlenz, line 53: **praise**: ❤️ ❤️ ❤️ to this style of testing. So glad it's catching on. 

([If anyone else is curious about it](https://derekndavis.com/posts/maintainable-testing-react-testing-library))

src/components/forms/TextInput/TextInput.test.tsx
```diff
@@ -11,6 +11,47 @@ describe('TextInput component', () => {
...
+
+    it('renders with class and placeholder', () => {
+      const { input, mask } = setup()
+      expect(input).toHaveClass('usa-masked')
+      expect(mask).toHaveTextContent('___ ___')
+    })
+
+    it('autoformats inputted text', () => {
+      const { input } = setup()
+      fireEvent.change(input, { target: { value: 'A1B2C3' } })
```
brandonlenz, line 44: **issue (convention, test-robustness)**: `userEvent` is preferred over `fireEvent` which would looks something like this (you'll need to `async` the test function and add the import from `@testing-library/user-event`)

more info: https://testing-library.com/docs/user-event/intro/

> fireEvent dispatches DOM events, whereas user-event simulates full interactions, which may fire multiple events and do additional checks along the way.

```suggestion
      const user = userEvent.setup()
      await user.type(input, 'A1B2C3')
```

src/components/forms/TextInput/TextInput.test.tsx
```diff
@@ -11,6 +11,47 @@ describe('TextInput component', () => {
...
+      expect((input as HTMLInputElement).value).toBe('A1B 2C3')
+    })
+
+    it('rejects entry based on charset', () => {
+      const { input } = setup()
+      fireEvent.change(input, { target: { value: 'A1B 2CC' } })
```
brandonlenz, line 50: **suggestion**: same thing here, should be `userEvent`

src/components/forms/TextInput/TextInput.tsx
```diff
@@ -50,21 +55,83 @@ export const TextInput = ({
...
+        inputProps.maxLength || !mask ? inputProps.maxLength : mask.length
+      }
+      onChange={mask ? handleChange : inputProps.onChange}
+      value={mask ? inputValue : inputProps.value}
       {...inputProps}
     />
   )
+
+  const [maskValue, setMaskValue] = useState(mask)
+  const [iValue, setIValue] = useState(<i></i>)
```
brandonlenz, line 82: **suggestion**: Perhaps the initial state of iValue should be `undefined` so we don't render empty tags. Thoughts?

src/components/forms/TextInput/TextInput.tsx
```diff
@@ -50,21 +55,83 @@ export const TextInput = ({
...
+  function handleChange(e: React.ChangeEvent<HTMLInputElement>): void {
```
brandonlenz, line 83: nit (convention): Can you update this to use arrow function syntax?

src/components/forms/TextInput/TextInput.tsx
```diff
@@ -50,21 +55,83 @@ export const TextInput = ({
...
+          data-testid={`${id}Mask`}>
+          {iValue}
+          {maskValue}
+        </span>
+        {input}
+      </span>
+    )
+  } else {
+    return input
+  }
```
brandonlenz, line 134: **suggestion (not blocking)**: On a quick scan, this looks like a pretty good reason to make InputMask its own component, and have it just render a TextInput in place of `{input}`. Since USWDS considers it a separate component as well, that makes it a bit more compelling

That would keep TextInput's code simple (and likely unchanged, which is good, means no potential for regressions), while leaning on componentization and composition more heavily, which are really good things.

Not gonna block on it because you might have tried that or discovered a reason why that won't work.  If it is possible (or if it might be possible) and we merge this, I'll recommend creating a github issue to split them apart and perform a subsequent breaking release.

src/components/forms/TextInput/TextInput.tsx
```diff
@@ -50,21 +55,83 @@ export const TextInput = ({
...
+      ? value.replace(/\W/g, '')
+      : value.replace(/\D/g, '')
+    const isInteger = (v: string) => !Number.isNaN(parseInt(v, 10))
+    const isLetter = (v: string) => (v ? v.match(/[A-Z]/i) : false)
+    const maskedNumber = '_#dDmMyY9'
+    const maskedLetter = 'A'
+    let newValue = ''
+    for (let m = 0, v = 0; m < maskData.length; m++) {
+      const isInt = isInteger(strippedValue[v])
+      const isLet = isLetter(strippedValue[v])
```
brandonlenz, line 97: **nit**: high preference for spelling things out, but naming is nitpicky

```suggestion
      const charIsInteger = isInteger(strippedValue[v])
      const charIsLetter = isLetter(strippedValue[v])
```

brandonlenz: **suggestion (code flow)**: This also looks like you're trying to balance two separate conditions in parallel. 

I'm wondering if you can have `if (isInteger(...))` `else if (isLetter(...))` code blocks to separate the logic, rather than trying to keep track of two mutually exclusive things at the same time

brandonlenz (changes requested):
Lemme know if you have any questions about this feedback!

src/components/forms/TextInput/TextInput.tsx
werdnanoslen: yeah, just requires setting the type more explicitly

src/components/forms/TextInput/TextInput.tsx
```diff
@@ -50,21 +55,82 @@ export const TextInput = ({
...
       'usa-input--medium': isMedium,
+      'usa-masked': mask,
     },
     className
   )
 
-  return (
+  const [inputValue, setInputValue] = useState('')
+  const [maskValue, setMaskValue] = useState(mask)
+  const [iValue, setIValue] = useState<React.DetailedHTMLProps<React.HTMLAttributes<HTMLElement>, HTMLElement>>()
```
brandonlenz, line 65: Does this not work? (And sorry, when I said `undefined` earlier, I think `null` is actually the right choice.

```suggestion
  const [iValue, setIValue] = useState<HTMLElement | null>(null)
```

src/components/forms/TextInput/TextInput.tsx
brandonlenz: (update to this thread [here](https://github.com/trussworks/react-uswds/pull/2538#discussion_r1286289595))

src/components/forms/TextInput/TextInput.tsx
werdnanoslen: No, compiler tells me, `Argument of type 'Element' is not assignable to parameter of type 'SetStateAction<HTMLElement | null>'.ts(2345)` (same result if i type the state as just `Element`

src/components/forms/TextInput/TextInput.tsx
werdnanoslen: I cribbed the logic from the USWDS implementation, and it seems to simplify the _treatment_ at the expense of some complicated _conditions_. I attempted to simplify the conditions, but it didn't result in particularly better looking code though I didn't test performance.

src/components/forms/TextInput/TextInput.tsx
werdnanoslen: yeah i also considered this, and my logic was just that this is a feature of text fields, but the points you make are probably more important and probably wouldn't cause too much extra thought on behalf of the user. Lemme quickly refactor this and submit one more commit for review

shkeating (approved):

werdnanoslen:
@gidjin once this is merged, could we please do a minor release to make this available for my current client?

src/components/forms/TextInput/TextInput.tsx
brandonlenz: Ah, yea sometimes we crib from USWDS, that's fair. In those instances I'm happy to just let it be. Please do call that out for me and I'll forgo being nitpicky. We're a wrapper for their components after all. I believe in some place we have comments pointing to the source USWDS lines if we straight up copy paste their code, which is also fine by me.

src/components/forms/TextInput/TextInputMask.tsx
brandonlenz, line 1: Awesome. Thanks for making this its own component. It's huge that this PR doesn't affect the TextInput component at all

brandonlenz (approved):
Approving as not to block. 

However, I would like to see a reorg (can be a follow on PR if releasing this is priority), where the folder structure looks like:

```
src/
  components/
    forms/
      TextInput/
        {...all our existing stuff}
      TextInputMask/
        TextInputMask.tsx
        TextInputMask.test.tsx
        TextInputMask.stories.tsx
```

Basically just to make sure that the tests and storybook examples are also separated out

src/components/forms/TextInput/TextInputMask.tsx
```diff
@@ -0,0 +1,100 @@
...
+    setIValue(<i>{newValue}</i>)
+    setInputValue(newValue)
+    inputProps.onChange
+  }
+
+  return (
+    <span className="usa-input-mask">
+      <span
+        className="usa-input-mask--content"
+        aria-hidden
```
brandonlenz, line 77: Apparently this needs a value and is not Boolean, but "Booleanish" https://stackoverflow.com/a/55061234/7700346

```suggestion
        aria-hidden="true"
```

src/components/forms/TextInput/TextInputMask.tsx
```diff
@@ -0,0 +1,100 @@
...
+    },
+    className
+  )
+
+  const [inputValue, setInputValue] = useState('')
+  const [maskValue, setMaskValue] = useState(mask)
+  const [iValue, setIValue] =
+    useState<
+      React.DetailedHTMLProps<React.HTMLAttributes<HTMLElement>, HTMLElement>
+    >()
```
brandonlenz, line 37: How about we just make this a string:

```suggestion
  const [iValue, setIValue] = useState('')
```

src/components/forms/TextInput/TextInputMask.tsx
```diff
@@ -0,0 +1,100 @@
...
+        (!charset && !isInt && matchesNumber) ||
+        (charset && ((matchesLetter && !isLet) || (matchesNumber && !isLet)))
+      ) {
+        break
+      } else {
+        newValue += maskData[m]
+      }
+    }
+    setMaskValue(mask.substring(newValue.length))
+    setIValue(<i>{newValue}</i>)
```
brandonlenz, line 68:
```suggestion
    setIValue(newValue)
```

src/components/forms/TextInput/TextInputMask.tsx
```diff
@@ -0,0 +1,100 @@
...
+    inputProps.onChange
+  }
+
+  return (
+    <span className="usa-input-mask">
+      <span
+        className="usa-input-mask--content"
+        aria-hidden
+        data-testid={`${id}Mask`}>
+        {iValue}
```
brandonlenz, line 79:
```suggestion
        {iValue && <i>{iValue}</i>}
```

src/components/forms/TextInput/TextInputMask.tsx
brandonlenz: A change required if making the state a string

src/components/forms/TextInput/TextInputMask.tsx
brandonlenz: required if making the state a string, will only render the `<i>` tags if there is a value to render (similar to USWDS)

src/components/forms/TextInput/TextInputMask.tsx
brandonlenz: Hmmm on closer inspection, this might be my IDE alerting me incorrectly. It looks like it's treating JSX as html. Since `type Booleanish = boolean | 'true' | 'false';`, this shorthand should work.

brandonlenz: Resolving, as I believe this is on my end. Sorry for the confusion!

brandonlenz: Confirmed, issue on my machine, your code works perfectly:
image

src/components/forms/TextInput/TextInput.tsx
werdnanoslen: yeah idk why i never tried this route, was adhering too closely to the original JS implementation...

brandonlenz (approved):

src/components/forms/TextInputMask/TextInputMask.tsx
```diff
@@ -0,0 +1,97 @@
...
+    inputProps.onChange
+  }
+
+  return (
+    <span className="usa-input-mask">
+      <span
+        className="usa-input-mask--content"
+        aria-hidden
+        data-testid={`${id}Mask`}>
+        <i>{iValue}</i>
```
brandonlenz, line 76: I'll reiterate that if you want the `<i>` tag to not show up until something has been typed (like the USDWS examples) you can do this

```suggestion
        {iValue && <i>{iValue}</i>}
```

werdnanoslen: yup, but i didn't see any reason to 🤷

PR #2539, opened 2023-08-01
feat: Remove peer dep on USWDS

lpsinger:
The styles, images, and fonts may be provided by a different package --- for instance, https://github.com/bruffridge/nasawds.

# Summary

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,6 @@
   },
   "homepage": "https://github.com/trussworks/react-uswds#readme",
   "peerDependencies": {
-    "@uswds/uswds": "3.4.1",
     "react": "^16.x || ^17.x || ^18.x",
     "react-dom": "^16.x || ^17.x || ^18.x"
   },


brandonlenz:
We're giving some thought internally to how best to handle custom -WDS implementations. 

Currently, the preferred method (though we've had some trouble in practice, could use help with that) would be to do the following:

1. Project uses ReactUSWDS
2. Project also wishes to use a custom USWDS implementation (A custom USWDS implementation is strictly USWDS with specific theme/settings overrides)
3. Project sets up ReactUSWDS, importing/using the custom UWDS implementation settings when composing their sass theme/settings

werdnanoslen:
Hi @lpsinger can you describe the problems having this as a peer dependency is causing? We're trying to think through different scenarios for other consumers where removing it as a peer dep may be problematic.

lpsinger:
> Hi @lpsinger can you describe the problems having this as a peer dependency is causing?

I can't think of any. The dependency is just unused.

werdnanoslen:
> > Hi @lpsinger can you describe the problems having this as a peer dependency is causing?
> 
> I can't think of any. The dependency is just unused.

Gotcha, we were discussing this today — the reason it's listed as a dependency (peer or not) is to save consumers a step when installing (it is based on uswds after all), and specifically as a peerDependency so that react-uswds works as expected in case a different version of the [core] uswds is also installed.

If this peerDep isn't causing problems, we lean toward keeping it in package.json, but if you can describe any issues (e.g. with performance) we can work on optimizing it :)

lpsinger:
> > > Hi @lpsinger can you describe the problems having this as a peer dependency is causing?
> > 
> > 
> > I can't think of any. The dependency is just unused.
> 
> Gotcha, we were discussing this today — the reason it's listed as a dependency (peer or not) is to save consumers a step when installing (it is based on uswds after all), and specifically as a peerDependency so that react-uswds works as expected in case a different version of the [core] uswds is also installed.

I don't recall whether there was a real problem that it was causing me when I opened this issue. I agree, it's not a big deal.

PR #2552, opened 2023-08-09
fix: Allows semibold to accomodate our typogrpahy story

werdnanoslen:
# Summary

Our typography story mentions 'semibold', but uswds [marked that weight as "unneeded"](https://github.com/uswds/uswds/blob/develop/packages/uswds-core/src/styles/settings/_settings-typography.scss#L333) so it appeared as normal text:
image

This PR overrides that to allow our story to look like it's intended:
image

Alternatively, we can undo this change and instead replace the "semibold" example with a "light" font example, which _is_ allowed by the base uswds package.

## Related Issues or PRs

None, just a little bug noticed by a client

## How To Test

Open [typography story](http://localhost:9009/?path=/story/components-typography-type-styles--type-styles), see that "text semibold" is now actually semibold

### Screenshots (optional)
see above

### DIFF
--- a/src/styles/_uswds-theme.scss
+++ b/src/styles/_uswds-theme.scss
@@ -2,5 +2,6 @@
 @use 'uswds-core' with (
   $theme-image-path: '@uswds/uswds/img',
   $theme-font-path: '@uswds/uswds/fonts',
-  $theme-show-notifications: false
+  $theme-show-notifications: false,
+  $theme-font-weight-semibold: 600
 );
--- a/src/styles/_uswds-theme.scss
+++ b/src/styles/_uswds-theme.scss
@@ -2,6 +2,6 @@
 @use 'uswds-core' with (
   $theme-image-path: '@uswds/uswds/img',
   $theme-font-path: '@uswds/uswds/fonts',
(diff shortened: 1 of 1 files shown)


brandonlenz (commented):
Since we wrap USWDS, generally we should follow all their recommendations and conventions. 

I think we should mirror them in this case. Do you need semibold for client work specifically?

werdnanoslen:
> Do you need semibold for client work specifically?

I do, so in their case, importing our react package, how would they override that?

werdnanoslen:
@brandonlenz ☝️

brandonlenz:
> I do, so in their case, importing our react package, how would they override that?

@werdnanoslen Your project should have a theme.scss file where uswds core settings are overridden, similar to what you're editing here. Does adding the semibold setting there work?

werdnanoslen:
> > I do, so in their case, importing our react package, how would they override that?
> 
> @werdnanoslen Your project should have a theme.scss file where uswds core settings are overridden, similar to what you're editing here. Does adding the semibold setting there work?

ah great then, we'll just add it there

PR #2553, opened 2023-08-10
fix: moved overrides to let us use utilities

werdnanoslen:
# Summary

Thanks to Joe, this establishes a pattern for us to replicate in using utilities, at least in the root styles folder

## Related Issues or PRs

None

## How To Test

Add any future override styles to the new _overrides.scss file


kleinschmidtj (approved):
LGTM

PR #2562, opened 2023-08-16
refactor: Introduce 'id' prop on <Tooltip>

ClayBenson94:
# Summary

BREAKING CHANGE: This commit introduces a required 'id' property on the <Tooltip>
component which replaces the random automatically generated one used previously.

Our application creates snapshot tests using `jest-snapshot`, and the randomly-generated IDs that come from this component make the snapshots no longer deterministic (and would fail on each run).

This PR is an attempt to address that. I can't say I'm in love with this solution as it does introduce a breaking change that would require IDs to be placed on `<Tooltip>` components across the board, so I am open to other solutions as well!

### DIFF
--- a/src/components/Tooltip/Tooltip.stories.tsx
+++ b/src/components/Tooltip/Tooltip.stories.tsx
@@ -33,13 +33,13 @@ Source: https://designsystem.digital.gov/components/tooltip/
 
 export const tooltipDefault = (): React.ReactElement => (
   <div className="margin-4">
-    <Tooltip label="Default">Default</Tooltip>
+    <Tooltip id="tooltip-default" label="Default">Default</Tooltip>
   </div>
 )
 
 export const tooltipTop = (): React.ReactElement => (
   <div className="margin-4">
-    <Tooltip position="top" label="Top">
+    <Tooltip id="tooltip-top" position="top" label="Top">
       Show on top
     </Tooltip>
   </div>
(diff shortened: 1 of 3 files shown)


haworku:
@ClayBenson94 Hey - that use case makes a lot sense to me! 

In terms of implementation - can we surface the `id?: string`  as an optional prop that falls back to the randomly generated id.  That way we avoid the breaking change but still have always have an id that can be accessed internally for tooltip

werdnanoslen:
@ClayBenson94 do you have some time to make the change Hana suggested? ☝️

shkeating (approved):
lookin good from design standpoint

mcmcgrath13:
This `id` randomness also causes issues for server side rendering. We could also use `React.useId()` instead of `useRef`, which is easy to mock for testing if needed and doesn't require an API change

brandonlenz:
@ClayBenson94 please take a look at @mcmcgrath13's solution in https://github.com/trussworks/react-uswds/pull/3094.

That solution would be a non-breaking change, which makes it easier to include.

And if you would still like to see an optional Id prop added, I think that would be a fine addition as well, but would recommend including it on https://github.com/trussworks/react-uswds/pull/3094

brandonlenz:
Based on https://github.com/trussworks/react-uswds/pull/3096, it sounds like adding an _optional_ id prop would be the next improvement to make. 

We can update this PR to include an optional id prop OR close this PR and create an issue for the optional prop

PR #2581, opened 2023-09-06
fix: text input mask fixes

nicholasguyett:
# Summary

Fixes the incorrect pass-through of the `inputRef` prop

Adds support for `defaultValue`

Adds support for using `TextInputMask` as a controlled component

Fixes support for externally set `onChange` hooks

Note: The `maskString` function contains the unmodified original masking implementation, just moved into a separate function so it can be called in both `onChange` and when setting the initial state.

## How To Test

Manually tested this change in an external (currently private) application that uses both `defaultValue` and `onChange`.


rogeruiz:
Thanks for the PR @nicholasguyett. I'll be taking a look at this and we will be incorporating in our next release. I'm currently looking through how to test this locally and appreciate your summary.

src/components/forms/TextInputMask/TextInputMask.tsx
```diff
@@ -29,42 +57,31 @@ export const TextInputMask = ({
...
+  useEffect(() => {
+    // Make sure this component behaves correctly when used as a controlled component
+    setValue(
+      maskString(
+        ((externalValue ?? defaultValue) as string) ?? ``,
+        mask,
+        charset
+      )
+    )
+  }, [externalValue])
```
rogeruiz, line 73: Great work on this and using `useEffect` API hook.

src/components/forms/TextInputMask/TextInputMask.tsx
```diff
@@ -29,42 +57,31 @@ export const TextInputMask = ({
...
+    const newValue = maskString(e.target.value, mask, charset)
+
     setMaskValue(mask.substring(newValue.length))
-    setIValue(newValue)
-    setInputValue(newValue)
-    inputProps.onChange
+    setValue(newValue)
+
+    // Ensure the new value is available to upstream onChange listeners
+    e.target.value = newValue
```
rogeruiz, line 82: Good catch here on ensuring other `onChange` listeners get the value here too.

rogeruiz (approved):
lgtm 🌈

I was able to test and build the project locally. Also, based on the _How To Test_ section this looks like was already tested in a private repository.

brandonlenz:
@all-contributors please add @nicholasguyett for bug, code

PR #2596, opened 2023-09-27
feat: Allow Tooltip label to be a string or React component

lpsinger:
# Summary

Allow tooltip labels to be a string or any React component.


src/components/Tooltip/Tooltip.tsx
```diff
@@ -232,7 +232,7 @@ export function Tooltip<FCProps = DefaultTooltipProps>(
         {triggerElement}
         <span
           data-testid="tooltipBody"
-          title={label}
+          title={typeof label === 'string' ? label : undefined}
```
brandonlenz, line 235: @lpsinger Instead of silently assigning `label` to `undefined` without a way to override, what do you think about:

1. Add an optional `title` prop which is a `string`
2. If the `title` prop is passed in/has a value, that value is assigned to the span's `title`.
3. If `title` is not passed in, then use `label` if it is a string

lpsinger: Sure thing. Done.

brandonlenz (approved):
Thanks for making that change! I believe this keeps the change backwards compatible as well, so this _won't_ need to involve making our next release a breaking release 👍🏻

PR #2597, opened 2023-09-27
chore: add admin group and design owners group

gidjin:
# Summary

This PR adds the `react-uswds-admin` github team as code owners to make that easier to manager. Also adds the `react-uswds-design-owners` team as owners of the storybook files. This is to facilitate having a designer provide input on changes that affect the look and feel of a component.


gidjin:
> @gidjin I had to update the @trussworks/react-uswds-design-owners role to have write access to the repository, otherwise they are not considered valid codeowners. All should be good now!

Thank you for that I didn't realize that was missing.

PR #2622, opened 2023-10-06
fix: Language selector: add prop for display lang for bug fix

### LINKED ISSUE #2621: [fix] LanguageSelector toggle broken when locale is persistent

**Describe the bug**

The current language selector component assumes that the displayed language should always the first item in the array upon load. However, in settings where the [i18n locale is saved](https://github.com/i18next/i18next-browser-languageDetector) (insert link to browser detector) or static sites where the locale is in the URL, this should not be the case.

For example, in the first case if you select the language toggle, the language is changed and saved to the browser session or cookie. This component will still assume the first language in the array is the one to display, so never updating the displayed language to change. 

Also, in static sites where the locale comes from the URL, the same behavior occurs. Even though the language toggle may go to the alternate language for the first click, the displayed language would not update.

**To Reproduce**

1. Go to a static site or SPA using a method of saving the locale in the browser.
2. Click on the language selector.
3. Refresh the page.
4. See that the content of the page may be translated, but the displayed label of the alternate language is not updated in the component.

**Expected behavior**

The component should accept an additional prop with the current i18n language locale in order to determine the alternate language to display. 

**Side quest**: The component in its dropdown form use this new prop to actually display the selected language instead of the label prop or the first item in the languages array. 

**Screenshots**

**Additional context**

kleinschmidtj:
# Summary

- Added a new optional prop to handle cases where the current language is persisted and the component should know which language to display. Defaults to setting the displayed language as the first item in the `langs` array.
- Broken out the drop down part of the selector into a separate file.
- Dropdown prefers to set the label with the new prop's label.

## Related Issues or PRs

closes #2621 

## How To Test

Use the new prop and confirm it overrides the existing behavior where the label is pulled from the first item in the `langs` array.

### Screenshots (optional)
Screenshot 2023-10-06 at 6 36 05 AM

Screenshot 2023-10-06 at 6 36 26 AM


werdnanoslen (commented):
I can't get the 2-lang variant to toggle (tested on Mac Firefox)
Oct-06-2023 11-46-16

kleinschmidtj:
@werdnanoslen Yeah, these stories are misleading. That prop would be used with a saved i18n locale. So instead of passing a static value as in the story, it would use `i18n.language` in conjunction with `onClick()` handlers that would use `i18n.changeLanguage()`.

I'll make sure to write unit tests with this in mind, but maybe we consider getting rid of the stories.

brandonlenz:
> @werdnanoslen Yeah, these stories are misleading. That prop would be used with a saved i18n locale. So instead of passing a static value as in the story, it would use `i18n.language` in conjunction with `onClick()` handlers that would use `i18n.changeLanguage()`.
> 
> I'll make sure to write unit tests with this in mind, but maybe we consider getting rid of the stories.

@kleinschmidtj if I'm understanding correctly, it should be possible to give your storybook example a useState (see Pagination.stories.tsx as an example), where the react state can serve as a mock for "saving and pulling the value from somewhere"

kleinschmidtj:
> > @werdnanoslen Yeah, these stories are misleading. That prop would be used with a saved i18n locale. So instead of passing a static value as in the story, it would use `i18n.language` in conjunction with `onClick()` handlers that would use `i18n.changeLanguage()`.
> > I'll make sure to write unit tests with this in mind, but maybe we consider getting rid of the stories.
> 
> @kleinschmidtj if I'm understanding correctly, it should be possible to give your storybook example a useState (see Pagination.stories.tsx as an example), where the react state can serve as a mock for "saving and pulling the value from somewhere"

Oh excellent @brandonlenz ! I have never done that and wasn't sure that was something I could do. I will look there. Thank you.

kleinschmidtj:
@brandonlenz @werdnanoslen Created two templates to test this out:

http://localhost:9009/?path=/story/components-languageselector--languages-display-prop-sandbox

http://localhost:9009/?path=/story/components-languageselector--languages-display-more-than-two-languages-prop-sandbox

Thanks to you both!

werdnanoslen (approved):
Approving conceptually but I haven't reviewed the code. Good addition!

brandonlenz (approved):

brandonlenz:
@kleinschmidtj Nice! 🚢

PR #2629, opened 2023-10-15
fix!: Label is mandatory for TimePicker component

### LINKED ISSUE #2611: [fix] time picker component should always have a label
**ReactUSWDS Version: 5.4.0 & USWDS Version: 3.6.0**

# Describe the bug

In our storybook examples of the Combo box component, we do not always include a label. 

In the USWDS docs about the time picker, it is explicitly called out that we must **always use a label** on this component. We should not include examples where it is unlabelled in our project storybook. The following examples lack a label:

- [Default Time Picker](https://trussworks.github.io/react-uswds/?path=/story/components-form-controls-time-picker--default-time-picker)

We could simply remove the example that shows the time picker with no label from storybook, or in this case, it may be helpful to set the label prop in this component to be required to help guide react-uswds users to ensure they don't forget it in their implementations. I could be convinced of either approach, as both have pros and cons. This could be potentially cumbersome to projects already using react-uswds that have already added labels of their own who later upgrade versions.

# To Reproduce

1. Go to [Default Time Picker](https://trussworks.github.io/react-uswds/?path=/story/components-form-controls-time-picker--default-time-picker) 
2. See there is no label

# Expected behavior

All visible versions of the time picker in our storybook should have labels to prevent confusion that they are optional, when they are mandatory for this component. 

# Additional context

- [USWDS Time Picker documentation](https://designsystem.digital.gov/components/time-picker/)
- [Labelling Controls - W3C](https://www.w3.org/WAI/tutorials/forms/labels/)
- file(s) to be edited: `src/components/forms/TimePicker/TimePicker.tsx` and/or `src/components/forms/TimePicker/TimePicker.stories.tsx`

# Screenshots

image

m7adeel:
# Summary
Resolved mandatory label prop requirement for the time picker component, enhancing compliance with USWDS documentation.

## Related Issues or PRs
Closes #2611

## How To Test
The developer can try to add a TimePicker component without a label prop which would result in a typescript error with a msg of (Property 'label' is missing)

### Screenshots 
Screenshot from 2023-10-15 08-29-03
Screenshot from 2023-10-15 08-27-25
Screenshot from 2023-10-15 08-27-05
Screenshot from 2023-10-15 08-26-54


m7adeel:
Hello @shkeating, I have worked on this issue can you kindly review it?

brandonlenz:
@m7adeel please update the tests for TimePicker to include a label, now that label is mandatory on this branch tests are failing as a result

brandonlenz:
Updated title to `fix!` as this will be a breaking change, mandating a major version bump on the next release

AnnaGingle (dismissed):
Looks good to me! Great work. I do have some feedback on the PR instructions, particularly the how to test section. I humbly suggest adding a sentence or two about how the PR relates to the design of the component. It's a great way to get new design contributors in the door.

src/components/forms/TimePicker/TimePicker.stories.tsx
```diff
@@ -54,6 +54,7 @@ export const defaultTimePicker = (
     <TimePicker
       id="appointment-time"
       name="appointment-time"
+      label="Appointment Time"
```
brandonlenz, line 57: USWDS' example uses sentence case for this label, we should match their example:

```suggestion
      label="Appointment time"
```

AnnaGingle (dismissed):
LGTM

AnnaGingle:
I did notice the time picker component label is using Start Case where it should be using sentence case.
-This Sentence Is Called A Start Case.
-This sentence is called a sentence case.
The [USWDS website](https://designsystem.digital.gov/components/time-picker/#component-code) uses sentence case & it’s better for accessibility.

brandonlenz:
Noticing the lockfiles got corrupted on this branch. Gonna try to resolve to move this along

brandonlenz:
> Noticing the lockfiles got corrupted on this branch. Gonna try to resolve to move this along

Ah right, I'd have to make a fork of your fork for those updates to display in github.

@m7adeel can you resolve the issues with the lock files? Looks like maybe an npm command was issued? 

My recommendation:
1. Delete package.lock.json (we use yarn on this project)
2. Revert any changes to yarn.lock. There should be no diff here as package.json was not changed. This PR includes code changes only

m7adeel:
@brandonlenz I have fixed the issues

brandonlenz (dismissed):

AnnaGingle (changes requested):
I did notice the time picker component label is using Start Case where it should be using sentence case.
This Sentence Is Called A Start Case.
This sentence is called a sentence case.
The [USWDS website](https://designsystem.digital.gov/components/time-picker/#component-code) uses sentence case & it’s better for accessibility.

brandonlenz:
> I did notice the time picker component label is using Start Case where it should be using sentence case. This Sentence Is Called A Start Case. This sentence is called a sentence case. The [USWDS website](https://designsystem.digital.gov/components/time-picker/#component-code) uses sentence case & it’s better for accessibility.

@AnnaGingle I'm seeing this PR updated to sentence case. Can you confirm?

brandonlenz (approved):

AnnaGingle (approved):
Looks good to me!

nmcconlogueOG:
I feel like this fix goes in the wrong direction, a component library should make it easier for the developers to implement the spec correctly - not enforce it at the api level.   Making label mandatory on TimePicker creates an inconsistency with DatePicker, which leaves label ownership to the consumer. This matters when using TimePicker inside a form library like RJSF with custom field templates —   the template owns the label rendering and positions it alongside hint text in the correct DOM order (label -> hint ->input). With a required label prop, TimePicker renders its own label internally, which you can't opt out of, so you end up with duplicate labels unless you pass an empty string — which is a hack that breaks accessibility. DatePicker avoids this entirely by not having a label prop, letting the consumer render it  externally. TimePicker should follow the same pattern, or at minimum make label optional so consumers can  suppress it when their layout owns the labeling.

PR #2636, opened 2023-10-22
fix: adding labels in combo box (default combo box, with default value)

### LINKED ISSUE #2608: [fix] combo box form element must always have a label
**ReactUSWDS Version: 5.4.0 & USWDS Version: 3.6.0**

# Describe the bug

In our storybook examples of the Combo box component, we do not always include a label. 

In the USWDS docs about the combo box, it is explicitly called out that we must always use a label on this component. We should not include examples where it is unlabelled in our project storybook. The following examples lack a label:

- [Default Combo box](https://trussworks.github.io/react-uswds/?path=/story/components-combo-box--default-combo-box-with-prop-options)
- [With Default Value](https://trussworks.github.io/react-uswds/?path=/story/components-combo-box--with-default-value)

Potentially, we should hard code a label right into the component file, and have that label's content be added as a prop to help guide react-uswds users to ensure they don't forget it in their implementations. I could be convinced of either approach, as both have pros and cons. This could be potentially cumbersome to projects already using react-uswds that have already added labels of their own who later upgrade versions.

It is also worth flagging though, that the combo box has other known accessibility issues that the USWDS team is aware of: 

> In late 2022, we found some combo box [usability issues](https://designsystem.digital.gov/components/combo-box/#accessibility-combo-box) when we were testing with people who use screen readers and alternative input devices. For more information, see [the issue on GitHub](https://github.com/uswds/uswds-site/issues/1898).

# To Reproduce

1. Go to [Default Combo box](https://trussworks.github.io/react-uswds/?path=/story/components-combo-box--default-combo-box-with-prop-options) and [With Default Value](https://trussworks.github.io/react-uswds/?path=/story/components-combo-box--with-default-value)
2. See there is no label

# Expected behavior

All visible versions of the combo box in our storybook should have labels to prevent confusion that they are optional, when they are mandatory for this component. 

# Additional context

- [USWDS Combo Box documentation](https://designsystem.digital.gov/components/combo-box/)
- [Labelling Controls - W3C](https://www.w3.org/WAI/tutorials/forms/labels/)
- file(s) to be edited: `src/components/forms/ComboBox/ComboBox.stories.tsx`

# Screenshots

image

m7adeel:
# Summary
According to the USWDS docs the combo box should always have a label. Some of the examples in the storybook do not have labels. So the commit adds labels in the component file `src/components/forms/ComboBox/ComboBox.stories.tsx` to avoid confusion that they are optional.

## Related Issues or PRs
Closes #2608

## How To Test

1. Go to Default Combo box and With Default Value
2. Users can see that there is a label added to them

### Screenshots (optional)

default
with_default_value


rileyorr (dismissed):
Great job on this - LGTM! 

The only thing I might consider either as part of this PR or a separate one is potentially removing the `With Label` use case since the default combo box already includes a label. It might help clean up the documentation; however, it might impact teams already using that component.

src/components/forms/ComboBox/ComboBox.stories.tsx
```diff
@@ -72,7 +74,7 @@ export const withLabel = (): React.ReactElement => {
 
   return (
     <Form onSubmit={noop}>
-      <Label htmlFor="fruit">Select a fruit</Label>
+      <Label htmlFor="fruit">Select a Fruit</Label>
```
brandonlenz, line 77: The labels should be sentence case ([see USWDS example](https://designsystem.digital.gov/components/combo-box/)):

```suggestion
      <Label htmlFor="fruit">Select a fruit</Label>
```

brandonlenz: (Same for the others added on this PR

m7adeel:
Hello @brandonlenz, Can you kindly review my PR

brandonlenz (approved):

PR #2500, opened 2023-07-17
fix: Set alternative text on SVG img icon in menu close button

sawyerh:
# Summary

`<svg>` elements with an `img` role must have an alternative text: https://dequeuniversity.com/rules/axe/4.6/svg-img-alt?application=msftAI

## Related Issues or PRs

This issue was introduced in https://github.com/trussworks/react-uswds/pull/2411, and began failing a11y tests in our project when 5.0.0 was released.

## How To Test

1. Open the `NavCloseButton` Storybook page
2. Change your viewport size so the button shows
3. Run an accessibility scanner like https://accessibilityinsights.io/

### Screenshots

Before:

CleanShot 2023-07-17 at 10 32 46@2x

After:

CleanShot 2023-07-17 at 10 34 25@2x

PR #2640, opened 2023-10-24
fix(a11y): move aria-label from svg icon to button

### LINKED ISSUE #2639: [fix] nav buttons ahould always have aria-label attributes

* ReactUSWDS: 5.5.0
* USWDS: 3.6.0

**Describe the bug**

[WCAG § 4.1.2](https://www.w3.org/TR/WCAG21/#name-role-value) states the following:

> For all [user interface components](https://www.w3.org/TR/WCAG21/#dfn-user-interface-components) (including but not limited to: form elements, links and components generated by scripts), the [name](https://www.w3.org/TR/WCAG21/#dfn-name) and [role](https://www.w3.org/TR/WCAG21/#dfn-role) can be [programmatically determined](https://www.w3.org/TR/WCAG21/#dfn-programmatically-determinable); states, properties, and values that can be set by the user can be [programmatically set](https://www.w3.org/TR/WCAG21/#dfn-programmatically-set); and notification of changes to these items is available to [user agents](https://www.w3.org/TR/WCAG21/#dfn-user-agents), including [assistive technologies](https://www.w3.org/TR/WCAG21/#dfn-assistive-technologies).

The `NavCloseButton` element, while it accepts an `aria-label` itself, by default only applies an `aria-label` to the enclosed `Icon.Close` rather than the `<button />` element. This means that using any further abstracted components (like `ExtendedNav`) will cause the resulting HTML to violate this rule.

The buttons within the modal elements already conform to this paradigm, only the nav elements appear to be effected.

**To Reproduce**

1. Go to the project storybook
2. Click on Components -> Header
3. Inspect any `*Button` element
4. Observe that `aria-label` is not present on button elements, but is present on child `<svg />` elements

**Expected behavior**

`<button />` elements should have `aria-label` attribute to meet WCAG § 4.1.2 requirements.

**Additional context**

This issue was initially identified using [HTML_Codesniffer](https://github.com/squizlabs/HTML_CodeSniffer) which we use alongside Axe and playwright to implement E2E testing.

seevee:
# Summary

## Related Issues or PRs

Closes #2639 

## How To Test

I tested with `yarn link` to an application using playwright, AXE, and HTML_CodeSniffer. I am yet not familiar enough with these packages to set up an example repo, but I can confirm that these changes address our E2E testing issues.

### DIFF
--- a/src/components/header/NavCloseButton/NavCloseButton.tsx
+++ b/src/components/header/NavCloseButton/NavCloseButton.tsx
@@ -18,7 +18,7 @@ export const NavCloseButton = ({
       aria-label="Close Navigation Menu"
       {...buttonProps}
       type="button">
-      <Icon.Close size={3} aria-hidden="true" />
+      <Icon.Close size={3} aria-hidden="true" alt="Close Icon" />
     </button>
   )
 }
--- a/src/components/header/NavCloseButton/NavCloseButton.tsx
+++ b/src/components/header/NavCloseButton/NavCloseButton.tsx
@@ -18,7 +18,7 @@ export const NavCloseButton = ({
       aria-label="Close Navigation Menu"
       {...buttonProps}
       type="button">
-      <Icon.Close size={3} aria-hidden="true" alt="Close Icon" />
+      <Icon.Close size={3} aria-hidden="true" />
     </button>
   )
 }
--- a/src/components/header/NavCloseButton/NavCloseButton.tsx
+++ b/src/components/header/NavCloseButton/NavCloseButton.tsx
@@ -15,9 +15,10 @@ export const NavCloseButton = ({
       className={classes}
       onClick={onClick}
       data-testid="navCloseButton"
+      aria-label="Close"
       {...buttonProps}
       type="button">
-      <Icon.Close size={3} aria-label="Close" />
+      <Icon.Close size={3} aria-hidden="true" />
     </button>
   )
 }
--- a/src/components/header/NavCloseButton/NavCloseButton.tsx
+++ b/src/components/header/NavCloseButton/NavCloseButton.tsx
@@ -15,7 +15,7 @@ export const NavCloseButton = ({
       className={classes}
       onClick={onClick}
       data-testid="navCloseButton"
-      aria-label="Close"
+      aria-label="Close Navigation Menu"
       {...buttonProps}
       type="button">
       <Icon.Close size={3} aria-hidden="true" />


brandonlenz:
Thanks for the PR!

This is conflicting with the last change to this component in https://github.com/trussworks/react-uswds/pull/2500

@sawyerh do you have thoughts on this change since you authored the last iteration?

sawyerh:
@brandonlenz Hey stranger 😄  No issues from my side. I think my original solution was based on the USWDS example code which uses an `alt` on their `img`, but this solution seems like it should be fine too.

CleanShot 2023-10-25 at 15 16 29@2x

seevee:
Some additional context - these changes will bring the NavCloseButton default behavior closer to parity with  ModalCloseButton:
https://github.com/trussworks/react-uswds/blob/59dffd138f745f38a222523360a4666c0c13b030/src/components/Modal/ModalCloseButton/ModalCloseButton.tsx#L14-L24

The aria-label being present on the svg within the button is probably not hugely different from applying it directly to the button, except that our a11y tests can't seem to drill down into the svg as content to satisfy the button requirements. See this error message when using the ExtendedNav default:

```
WCAG2AA.Principle4.Guideline4_1.4_1_2.H91.Button.Name

This button element does not have a name available to an accessibility API. Valid names are: title undefined, element content, aria-label undefined, aria-labelledby undefined.
```

brandonlenz:
> EDIT: would it be acceptable to add an `alt="Close"` to the icon properties alongside the button's `aria-label="Close"`? That seems like it should let us satisfy both the svg-img-alt rule from https://accessibilityinsights.com as well as the Button.Name rule from our AXE/Playwright setup.

@seevee I _think_ that sounds fine. @shkeating this one is pretty accessibility focused. Do you have any thoughts about the changes or suggestion?

src/components/header/NavCloseButton/NavCloseButton.tsx
```diff
@@ -15,9 +15,10 @@ export const NavCloseButton = ({
       className={classes}
       onClick={onClick}
       data-testid="navCloseButton"
+      aria-label="Close"
       {...buttonProps}
       type="button">
-      <Icon.Close size={3} aria-label="Close" />
```
shkeating, line 20: this is an acceptable implementation to me, my ANDI output (which shows the text data that is received by screen readers visually) shows the close button labelled as 'Close' as intended. 

image


However, I would encourage you to get more specific in the accessible name. Let's say we had another component in a page layout that also had a close icon. If both are named "close" they would be indistinguishable for a screen reader user navigating by links on a page.

the aria label is correct but I would change it to "close menu" or "close navigation" to avoid duplicate accessible names. If there was two components using close buttons under this pattern in the same page layout, we would see this ANDI error: [⚠ Non-unique button: same name/description as another button](https://www.ssa.gov/accessibility/andi/help/alerts.html?non_unique_button).

shkeating (changes requested):

src/components/header/NavCloseButton/NavCloseButton.tsx
brandonlenz: What about enabling the prop to be passed in? 

Whatever default value we provide here, whether "close" or "close menu" might be nice to have the option to override it. That would require an `iconProps` param passed in where we could then get `iconProps['aria-label']`, and if present, use that instead of the default.

I'm fine with that being an enhancement later on too though.

seevee: @shkeating thanks for the review and the advice! I've updated the `aria-label` to `"Close Navigation Menu"` to make it more unique and descriptive.

@brandonlenz Allowing the label to be passed in as a prop sounds like the ideal path to me! I think this would be especially useful if we want to show localized aria labels.

I think there's a couple of other components that could use that enhancement as well, so I'll keep the diff on this PR as lean as possible for now and open a new PR targeted at using those props as overrides. Let me know if that works!

brandonlenz: That works for me.

brandonlenz (dismissed):

brandonlenz (dismissed):

seevee:
@brandonlenz apologies for the churn on this, looks like we can't add an `alt` attribute to an svg.

brandonlenz:
@seevee no worries! Thanks for seeing it through

brandonlenz:
I think we'll need @shkeating to give it another approval anyways since she requested changes

shkeating (approved):

shkeating:
all set! great work all!

PR #2642, opened 2023-10-24
fix: ExtendedNav Tab Order

### LINKED ISSUE #2641: [fix]: ExtendedNav Tab Order

ReactUSWDS: 5.5.0
USWDS: 3.6.0

**Describe the bug**

When adding secondary elements to the ExtendedNav component the secondary components will be placed after the primary components in the DOM and re-positioned above the primary components with CSS. This means that the DOM order does not match the visual order

[C27: Making the DOM order match the visual order](https://www.w3.org/TR/WCAG20-TECHS/C27.html)
> The objective of this technique is to ensure that the order of content in the source code is the same as the visual presentation of the content. [...] Each order may be meaningful in itself but may cause confusion for assistive technology users.

Which can cause problems for users who use assistive technologies and should be able to Tab through the focusable elements in the same order as they appear visually.

**To Reproduce**

1. Go to the [Extended Header](https://trussworks.github.io/react-uswds/?path=/story/components-header--extended-header) component in storybook
2. Press Tab
3. Error: The first element that is focused is the `Nav Label`

**Expected behavior**
The first element focused should be `Simple link one` above the search bar

**Screenshots**

https://github.com/trussworks/react-uswds/assets/8581332/a3a707a6-d77e-49db-81a3-a8fae491e9c8

**Additional context**

There is a similarly related bug ticket in the USWDS repo https://github.com/uswds/uswds/issues/3920

- Chrome

bbemis017:
# Summary

This PR re-orders the primary/secondary elements in the ExtendedNav so that it matches the visual order that a user sees. When a user uses Tabs through the component,  they will iterate through elements in the same order that they appear

## Related Issues or PRs

closes #2641 

## How To Test

Open Storybook locally and go to the Extended Header Page
Press Tab and confirm the first element focused is `Simple line one`
Then followed by
`Simple link two`
The search text box
The search button
Nav Label

### Screenshots (optional)

https://github.com/trussworks/react-uswds/assets/8581332/e21f8bf4-be75-4605-bb8a-d9e6db4271f2


brandonlenz:
Thanks for the PR! I'll wonder what the designers think about this. 

We generally try to match whatever USWDS' implementation brings, for better (most of the time), or for worse.

Our markup before this change matches USWDS: https://designsystem.digital.gov/components/header/#extended-header

The [USWDS example](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/iframe.html?id=components-header--extended&args=&viewMode=story) doesn't seem to include the secondary items in the tab order at all from a quick test (firefox), so that's puzzling to me as well

bbemis017:
> The [USWDS example](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/iframe.html?id=components-header--extended&args=&viewMode=story) doesn't seem to include the secondary items in the tab order at all from a quick test (firefox), so that's puzzling to me as well

I'm not sure if I'm seeing the same behavior as you, But when I try it in  firefox it skips over the links that are secondary items but it still focuses the search box and search button. Which are also secondary items in that example.

This is different than the behavior I see in chrome though, So I think theres a firefox setting that changes what is focusable. This [stackoverflow](https://stackoverflow.com/a/11713537) seems related

Issue #2540, opened 2023-08-02
[fix] Issues properly using with the Next.js "App Router" architecture (server components)


sawyerh:
**ReactUSWDS Version & USWDS Version:**

```json
"@trussworks/react-uswds": "5.1.1",
"@uswds/uswds": "3.1.0",
"next": "13.4.12",
"react": "18.2.0",
"react-dom": "18.2.0"
```

**Describe the bug**

When any ReactUSWDS component is imported into a "server side component", the following error is thrown:

```
error Error [TypeError]: Super expression must either be null or a function
    at eval (webpack-internal:///(rsc)/./node_modules/@trussworks/react-uswds/lib/index.js:66:73)
    at eval (webpack-internal:///(rsc)/./node_modules/@trussworks/react-uswds/lib/index.js:76:22)
    at 483 (webpack-internal:///(rsc)/./node_modules/@trussworks/react-uswds/lib/index.js:212:18)
    at r (webpack-internal:///(rsc)/./node_modules/@trussworks/react-uswds/lib/index.js:651:24)
    at eval (webpack-internal:///(rsc)/./node_modules/@trussworks/react-uswds/lib/index.js:9651:22)
    at eval (webpack-internal:///(rsc)/./node_modules/@trussworks/react-uswds/lib/index.js:10333:10)
    at eval (webpack-internal:///(rsc)/./node_modules/@trussworks/react-uswds/lib/index.js:10334:6)
    at eval (webpack-internal:///(rsc)/./node_modules/@trussworks/react-uswds/lib/index.js:3:30)
    at eval (webpack-internal:///(rsc)/./node_modules/@trussworks/react-uswds/lib/index.js:4:2)
    at (rsc)/./node_modules/@trussworks/react-uswds/lib/index.js (/Users/sawyer/Dev/Projects/react-uswds-server-components/.next/server/app/react-uswds/page.js:1559:1) {
  digest: undefined
}
```

**To Reproduce**

I've created a repo with the reproduction here: https://github.com/sawyerh/react-uswds-server-components-bug

1. Install the repo (`npm install`)
2. Run the dev server (`npm run dev`)
3. View `http://localhost:3000/react-uswds` and observe an error
4. View `http://localhost:3000/react-uswds-use-client` and observe a rendered ReactUSWDS component

The difference between 3 and 4 is that 4 includes a `"use client"` directive at the top of its page file, forcing it to be a client-side component, as opposed to server component.

**Expected behavior**

[...]

Issue #192, opened 2020-05-21
[feat] Investigate `pkg.module` and/or rollup for better module support & tree shaking


suzubara:
**Does your feature request relate to a specific USWDS feature?**
No

**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
Investigate using `pkg.module` and whether it is worth changing our build / package config.

**Describe alternatives you've considered**
We can choose to do nothing.

**Additional context**
Port from initial roadmap -- motivation behind this is to optimize the library exports for better performance & modern JS applications.
- https://github.com/rollup/rollup/wiki/pkg.module
- https://stackoverflow.com/questions/41289200/output-an-es-module-using-webpack

Morsey187:
I've had a quick look at using webpack V5 a while ago now, I managed to get it to output an ES6 module however it seems like this feature is very much [experimental](https://github.com/webpack/webpack/issues/2933#issuecomment-774253975) as it opened up a rabbit hole of issues trying to get the ES module to work with my Next.js build.

I'd like to iterate @jkjustjoshing 's queries above, and see if there would be any appetite in using rollup.js as the approach instead? I think either of his options sound fine. Otherwise I could have another look at the webpack changes I made previously and open a PR with some notes with the issues I encountered so others can take a look? The changes required for webpack were pretty minior, but it seems like a concerning route if webpack is likely to cause more issues down the line.

I'd also like to know what the current teams resourcing is like in terms of supporting a PR, or whether this issue is on your roadmap?

This would also be a great issue to solve from an environmental perspective, I'd imagine there are quite a few high traffic sites that could split their JavaScript bundles in half.

lpsinger:
Did you consider esbuild as an alternative to rollup?

jpandersen87:
I have a proof-of-concept of using vite/vitest here: https://github.com/trussworks/react-uswds/pull/2716

PR #2656, opened 2023-11-10
feat: Add “use client” to fix React Server Components

sawyerh:
# Summary

This adds the [`"use client"`](https://react.dev/reference/react/use-client) directive to the transpiled JS module to support apps utilizing [React Server Components](https://nextjs.org/docs/app/building-your-application/rendering/server-components):

Some context from the React docs on this directive:

> When a file marked with 'use client' is imported from a Server Component, [compatible bundlers](https://react.dev/learn/start-a-new-react-project#bleeding-edge-react-frameworks) will treat the module import as a boundary between server-run and client-run code.

> For apps that use React Server Components, the app is server-rendered by default. 'use client' introduces a server-client boundary in the [module dependency tree](https://react.dev/learn/understanding-your-ui-as-a-tree#the-module-dependency-tree), effectively creating a subtree of Client modules.

Without this directive, apps attempting to use React Server Components are greeted with an error screen. See #2540.

I think this PR would be highly impactful to Next.js applications in particular, supporting the adoption of the new [App Router](https://nextjs.org/docs) and newer React features.

## Additional context

I don't think this is the perfect solution to supporting React Server Components, but I think it's a quick fix. Longer term, #192 is likely a better approach, so that components without any React hooks or event listeners _can_ be a server component.

## Related Issues or PRs

- #2540

## How To Test

1. Clone https://github.com/sawyerh/react-uswds-server-components-bug
1. `npm install`
1. `npm run dev`
1. Open http://localhost:3000/react-uswds and observe an error
1. Stop the dev server, then run: `npm link @trussworks/react-uswds --save`
1. Open http://localhost:3000/react-uswds and observe the error is gone

## Screenshots

CleanShot 2023-11-09 at 17 44 42@2x

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,6 +1,8 @@
 const path = require('path')
 const MiniCssExtractPlugin = require('mini-css-extract-plugin')
 const ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin')
+const TerserPlugin = require('terser-webpack-plugin')
+const { BannerPlugin } = require('webpack')
 
 const uswdsIncludePaths = [
   './node_modules/@uswds',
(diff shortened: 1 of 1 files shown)


sawyerh:
@brandonlenz Just checking in to see if this is a change y'all would be open to or if I should pursue an alternative. (Tagging you since you're listed in the readme as the active maintainer).

brandonlenz (approved):
I am good with this change. Looks like a quick win

Issue #2587, opened 2023-09-14
[feat] Export component props


brandonlenz:
**Does your feature request relate to a specific USWDS component?**

All components, we should make it a convention going forward to export their props

**Is your feature request related to a problem? Please describe.**

Sometimes consumers want to get TS access to the props our components accept. It's not always easy/intuitive to do so.

**Describe the solution you'd like**

export all our component prop type definitions, similar to https://github.com/react-bootstrap/react-bootstrap/blob/master/src/index.tsx. 

Note: We union our props with `IntrinsicElements[THE_ROOT_ELEMENT]` pretty often, inline in the component args. We would want those exported with the type, so it might require small refactors to perform the union when declaring the type instead.

**Describe alternatives you've considered**

The workaround today is for consumers to use [ComponentProps](https://www.totaltypescript.com/react-component-props-type-helper), e.g. 
```TSX
type TheProps = ComponentProps<typeof TheReactUswdsComponent>
```

f3ve:
Is anyone working on this? I'd be happy to pick it up if not. Working on a new project and could really use this feature.

PR #2666, opened 2023-11-20
feat: Add all component prop type definitons as exports

RobKohr:
When creating a wrapper that extends a uswds component, it is useful to import in the Props that the wrapped component has so that the wrapped component’s prop types can be extended rather than having to copy and paste them.

To allow this, I am adding exports for all Props type definitions.

This shouldn’t have any impact on any projects that use react-uswds.

# Summary

## Related Issues or PRs

https://github.com/trussworks/react-uswds/issues/2667
#2728

## How To Test

Nothing should change for the use of the package, but anything importing the package should also be able to import the Props type definition for any component.


rswerve:
Thanks @RobKohr!  Take a look at the [existing issue](https://github.com/trussworks/react-uswds/issues/2587) for this change; there might be a few more considerations before we can call it done.

werdnanoslen:
> Thanks @RobKohr! Take a look at the [existing issue](https://github.com/trussworks/react-uswds/issues/2587) for this change; there might be a few more considerations before we can call it done.

@RobKohr I'd love to get this merged in the next release if you have some time to review the above. Do you need any help?

f3ve:
Is any help needed on this? I'd be happy to offer my assistance. I'm working on a new project for work and would really like to have this feature.

brandonlenz:
The change adding our component exports was merged in https://github.com/trussworks/react-uswds/pull/3118.

Closing this PR since I think we should be covered with Lea's change!

PR #2680, opened 2023-12-27
feat: updates for USWDS v3.7

### LINKED ISSUE #2682: [feat] Update to USWDS 3.7.x

No, global update

3.7+ (3.7.1 at time of writing)

Just an effort to get us updated to the latest uswds version.
See PR body of https://github.com/trussworks/react-uswds/pull/2680 for details on changes, already documented nicely by @werdnanoslen there!

werdnanoslen:
# Summary

Resolves #2682

The following changes in the 3.7 and 3.7.1 updates involved more than just scss updates. Changes that are checked off have been committed to this PR; the others are available for others to contribute updates to (or remove if no update is needed):

* [x] ~https://github.com/uswds/uswds/pull/5551~
      **edit**: this one was actually a USWDS-specific implementation change, no change needed for us
* [x] https://github.com/uswds/uswds/pull/5374
* [x] https://github.com/uswds/uswds/pull/5589
* [x] https://github.com/uswds/uswds/pull/5315
    **edit**: this was also a USWDS-specific refactor, but I did add the console errors from here
* [x] https://github.com/uswds/uswds/pull/5491
* [x] https://github.com/uswds/uswds/pull/5413
* [x] https://github.com/uswds/uswds/pull/5472

### DIFF
--- a/src/components/Identifier/Identifier/Identifier.stories.tsx
+++ b/src/components/Identifier/Identifier/Identifier.stories.tsx
@@ -132,7 +132,8 @@ export const identifierDefault = (): React.ReactElement => (
         <IdentifierLogo href="#">{testIdentifierLogo}</IdentifierLogo>
       </IdentifierLogos>
       <IdentifierIdentity domain="domain.edu.mil.gov">
-        <span aria-hidden="true">An</span>{` official website of the `}
+        <span aria-hidden="true">An</span>
+        {` official website of the `}
         <Link href="#">Test Agency Name</Link>
       </IdentifierIdentity>
     </IdentifierMasthead>
@@ -173,7 +174,8 @@ export const multipleParentsAndLogos = (): React.ReactElement => (
         <IdentifierLogo href="#">{testIdentifierLogo}</IdentifierLogo>
       </IdentifierLogos>
       <IdentifierIdentity domain="domain.edu.mil.gov">
-        <span aria-hidden="true">An</span>{` official website of the `}
+        <span aria-hidden="true">An</span>
+        {` official website of the `}
         <Link href="#">Test Agency Name</Link>
         {` and the `}
         <Link href="#">Other Test Agency Name</Link>
@@ -218,7 +220,8 @@ export const moreThanTwoParentsAndLogos = (): React.ReactElement => (
         <IdentifierLogo href="#">{testIdentifierLogo}</IdentifierLogo>
       </IdentifierLogos>
       <IdentifierIdentity domain="domain.edu.mil.gov">
-        <span aria-hidden="true">An</span>{` official website of the `}
+        <span aria-hidden="true">An</span>
+        {` official website of the `}
         <Link href="#">Test Agency Name</Link>
         {`, `}
         <Link href="#">Second Test Agency Name</Link>
@@ -239,7 +242,8 @@ export const noLogosEnglish = (): React.ReactElement => (
   <Identifier>
     <IdentifierMasthead aria-label="Agency identifier">
       <IdentifierIdentity domain="domain.edu.mil.gov">
-        <span aria-hidden="true">An</span>{` official website of the `}
+        <span aria-hidden="true">An</span>
+        {` official website of the `}
         <Link href="#">Test Agency Name</Link>
       </IdentifierIdentity>
     </IdentifierMasthead>
@@ -276,7 +280,8 @@ export const taxDisclaimerEnglish = (): React.ReactElement => (
         <IdentifierLogo href="#">{testIdentifierLogo}</IdentifierLogo>
       </IdentifierLogos>
       <IdentifierIdentity domain="domain.edu.mil.gov">
-        <span aria-hidden="true">An</span>{` official website of the `}
+        <span aria-hidden="true">An</span>
+        {` official website of the `}
         <Link href="#">Test Agency Name</Link>
         {`. Produced and published at taxpayer expense.`}
       </IdentifierIdentity>
@@ -320,7 +325,8 @@ export const taxDisclaimerAndMultipleParentsAndLogos =
           <IdentifierLogo href="#">{testIdentifierLogo}</IdentifierLogo>
         </IdentifierLogos>
         <IdentifierIdentity domain="domain.edu.mil.gov">
-          <span aria-hidden="true">An</span>{` official website of the `}
+          <span aria-hidden="true">An</span>
+          {` official website of the `}
           <Link href="#">Test Agency Name</Link>
           {` and the `}
           <Link href="#">Other Test Agency Name</Link>
--- a/src/components/Identifier/Identifier/Identifier.stories.tsx
+++ b/src/components/Identifier/Identifier/Identifier.stories.tsx
@@ -132,7 +132,7 @@ export const identifierDefault = (): React.ReactElement => (
         <IdentifierLogo href="#">{testIdentifierLogo}</IdentifierLogo>
       </IdentifierLogos>
       <IdentifierIdentity domain="domain.edu.mil.gov">
-        {`An official website of the `}
+        <span aria-hidden="true">An</span>{` official website of the `}
         <Link href="#">Test Agency Name</Link>
       </IdentifierIdentity>
     </IdentifierMasthead>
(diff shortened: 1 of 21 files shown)


brandonlenz:
Made feat since we want this to show in the changelog

brandonlenz:
Added `Resolves #2682` to PR body

src/components/Identifier/Identifier/Identifier.stories.tsx
```diff
@@ -132,7 +132,8 @@ export const identifierDefault = (): React.ReactElement => (
         <IdentifierLogo href="#">{testIdentifierLogo}</IdentifierLogo>
       </IdentifierLogos>
       <IdentifierIdentity domain="domain.edu.mil.gov">
-        {`An official website of the `}
+        <span aria-hidden="true">An</span>
+        {` official website of the `}
```
brandonlenz, line 136: Yay no more screen reader `Unofficial` vs `An official` silliness!

src/components/Modal/Modal.tsx
```diff
@@ -141,6 +141,13 @@ export const ModalForwardRef: React.ForwardRefRenderFunction<
   const ariaLabelledBy = divProps['aria-labelledby']
   const ariaDescribedBy = divProps['aria-describedby']
 
+  if (undefined === ariaLabelledBy) {
+    console.error(`${id} is missing aria-labelledby attribute`)
+  }
+  if (undefined === ariaDescribedBy) {
+    console.error(`${id} is missing aria-describedby attribute`)
+  }
```
brandonlenz, line 149: Any reason not to check for general falsiness so that we also check for null values? That's what USWDS is doing (https://github.com/uswds/uswds/pull/5315/files#diff-c7fd10b2638da39aa2cdf0583b9f14f1b05d5451909df6bbcd5ec528d74102afR223-R227)

```suggestion
  if (!ariaLabelledBy) {
    console.error(`${id} is missing aria-labelledby attribute`)
  }
  if (!ariaDescribedBy) {
    console.error(`${id} is missing aria-describedby attribute`)
  }
```

src/components/Modal/Modal.test.tsx
```diff
@@ -202,6 +202,23 @@ describe('Modal component', () => {
     expect(modalWindow).not.toHaveAttribute('aria-describedby')
   })
 
+  it('throws an error if labelledby or describedby is null', async () => {
```
brandonlenz, line 205: Currently the codes is checking for `undefined` specifically (See comment on that). A test for null would be good as well though!

```suggestion
  it('throws an error if labelledby or describedby is undefined', async () => {
```

src/components/forms/DatePicker/i18n.ts
```diff
@@ -38,8 +38,8 @@ export const EN_US = {
   backOneMonth: 'Navigate back one month',
   forwardOneYear: 'Navigate forward one year',
   forwardOneMonth: 'Navigate forward one month',
-  clickToSelectMonth: 'Click to select month',
-  clickToSelectYear: 'Click to select year',
+  selectMonth: 'Select month',
+  selectYear: 'Select year',
```
brandonlenz, line 42: 👍🏻

brandonlenz: Thanks for renaming the object key as well! 🎉

src/components/forms/RangeInput/RangeInput.tsx
```diff
@@ -17,47 +19,38 @@ type RangeInputProps = {
...
       ref={inputRef}
       type="range"
       {...inputProps}
-      aria-valuemin={ariaMin}
-      aria-valuemax={ariaMax}
-      aria-valuenow={ariaValue}
-      onChange={(e): void => onValueChange(e)}
+      min={rangeMin}
+      max={rangeMax}
+      defaultValue={value}
```
brandonlenz, line 52: I think this should be:

```suggestion
      value={value}
```
React wants only a controlled or uncontrolled input. In our case, we're using a react state to control the value, so its a controlled input, so we should be using the `value` prop

src/components/forms/RangeInput/RangeInput.tsx
```diff
@@ -17,47 +19,38 @@ type RangeInputProps = {
...
-  }
-  const [ariaValue, setAriaValue] = React.useState<number | undefined>(
-    convertValueType(calculatedAriaValueNow)
-  )
-  const onValueChange: React.ChangeEventHandler<HTMLInputElement> = (e) => {
-    if (!inputProps['aria-valuenow']) setAriaValue(e.target.valueAsNumber)
-  }
+  const rangeUnit = textUnit || defaultUnit
+  const rangePreposition = textPreposition || defaultPreposition
+  const rangeValue = inputProps.defaultValue || defaultVal
```
brandonlenz, line 38: I'll make some renaming suggestions that I think will help with the value/defaultValue controlled/uncontrolled thing for future maintenance of the component.

We also probably want to check for undefined specifically here, since passing in `0` as the default value is `falsey`, but I think would be a legitimate thing to want to do

```suggestion
  const startingValue = defaultValue !== undefined ? defaultValue : defaultVal
```

src/components/forms/RangeInput/RangeInput.tsx
```diff
@@ -17,47 +19,38 @@ type RangeInputProps = {
...
 }: RangeInputProps & JSX.IntrinsicElements['input']): React.ReactElement => {
   const classes = classnames('usa-range', className)
   // input range defaults to min = 0, max = 100, step = 1, and value = (max/2) if not specified.
   const defaultMin = 0
   const defaultMax = 100
-  const { min, max, defaultValue } = inputProps
+  const defaultVal = Math.round(defaultMax / 2)
+  const defaultUnit = ''
+  const defaultPreposition = 'of'
+  const { min, max } = inputProps
```
brandonlenz, line 33:
```suggestion
  const { min, max, defaultValue } = inputProps
```

brandonlenz: I also wonder if we should parse the `value` prop from `inputProps` and try to use that in case someone is passing it in 🤔 

That would look like 

```suggestion
  const { min, max, defaultValue, value: valueProp } = inputProps
```

src/components/forms/RangeInput/RangeInput.tsx
brandonlenz: If we wanted to also accept `value` from input props as `valueProp`, we'd want that value used here as well:

```suggestion
  const startingValue =
    valueProp !== undefined
      ? valueProp
      : defaultValue !== undefined
      ? defaultValue
      : defaultVal
```

src/components/forms/RangeInput/RangeInput.tsx
brandonlenz, line 1: We should probably rename this component, but lets do that separately from this PR:

https://github.com/trussworks/react-uswds/issues/2686

brandonlenz (changes requested):
I recommend making the change from defaultValue to value in the RangeInput specifically.

Everything else looks good. Some other optional suggestions to go along with the change I'm interested in

src/components/Modal/Modal.tsx
werdnanoslen: I swear I did that at first and it somehow failed tests, but now they pass. Anyway, fixed 🤷

src/components/forms/RangeInput/RangeInput.tsx
```diff
@@ -17,47 +19,38 @@ type RangeInputProps = {
...
+  const callout = `${value} ${rangeUnit} ${rangePreposition} ${rangeMax}`
 
   return (
     <input
       data-testid="range"
+      aria-valuetext={callout}
       className={classes}
       ref={inputRef}
       type="range"
       {...inputProps}
```
brandonlenz, line 49:
```suggestion
      {...remainingInputProps}
```

src/components/forms/RangeInput/RangeInput.tsx
brandonlenz: Just realized, `defaultValue` passed in still gets applied to the input triggering the console error. If we pull out `remainingProps` and then spread them below, its all good

```suggestion
  const {
    min,
    max,
    defaultValue,
    value: valueProp,
    ...remainingInputProps
  } = inputProps
```

brandonlenz (approved):

brandonlenz:
Thanks for making those changes!!

PR #2701, opened 2024-01-11
feat: moved overlay into header component

### LINKED ISSUE #1948: [feat] Include usa-overlay div in Header?

Relates to Header

It looks like we always include `<div className={`usa-overlay ${expanded ? 'is-visible' : ''}`}></div>` immediately preceding uses of the Header component. This div covers the entire page and is used in displaying the menu on mobile devices.  

Would a user ever _not_ want this div? If not, I would like that div included in the Header component. 

We could do nothing. Consumers of the package can add that div if they like.
The other option is to make the overlay its own, albeit simple component.

werdnanoslen:
# Summary

Moved page overlay into header component, accessible from a prop

This updates the Header stories as well as the template stories using the header

## Related Issues or PRs

Resolves #1948 

## How To Test

Run new test in header.test.tsx

### Screenshots (optional)
Still looking normal:

image

image


shkeating (approved):

brandonlenz (approved):
I like this. 🚢

Issue #1742, opened 2021-11-08
[docs] Document localization and internationalization - How do we improve Spanish support?


mgifford:
**Does this documentation relate to a specific USWDS feature?**
Content should need to be available in English & Spanish

**Is your documentation related to a problem or inconsistency? Please describe.**
There is nothing that I can find for supporting bilingual content

**Describe the documentation you'd like**

I assume that we'd be extending it with https://www.npmjs.com/package/react-localization but haven't tested it.

haworku:
Reviving this issue after seeing #2399.

I think it would be good to have more of a pattern for how we add lang to the application. For example instead of adding lang props as standalone props

werdnanoslen:
> 
> ```
> type FileInputProps = {
 > id: string
>  name: string
>  lang?: {dragText?: string, chooseText?: string}
> }
> ```
> That makes it easier for consumers to know where to pass in their i18n localized strings. It also follows the pattern on react-uswds of surfacing props and compose-able React components and allow applications to customize for their use case.

I'd be in favor of that. We'd also have to pass in the lang identifier (e.g. "`lang=es`, `lang=en-GB`")

haworku:
@werdnanoslen Oh right I forgot about `lang` being a reserved html attribute also. I think that is a case then for separating this into a more distinctly named prop since folks might think lang means the lang locale. Maybe we call it `i18n` or something like that? or `dicitionary` ?  I'm trying to make a prop that would represent the lang dictionary to use for that component, the keys are the specific keys of text chunks we have in a component. And then if nothing is set for that key, we continue defaulting to english for that. If folks are using an i18n librayr they would just pass their lang strings into that (and they would already be localized based on what they are using in the app).

PR #2707, opened 2024-01-17
feat: parametrized StepIndicator strings

### LINKED ISSUE #2705: [feat] Support multiple languages on Step Indicator

Step Indicator

Latest

Yes. We have a site that needs to support multiple languages. The Step Indicator has the word "of" hardcoded in the component, so in Spanish, we need the step indicator's current step as "2 de 5" instead of "2 of 5."

Perhaps an optional prop to translate the text? Something like `translationForOf` that accepts a string. If not provided, it defaults to "of"

N/A

**Additional context**

Something to support this:
Screenshot 2024-01-16 at 7 54 22 PM

werdnanoslen:
# Summary

Parametrized strings in StepIndicator component to allow translation of "Step" and "of" strings. 

There are still more English strings hard coded in the StepIndicatorStep subcomponent, as well as other components, but this narrow improvement is to allow for a client project to proceed. I will continue pushing for resources to prioritize an i18n strategy (see https://github.com/trussworks/react-uswds/issues/1742)

## Related Issues or PRs

Resolves #2705 

## How To Test

- Run new tests
- Use storybook controls and inspect 👇 

### Screenshots (optional)
image

### DIFF
--- a/src/components/stepindicator/StepIndicator/StepIndicator.stories.tsx
+++ b/src/components/stepindicator/StepIndicator/StepIndicator.stories.tsx
@@ -18,10 +18,27 @@ Updates users on their progress through a multi-step process.
       },
     },
   },
+  argTypes: {
+    stepText: {
+      control: 'text',
+    },
+    ofText: {
+      control: 'text',
+    },
+  },
+  args: {
+    stepText: 'Step',
+    ofText: 'of',
+  }
 }
 
-export const defaultStepIndicator = (): React.ReactElement => (
-  <StepIndicator headingLevel="h4">
+type StorybookArguments = {
+  stepText: string
+  ofText: string
+}
+
+export const defaultStepIndicator = (args: StorybookArguments): React.ReactElement => (
+  <StepIndicator headingLevel="h4" ofText={args.ofText} stepText={args.stepText}>
     <StepIndicatorStep label="Personal information" status="complete" />
     <StepIndicatorStep label="Household status" status="complete" />
     <StepIndicatorStep label="Supporting documents" status="current" />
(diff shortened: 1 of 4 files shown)


src/components/stepindicator/StepIndicator/StepIndicator.tsx
```diff
@@ -16,6 +16,8 @@ type StepIndicatorProps = {
     HTMLHeadingElement
   >
   headingLevel: HeadingLevel
+  stepText?: string
```
shimonacarvalho, line 19: If there are more of these, would it be better to do an object with multiple entries?
```
text: { step?: string, of?: string }
```
That way we can extend without adding loads more props?

werdnanoslen: These are the only two strings on the top-level component, so would it still be beneficial to put it in such an object? 

The other strings are on the StepIndicatorStep sub-component, which _could_ use such a treatment, but I don't think we have that pattern on other components and wanted to keep this PR tight enough that we can unblock the project requirement.

shimonacarvalho: Okay no worries, that works. Lemme test.

src/components/stepindicator/StepIndicator/StepIndicator.stories.tsx
```diff
@@ -70,8 +108,13 @@ export const smallCounters = (): React.ReactElement => (
...
-export const differentHeadingLevel = (): React.ReactElement => (
-  <StepIndicator headingLevel="h2">
+export const differentHeadingLevel = (
+  args: StorybookArguments
+): React.ReactElement => (
+  <StepIndicator
+    headingLevel="h2"
+    ofText={args.ofText}
+    stepText={args.stepText}>
     <StepIndicatorStep label="Personal information" status="complete" />
```
shimonacarvalho, line 118: Recommend removing the passed in text in one of the tests so that we can ensure the fallback to english works correctly.

shimonacarvalho (approved):
LGTM, tested in storybook.
One suggestion about ensuring the fallback is tested.

src/components/stepindicator/StepIndicator/StepIndicator.stories.tsx
werdnanoslen: added

shimonacarvalho (approved):
Tested and passed.

PR #2716, opened 2024-01-21
feat!: Webpack/Jest to Vite/Vitest migration

jpandersen87:
# Summary

This is a conversion from using webpack/jest as bundler/test library to vite/vitest. This will ensure the react-uswds library can support the new ESM standard, while maintaining backwards compatibility with CJS. All npm scripts have been updated accordingly to perform the same intended actions. Building is two steps: build the library and build the uswds assets (css, related files, etc.) dictated by the `--mode uswds` argument (running just `build` will do these steps for you).

Vite
- Output bundles for all formats Vite supports (umd, cjs, iife, es)
- Output source maps for all bundles
- package.json updated with (hopefully) best practices so that the correct bundle will always be selected by users of this library
- Parity with all other outputs (styles and assets)

Vitest
- Maintains parity with jest configuration (coverage, etc.)

Also included:
- new eslinting for storybook and vitest with updates to fix warnings/errors they detected
- husky added to project, with the precommit script set to run checks on: lint, format, test related (no more forgetting to run that darn format:fix!)
- all dependencies up-to-date as of 2/4/24

Notes:
- happo unfortunately still requires webpack and babel (if you ever move off of happo, these dependencies can likely be dropped)
- the happo and happo-ci scripts were updated to properly point to the renamed `.happo.cjs` file
- A new `bundles.test.tsx` file was added that is excluded during normal testing and is ran after building to do basic importing and rendering tests on each bundle format

## Related Issues or PRs

#192 

## How To Test

- Using whichever means available from your local project's package manager of choice: have it use this vite build for the react-uswds dependency
- Verify your project continues to work as expected


jpandersen87:
As mentioned, this is merely a proof-of-concept to show one direction that the project can go to not use webpack. It is intended foremost to be educational and provide insight. There are no expectations that this goes beyond draft, but if it is desired then necessary cleanup will happen (ex: vite config for various builds).

Vite uses a mix of esbuild and rollup under the hood. The lib mode has it's fair share of warts to work around like any bundler but I think I have it tamed. I have it creating bundles for every format it supports, because why not XD . Vite/Vitest are ES-first, thus the move from Jest (ES support on Jest is still experimental). With jest migrated to vitest the configuration needed to run vite, vitest, and storybook is now capable of being handled under one roof. There are some important differences in how jest/vitest work (especially mocks) that need to be considered. With webpack/jest usage removed, I was able to drop all of their directly-related dependencies along with babel ones. The version of JSDom being used apparently requires node 18+ so there were project changes in that regard (drop node 16 related stuff). USWDS since 3.0 has not supported IE so the IE11 line in the browser support list was dropped and focus-trap-react updated to current major version. The JSDom workaround for focus-trap-react uses a mock for the react library that enforces the necessary setting to always be disabled instead of the recommended mock for its dependency as I currently could not get vitest to handle loading it directly for mocking (aka, it currently is not dev-friendly to tell Vitest to handle loading dependencies-of-dependencies. direct dependencies that you import in your code mocks fine). There are warnings that appear now during linting, mostly from the new storybook plugin. Getting the eslint config right to handle cli-side and lib-side stuff properly is where i'm the weakest.

jpandersen87:
Tagging current maintainer on rotation: @werdnanoslen

jpandersen87:
Additional work to do a dependency-roundup of sorts to get all dependencies updated (including the React 18 upgrade) without errors can be done if moving this beyond draft is desired (as this PR would be a good time to do it due to its slight disruptiveness).

werdnanoslen:
@jpandersen87 we'd like to test this out on some internal projects soon — if this is ready, can you resolve the conflict please?

jpandersen87:
@werdnanoslen Branch updated

AnnaGingle (approved):

jpandersen87:
> @jpandersen87 thanks so much for bearing with us while we found opportunities on projects to test this out! We'd be happy to approve and merge this as soon as those remaining conflicts are resolved. 🚀

I take that as meaning tests have gone well? Branch has been updated.

werdnanoslen (approved):

brandonlenz (approved):
Doesn't need my approval, I don't think, but putting the stamp here since I tested it with success on my current project.

brandonlenz:
Just disabled the Node16 tests branch protection rule, since this PR removes tests for that legacy node version as well (rightly so). Once the pipeline re-runs, I expect it to be all green this time! 🤞🏻

mdmower-csnw:
In v8, I'm getting an error when I try to import `index.css` in my React project. v7 works fine.

```
ERROR in ./src/client/index.tsx 6:0-47
Module not found: Error: Package path ./lib/index.css is not exported from package C:\<path-to-project>\node_modules\@trussworks\react-uswds (see exports field in C:\<path-to-project>\node_modules\@trussworks\react-uswds\package.json)
```

and the import in `index.tsx` looks like:
```ts
import '@trussworks/react-uswds/lib/index.css';
```

PR #2726, opened 2024-01-22
feat: add DateSelector dateFormat prop

lpsinger:
# Summary


werdnanoslen (commented):
This seems useful — I'd just recommend updating `DatePicker.stories.tsx` to demonstrate how to use it. Do you mind if I commit to this branch with the story addition?

lpsinger:
> This seems useful — I'd just recommend updating `DatePicker.stories.tsx` to demonstrate how to use it. Do you mind if I commit to this branch with the story addition?

Go for it!

werdnanoslen:
> > This seems useful — I'd just recommend updating `DatePicker.stories.tsx` to demonstrate how to use it. Do you mind if I commit to this branch with the story addition?
> 
> Go for it!

Pardon the accidental close, [here's the commit](https://github.com/lpsinger/react-uswds/commit/68763b5b5440efdd224565227eb5be50b471bb08)

AnnaGingle (approved):
LGTM!

lpsinger:
Any update on this?

werdnanoslen:
> Any update on this?

Just waiting on code reviews from our engineers (for this and the other PRs), pardon the delay

lpsinger:
I fixed a form validation bug. Please review again.

src/components/forms/DatePicker/utils.tsx
```diff
@@ -464,14 +469,23 @@ export const isDateInvalid = (
   if (dateString) {
     isInvalid = true
 
-    const dateStringParts = dateString.split('/')
-    const [month, day, year] = dateStringParts.map((str) => {
+    const dateStringParts = dateString.split(
```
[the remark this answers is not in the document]
lpsinger: Should be fixed now.

src/components/forms/DatePicker/DatePicker.tsx
```diff
@@ -108,8 +118,7 @@ export const DatePicker = ({
 
   const handleSelectDate = (dateString: string, closeCalendar = true): void => {
     const parsedValue = parseDateString(dateString)
```
dakota002, line 120: This call of `parseDateString`  needs `dateFormat` as an arg

lpsinger:
I rebased this again. Any updates on when someone will review this?

lpsinger:
It's been a few more months since the last time I checked on this. Would someone please review it?

src/components/forms/DatePicker/constants.ts
```diff
@@ -34,3 +34,6 @@ export const YEAR_CHUNK = 12
 export const DEFAULT_MIN_DATE = '0000-01-01'
 export const DEFAULT_EXTERNAL_DATE_FORMAT = 'MM/DD/YYYY'
```
shkeating, line 35: I am good with passing the Happo diffs if you can explain why the switch to all caps for the help text? my thoughts would be to not change the default text if it can be avoided, it was originally in lowercase, which an argument could be made is more legible for folks with certain vision impairments.

lpsinger: I don't know why the case changed. I see what you mean, but I don't see anything in my patch that would cause it to do that.

lpsinger:
Rebased again. Would someone please review this?

src/components/forms/DatePicker/constants.ts
abbyoung: Looks like the hint text case change is only happening in the `.stories.tsx` file. It was changed from the static `mm/dd/yyyy` to `{argTypes.dateFormat ?? DEFAULT_EXTERNAL_DATE_FORMAT}`. Since we're using the constant value in the story now instead of hardcoding it, it's appearing as uppercase. I don't think anything needs to change.

abbyoung: If we want to keep the help text consistent as a valid style example, we can add something like `{argTypes.dateFormat ?? DEFAULT_EXTERNAL_DATE_FORMAT.toLowerCase()}`

brandonlenz: Maybe just in the .stories.tsx file do `DEFAULT_EXTERNAL_DATE_FORMAT.toLowerCase()` and call it a day (or hard code it. Lots of USWDS things like this are hard coded intentionally, and that's okay. Consumers just have to ensure when they use it that it makes sense)

lpsinger: Would you please push whatever changes you want to make to this?

PR #2727, opened 2024-01-22
fix: Exclude placeholder prop from Modal props

mdmower-csnw:
# Summary

The types generated for Modal props indicate that `placeholder` is available because `@types/react` prior to v18.2.43 has `placeholder` as an attribute in `HTMLAttributes`. Beginning with `@types/react` v18.2.43, the `placeholder` attribute has been removed from `HTMLAttributes`. As a result, projects that consume this package but are themselves on `@types/react` v18.2.43 or newer no longer recognize `placeholder` as an optional property; they see it as stemming from these Modal props rather than `JSX.IntrinsicElements['div']`, and that it is required, not optional.

Patch Modal to manually omit placeholder from available properties without updating packages, which would require a more significant lift.

Ref: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/67170

## How To Test

1. Add react-uswds to a TypeScript project that has a dependency on `@types/react` v18.2.48.
2. Use the `<Modal>` component without defining the `placeholder` attribute.
3. Attempt to build the project -- the TypeScript compiler will report an error that `placeholder` is required but not defined:
   ```
   [tsl] ERROR in /path/sanitized/MyComponent.tsx(95,10) 
      TS2741: Property 'placeholder' is missing in type '{ children: Element[]; ref: RefObject<ModalRef>; "aria-describedby": string; id: string; className: string; }' but required in type 'Pick<ModalProps, "children" | "id" | "onError" | "key" | "aria-label" | "className" | "defaultChecked" | "defaultValue" | "suppressContentEditableWarning" | "suppressHydrationWarning" | ... 249 more ... | "isInitiallyOpen">'.
   ```

## Additional Notes

1. This PR would be obsoleted by #2714 since that PR also updates `@types/react` to v18.2.48. That PR requires a major version update though, whereas this fix is safe to apply as a patch update.
1. Modal was verified to be the only component affected by this issue by grepping built `lib/components` for string `"placeholder"`. The only other component where the `placeholder` attribute appears explicitly is `FileInput` and there it is supported (`HTMLInputElement` instead of `HTMLDivElement`).

### DIFF
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -23,7 +23,8 @@ interface ModalComponentProps {
   isInitiallyOpen?: boolean
 }
 
-export type ModalProps = ModalComponentProps & Omit<JSX.IntrinsicElements['div'], 'placeholder'>
+export type ModalProps = ModalComponentProps &
+  Omit<JSX.IntrinsicElements['div'], 'placeholder'>
 
 export type ModalRef = {
   modalId: string
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -23,7 +23,7 @@ interface ModalComponentProps {
   isInitiallyOpen?: boolean
 }
 
-export type ModalProps = ModalComponentProps & JSX.IntrinsicElements['div']
+export type ModalProps = ModalComponentProps & Omit<JSX.IntrinsicElements['div'], 'placeholder'>
 
 export type ModalRef = {
   modalId: string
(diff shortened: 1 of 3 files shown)


werdnanoslen:
Hey @mdmower-csnw thanks for this PR! We decided to proceed with #2714 — is there anything in this PR that you'd still like to submit, or should we close this?

mdmower-csnw:
@werdnanoslen - Thanks, that sounds fine. Feel free to close this.

It's worth keeping this PR in mind if an issue report comes in from a user on v5 or v6 who is not ready to jump to v7. They can either stay on `@types/react@18.2.42` or this project could release patch updates to previous major releases (e.g. apply this commit on top of https://github.com/trussworks/react-uswds/tree/5.5.0 and release v5.5.1).

werdnanoslen:
> or this project could release patch updates to previous major releases (e.g. apply this commit on top of https://github.com/trussworks/react-uswds/tree/5.5.0 and release v5.5.1).

Fair point! If anyone requires such a patch—and maybe comments here—we could re-open this PR to apply as you suggested. I'll close for now though.

PR #2750, opened 2024-02-07
fix: replace card button with link

### LINKED ISSUE #2696: Replace Card's Buttons with Links
**Which USWDS component(s) does this relate to?**
Card

**What USWDS Version was this component introduced by?**
3.5.0 (missed in #2682)

**Describe the bug**
Replace `Button`s with `Link`s using the`usa-button` class, and be sure spacebar activation works, per https://github.com/uswds/uswds/pull/4385

**Code snippet**

Before:

```
<Card>
  ...
  <CardFooter>
    <Button type="button" secondary>
      Stick out
    </Button>
  </CardFooter>
</Card>
```

After:

```
<Card>
  ...
  <CardFooter>
    <Link href={"#"} className="usa-button">
      Stick out
    </Link>
  </CardFooter>
</Card>
```

**Further detail**

https://trussworks.github.io/react-uswds/?path=/docs/components-typography-link--default#styled-as-button

klalicki:
# Summary

This PR replaces `<Button>` components with `<Link>` components styled to look like buttons in the Storybook files for the Card component. Adds styling-related props from the original `<Button>` components (`outline`, `secondary`) via the corresponding CSS classes from USWDS - ie `.usa-button--secondary`
This PR also adds an optional `allowSpacebarActivation` prop to the Link component. When this prop is set to `true` the link can be activated with the spacebar in addition to the Enter key.

## How To Test
 
 Navigate to _Components > Card > Card Examples_, _Components > Card > Card Tests_, and _Components > Typography > Link > Styled As Button_ in Storybook.

For each button:
- navigate to it via keyboard
- confirm that it uses an `<a>` element instead of a `<button>`
- confirm that pressing the spacebar activates the link (will jump to top of page)

Note - visual regression testing will require review - this PR adds additional text to the story for _Link Styled As A Button_ specifying that it can be activated with the spacebar.

### Screenshots (optional)

Closes #2696


klalicki:
A couple of comments/questions for @werdnanoslen / whoever else at Truss is reading this:
- The new code adds a `onKeyDown` event handler to the `a` element returned by `<Link>`. This event listener would be overridden if a developer passes in their own custom `onKeyDown` handler. A developer _probably shouldn't_ pass a keyboard event listener to this component, but is this a scenario that I should account for? I could add the event listener with addEventListener in useEffect if so.
- The new code currently only adds the keyboard listener to the component when the default element `<a>` is being used. If a developer is using an `asCustom` component with `Link`, the keyboard listener is not added. My thought behind this was that you can't guarantee that the component passed to this one for `asCustom` will actually accept an event listener as a prop. Should I pass the event listener to the custom component anyway? I imagine this would most often be used for frameworks with routing like Next.js (whose Link component _does_ accept event listeners)
- possibly unrelated to the scope of the original issue - in order for the <Link> styled as a button to look correct, I have to also give it `variant="unstyled"` to get rid of the standard link text color, or else the default link text color overrides the white text color from `.usa-button`. Should the combination of characteristics  for Link styled as a Button (spacebar handler, variant="unstyled", and class of "usa-button") be combined into one property ie `<Link variant="button" href={"..."}>` or is it preferable to keep the styling and spacebar functionality separate?

werdnanoslen (approved):
Approving for design, accessibility, and meeting the issue AC. I'll defer to someone else on code review.

werdnanoslen:
> possibly unrelated to the scope of the original issue - in order for the styled as a button to look correct, I have to also give it `variant="unstyled"` to get rid of the standard link text color, or else the default link text color overrides the white text color from `.usa-button`. Should the combination of characteristics for Link styled as a Button (spacebar handler, variant="unstyled", and class of "usa-button") be combined into one property ie `<Link isButton href={"..."}>` or is it preferable to keep the styling and spacebar functionality separate?

I think the way you've done it maximizes user control and exposes underlying components in a clear way that `isButton` would obfuscate too much imo. So I like this approach!

klalicki:
Can @brandonlenz or someone else from Truss AppEng take a look at this when you have time?

src/components/Link/Link.tsx
```diff
@@ -33,6 +34,13 @@ export function isCustomProps<T>(
 ): props is CustomLinkProps<T> {
   return 'asCustom' in props
 }
+// keyboard handler for 'link as a button'
+const handleKeyDown = (e: React.KeyboardEvent) => {
```
brandonlenz, line 38: I was hoping typing this as `const handleKeyDown: React.KeyboardEventHandler<HTMLAnchorElement>` would maybe avoid the need to cast e.target, but alas, it does not. So IMO, this is fine.

brandonlenz (approved):

PR #2751, opened 2024-02-09
fix: Adds yarn as a devDep

dakota002:
# Summary

This PR adds yarn as a dev dependency. If you install a version of this package from git, (ex. github:lpsinger/react-uswds#dateFormat) the `prepare` script will run on `npm install`. This causes npm install to fail if yarn is not present on the user's machine. 

From the npm docs:

NOTE: If a package being installed through git contains a prepare script, its dependencies and devDependencies will be installed, and the prepare script will be run, before the package is packaged and installed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -119,7 +119,8 @@
     "ts-jest": "^29.1.1",
     "typescript": "^5.1.6",
     "webpack": "^5.52.1",
-    "webpack-cli": "^5.0.1"
+    "webpack-cli": "^5.0.1",
+    "yarn": "^1.22.21"
   },
   "resolutions": {
     "loader-utils": "2.0.4",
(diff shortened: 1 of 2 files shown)


mdmower-csnw:
This would be appreciated. It was a necessary addition to our (hopefully temporary) fork in order for CI agents to run `prepare` when consuming the package via GitHub instead of npmjs. https://github.com/CSNW/react-uswds/commit/202dce5335d9bd28c5bd5335d0e2e8bc41e2d319

jpandersen87:
I added this change in #2716 in case that PR lands before this one to ensure folks get it asap

werdnanoslen:
We're in the process of (pending some Github action changes) merging #2716 with this change incorporated, so I'm going to close this PR. Thanks for this @dakota002!

PR #2775, opened 2024-02-20
build(deps-dev): Bump prettier and stylelint

werdnanoslen:
# Summary

Updates prettier and stylelint deps. Dependabot for some reason isn't updating stylelint, so that might need a follow-up investigation or perhaps this triggers its attention 🤷 . Anyway, stylelint being outdated was causing some errors in dependabot's upgrade of prettier in #2748. So this upgrades both and also tells it to ignore USWDS functions in our _overrides file. For some reason, stylelint accepted as valid, but did not seem to use, the proper way (vs the `disable-next-line` way in this PR) of ignoring them in the rc file:
```
    "function-no-unknown": [
      true, 
      {
        "ignoreFunctions": ["line-height", "units"]
      }
    ]
```

## Related Issues or PRs

Supersedes and fixes lint error in https://github.com/trussworks/react-uswds/pull/2748

## How To Test

`yarn lint` should return no warnings or errors


src/styles/_overrides.scss
werdnanoslen, line 1: I tried using a more targeted rule (see summary above) in the rc file, but lint just ignored it. If anyone finds a way of making that work, remove these comment lines

werdnanoslen:
Superceded by #2716

Issue #2809, opened 2024-03-08
[fix] Language selector for two languages has invalid ARIA attribute


jaredcunha:
**ReactUSWDS Version & USWDS Version:**

There on the two-language variants of the Language Selector (button or link), there is an accessibility defect coming from `aria-controls="language-options"`. An element with `id="language-options"` would not need to exist for this control.

While `aria-controls` might not be function, this is showing up as a violation in automated tooling.

**To Reproduce**

1. Click on the language selector button (or link)
2. Inspect element
3. Note the `aria-controls` attribute

**Expected behavior**

The `aria-controls` attribute should not be on this version of the Language Selector

**Screenshots**

<img width="556" alt="Screenshot 2024-03-08 at 11 20 23 AM" src="https://github.com/trussworks/react-uswds/assets/1094951/fe0dec95-05f7-4572-8674-a33da7f40cd2">

**Additional context**

**Device and Browser Information (please complete the following information if describing a UI bug):**

PR #2810, opened 2024-03-08
fix: use aria-controls only when appropriate on LanguageSelector

werdnanoslen:
# Summary

Uses aria-controls on LanguageSelector only if specified, and for this component only uses it when appropriate

## Related Issues or PRs

#2809 

## How To Test

Check Storybook's accessibility tab for LanguageSelector for errors


werdnanoslen:
Merged this into #2819

PR #2819, opened 2024-03-12
fix: a11y omnibus package

### LINKED ISSUE #2809: [fix] Language selector for two languages has invalid ARIA attribute

### LINKED ISSUE #2811: [fix]: Make icons render accessibly and as intended by USWDS

[Has existed for some time]

**Describe the bug**

Icons in react-uswds do not have critical a11y features as [described here by USWDS](https://designsystem.digital.gov/components/icon/#accessibility-guidance).

**To Reproduce**

1. Go to any individual Icon story in storybook
2. Click on the Accessibility tab
3. See serious issues

**Expected behavior**

Each Icon should have...
* `aria-hidden` if icons are to be used on their own for visual purposes, exposed as a prop. If `false` (not hidden), the following should kick in...
* `title` inside the `<svg>`
* `aria-labelledby` referencing the above `title` by `id`

**Additional context**

These are 508 violations

werdnanoslen:
# Summary

Fixes a bunch of a11y issues that [Playwright](https://playwright.dev/docs/accessibility-testing) surfaced (322 of 553 tests failed):

**Source changes:**
- Required aria-hidden or aria-label or aria-labelledby on all `Icons`, defaulting to the former being true in storybook. Removed part of a test to this end.
- Changed `CardHeader` element from `header` to `div` to avoid [a11y issues](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/header#accessibility_concerns) that Storybook helpfully demonstrated, and it's how [USWDS implements](https://designsystem.digital.gov/components/card/#component-code) it anyway (probably for the same reason). This might be breaking in case anyone used css to target the rendered semantic `header` element. If so, please mark as breaking.
- Hid `Pagination`, `SocialLinks`, and `SearchButton`'s decorative svgs from screen readers
- Replaced presentation role with aria-label to fix a11y issue, and per USWDS implementation
- Used aria-controls on `LanguageSelector` only if specified, and for this component only uses it when appropriate

**Storybook-only changes:**
- Added labels to all `DatePicker`, `InputPrefix`, `InputSufffix`, `RangeInput`, `Select`, `Textarea`, and `TextInput` stories to fix a11y warnings there, and tidied up a bit
- Changed `Tag` story colors to meet contrast requirements
- Hid `IconList` icons from screen readers
- Changed `Footer`'s `h3`s to `p`s since class already handled styling and use of a heading seemed unwarranted
- Tidied `Pagination`'s stories
- Disabled 'skip-link' rule for `ExtendedNav` due to false-positive in Storybook

## Related Issues or PRs

Resolves #2811 
Resolves #2809

## How To Test

If you have Playwright installed locally, run `yarn test-storybook`. Otherwise, check the Accessibility tab of all the stories


werdnanoslen:
Hmm, ok maybe we shouldn't run this as a storybook test runner, since it requires a deployed storybook (sorta like how USWDS does it on their PRs). Thoughts on whether to merge this as _just_ a11y updates, with an alternative CI for a11y, or can we still make this work?

src/components/Tag/Tag.stories.tsx
```diff
@@ -20,9 +20,9 @@ Source: https://designsystem.digital.gov/components/tag/
 export const DefaultTag = (): React.ReactElement => <Tag>My Tag</Tag>
 
 export const CustomBg = (): React.ReactElement => (
-  <Tag background="#e44608">My Tag</Tag>
+  <Tag background="#d83933">My Tag</Tag>
```
kleinschmidtj, line 23: Totally a nit, but this could be a different color (that's also accessible of course). Right now this is the same as `bg-secondary` so there's no visual difference looking at the CustomBg and CustomClass stories.

werdnanoslen: I remember doing that on purpose, but I forget why 😅

werdnanoslen: i think because it _looked_ like bg-secondary, but it wasn't, so I was trying to keep it within the palette

kleinschmidtj:
> yarn test-storybook

@werdnanoslen I looked at the code and it looks good to me. I just wanted to run this testing command, but I don't see it the command in `package.json`. Is this new? Should this have been added as a new script?

werdnanoslen:
> > yarn test-storybook
> 
> @werdnanoslen I looked at the code and it looks good to me. I just wanted to run this testing command, but I don't see it the command in `package.json`. Is this new? Should this have been added as a new script?

oh whoops, I'll remove it. I started this ticket by seeing if we should run playwright as a GH task, but I didn't figure out how and pivoted to just fixing the issues it found from running locally.

kleinschmidtj (approved):
LGTM

PR #2843, opened 2024-03-21
ci: change action used for storybook deploy

jpandersen87:
# Summary
Changes the github action used for storybook deployment to one suggested by Storybook in their documentation on deployment to Github Pages (see: https://storybook.js.org/docs/sharing/publish-storybook#github-pages) in order to hopefully resolve errors in the current storybook deployment.

### DIFF
--- a/.github/workflows/deploy-storybook.yml
+++ b/.github/workflows/deploy-storybook.yml
@@ -26,19 +26,8 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version-file: '.node-version'
+          cache: 'yarn'
 
-      - name: Get yarn cache directory path
-        id: yarn-cache-dir-path
-        run: echo "::set-output name=dir::$(yarn cache dir)"
-
-      - uses: actions/cache@v1
-        id: yarn-cache
-        with:
-          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
-          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-yarn-
-            
       - name: Deploy 🚀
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-storybook.yml
```diff
@@ -42,9 +47,9 @@ jobs:
         run: yarn build-storybook
```
brandonlenz, line 47: I wonder if we need this step still? (And the `yarn install` above it too?)

Looks like this github action does both of those for us

gidjin: Agree on this feedback, not sure we need the steps Brandon noted. Can we trigger this build before merging it? I am not sure and I know that's not always possible with GHA.

werdnanoslen: i dont think so aside from testing on your own repo

werdnanoslen: Oh yeah, actually, i just duplicated the branch to our repo and it's now available in the Actions tab 
Screenshot 2024-03-22 at 11 46 18 AM

Should I run it?

brandonlenz: @jpandersen87 it was easier for us to test and iterate on a Truss-owned branch. If you make your PR diff match https://github.com/trussworks/react-uswds/pull/2844/files, I'm happy to approve so you get the credit!

Thanks for taking the time! PS: Our latest storybook deploy is live thanks to this change: https://trussworks.github.io/react-uswds/

brandonlenz (approved):

.github/workflows/deploy-storybook.yml
jpandersen87: haha, I was watching the deploy and saw the node16 warnings and made the action version update changes literally the same time as you did

PR #2847, opened 2024-03-22
ci: dangerfile and storybook brandimage fix

jpandersen87:
# Summary

- Updates the shouldTest flag in dangerfile so that it will properly be false if PR data isn't present
- Fixes storybook brandImage url (find `uswds.svg` relative, not from root)

## How To Test

- dangerfile steps in other workflows besides PR should no longer fail from attempting to run and encountering errors


ajfarkas (approved):
Thanks, we've been working on CI tests today!

PR #2868, opened 2024-04-04
fix: moved smooth-scroll application to useEffect

### LINKED ISSUE #2867: [fix] smooth scrolling css leaks from component to whole package

At least react-uswds v9.0.0

**Describe the bug**

This smooth scrolling leaks to any page where this package is used

https://github.com/trussworks/react-uswds/blob/db2e4f6e550ec316ec5637aeded3d720fec88d7b/src/components/InPageNavigation/InPageNavigation.module.scss#L1-L5

**To Reproduce**

1. Use react-uswds on a page, use any component but InPageNavigation
2. See that the smooth scrolling is applied to the `html` root

**Expected behavior**

It should only be applied where the InPageNavigation component is used.

**Additional context**

All browsers

werdnanoslen:
# Summary

Fixes leak of smooth-scroll, limits it just to pages where component is being used

Also changed a variable to be more descriptive

## Related Issues or PRs

Resolves #2867

## How To Test

Use storybook as an example, comparing the [live version](https://trussworks.github.io/react-uswds/?path=/docs/components-in-page-navigation--docs) (where smooth-scroll is applied on all pages after loading the component) vs what you can build locally, shown below:

Apr-04-2024 14-31-50

### DIFF
--- a/src/components/InPageNavigation/InPageNavigation.module.scss
+++ b/src/components/InPageNavigation/InPageNavigation.module.scss
@@ -1,5 +1,5 @@
 @media not (prefers-reduced-motion) {
-  html {
+  .smooth-scroll {
     scroll-behavior: smooth;
   }
 }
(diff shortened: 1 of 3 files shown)


kimallen (commented):
Looks good to me! I wonder if there would be value in including a test that checks for the class?

werdnanoslen:
> Looks good to me! I wonder if there would be value in including a test that checks for the class?

thanks, added

kimallen (approved):
Thanks for adding the test. Ship!

PR #2893, opened 2024-04-18
fix: Duplicate classes and invalid aria props on LanguageSelector

sawyerh:
# Summary

This fixes a few issues with the `LanguageSelector` implementation:

1. The `usa-language-container` class was being applied to both the parent `div` and the inner `button`, which causes layout issues due to the button receiving unexpected margins and relative positioning
2. #2809

An assumption I am making in this PR is that the intent of the `className` prop is to apply custom styling to the button, which is how it's demonstrated in Storybook.

## How To Test

Storybook

### Screenshots (optional)

CleanShot 2024-04-18 at 15 08 34@2x

### DIFF
--- a/src/components/LanguageSelector/LanguageSelector.tsx
+++ b/src/components/LanguageSelector/LanguageSelector.tsx
@@ -14,6 +14,9 @@ export type LanguageSelectorProps = {
   label?: string
   langs: LanguageDefinition[]
   small?: boolean
+  /**
+   * Custom classes for the call-to-action
+   */
   className?: string
   displayLang?: string
 }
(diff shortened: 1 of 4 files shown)


src/components/LanguageSelector/LanguageSelector.tsx
```diff
@@ -14,6 +14,9 @@ export type LanguageSelectorProps = {
   label?: string
   langs: LanguageDefinition[]
   small?: boolean
+  /**
+   * Custom classes for the call-to-action
+   */
```
sawyerh, line 19: ℹ️ This shows in Storybook: 

CleanShot 2024-04-18 at 15 09 03@2x

src/components/LanguageSelector/LanguageSelectorButton.tsx
```diff
@@ -27,8 +25,6 @@ export const LanguageSelectorButton = ({
     <button
       data-testid="languageSelectorButton"
       className={classes}
-      aria-expanded={isOpen}
-      aria-controls="language-options"
```
sawyerh, line 31: ℹ️ These are now passed in as props from `LanguageSelectorDropdown` since that's the only context they need applied

PR #2917, opened 2024-05-10
Ruizaj a11y issue fixes

ruizajtruss:
# Summary

### DIFF
--- a/src/components/LanguageSelector/LanguageSelectorDropdown.tsx
+++ b/src/components/LanguageSelector/LanguageSelectorDropdown.tsx
@@ -61,6 +61,7 @@ const LanguageSelectorDropdown: React.FC<LanguageSelectorProps> = ({
         <li className="usa-language__primary-item">
           <LanguageSelectorButton
             className={classes}
+            aria-controls="language-options"
             label={displayLabel?.label || label || langs[0].label}
             isOpen={isOpen}
             onToggle={() => setIsOpen((prevIsOpen) => !prevIsOpen)}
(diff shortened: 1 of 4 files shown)


werdnanoslen:
Thanks! Are any of these also in [here](https://github.com/trussworks/react-uswds/pull/2819)?

src/components/Search/SearchButton/SearchButton.tsx
```diff
@@ -44,6 +44,7 @@ export const SearchButton = ({
           name={buttonText}
           size={3}
           aria-hidden={true}
+          aria-label="Magnifying glass search icon"
```
ajfarkas, line 47: **💣 chore:** There's no need to label this icon, ‏because `aria-hidden=true`.

Detail:  
This icon is hidden from screen readers because it is decorative. The button is already labeled with text, so adding the `aria-label` would cause AT to read, "Search. Magnifying glass search icon".  
I find it helpful to aim for a minimum of text. For instance if I do want to label the Search icon, I would simply use "Search". The fact that it's a magnifying glass (or an icon) doesn't add anything for someone who can't see it.

src/components/LanguageSelector/LanguageSelectorDropdown.tsx
```diff
@@ -62,6 +62,7 @@ const LanguageSelectorDropdown: React.FC<LanguageSelectorProps> = ({
         <li className="usa-language__primary-item">
           <LanguageSelectorButton
             className={classes}
+            aria-controls="language-options"
```
ajfarkas, line 65: **💣 chore:** ‏This is a duplicate property, we should remove it.

The ID of the controlled menu is passed as a `const` via the `controls` attribute on this component. It's preferable to use the `const` anyway, so that updating the menu ID does not require an additional update to the `controls` prop.

ajfarkas (changes requested):
I like what we're trying to do here, though I'm not sure we need any of these changes. `aria-live="off"` is more appropriate, but I would not expect the character count to change if the user is not focused on that element.

Should we close this?

Issue #2918, opened 2024-05-10
[fix] DatePicker component uses defaultProps which is to be deprecated and  causes console warning.


Steven7926:
**ReactUSWDS Version & USWDS Version:**

    "@trussworks/react-uswds": "^9.0.0",
    "@uswds/uswds": "3.7.1",
**Describe the bug**

When building a form using your component library I kept running into a  console waring stating: 

> "Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead."

Upon further investigation I found that  this was coming from the DatePicker component. At the  end  of DatePicker.tsx you are using 
DatePicker.defaultProps = {
  minDate: DEFAULT_MIN_DATE,
}
to set a minimum allowable date for the datepicker component, this  is trggering the  warning.  
defaultProps is also used in the Tooltip component. 
**To Reproduce**

1. Create a page and use the datepicker  and tooltip components
2. Render that page
3. Open dev tools in the browser and view the console
4. See error:  "Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead."

**Expected behavior**

No warning should appear in the console to prep your library for deprecating feature support.

**Screenshots**

**Additional context**

**Device and Browser Information (please complete the following information if describing a UI bug):**
Google Chrome running a Vite React application on Windows.

aligg:
@Steven7926 and @brandonlenz (saw you've reviewed some things recently) - put up a quick PR for this one in case anyone has cycles to review: https://github.com/trussworks/react-uswds/pull/2949

Looks like workflows need approval from a maintainer as well.

aligg:
A fix for this is now merged and I imagine will go out with the next release.

PR #2949, opened 2024-06-14
fix: remove defaultProps from DatePicker and Tooltip

aligg:
# Summary
- remove use of defaultProps from DatePicker and Tooltip (DatePicker already sets the default ([code](https://github.com/trussworks/react-uswds/blob/2a011471920c2f893f0cf294752b61af0f30f2eb/src/components/forms/DatePicker/DatePicker.tsx#L63)) and I added in the position default in Tooltip) 
- This is my first time contributing so guidance appreciated if I missed anything 

## Related Issues or PRs
https://github.com/trussworks/react-uswds/issues/2918

## How To Test
I ran through the unit tests locally - can add additional testing if we think we need. 

### Screenshots (optional)
N/A


Steven7926 (approved):
Looks Good, TY!

src/components/Tooltip/Tooltip.tsx
```diff
@@ -56,7 +60,7 @@ export function Tooltip<
   const [wrapTooltip, setWrapTooltip] = useState(false)
   const [positionStyles, setPositionStyles] = useState({})
 
-  const { position, wrapperclasses, className } = props
+  const { wrapperclasses, className } = props
```
brandonlenz, line 63: One thing that might be nice is to continue to do prop destructing in one place. 

I don't have a strong opinion on whether `position` and its default value is moved here, or whether `wrapperClasses` and `className` destructuring are moved to the functional component argument. Other than perhaps that most of our components destructure in the component arg, so that'd be a point for consistency. 

@aligg What do you think?

brandonlenz (dismissed):
Change is good as is, one nit/suggestion I'd like to hear thoughts on before merging

src/components/Tooltip/Tooltip.tsx
aligg: @brandonlenz I just pushed up a change to move the prop destructuring into the component arg. If that looks good lmk and hopefully will be good to merge!

brandonlenz (approved):

src/components/Tooltip/Tooltip.tsx
brandonlenz: LGTM! Thank you!

PR #2848, opened 2024-03-22
[feat] USWDS 3.8.0 features

werdnanoslen:
# Summary

These are the features released by [USWDS in 3.8.0](https://github.com/uswds/uswds/releases/tag/v3.8.0):
* Added styles for indeterminate checkboxes and radios
[WIP]

## Related Issues or PRs

—

## How To Test

Confirm things work the same as release/PR notes in https://github.com/uswds/uswds/releases/tag/v3.8.0

PR #2962, opened 2024-06-24
feat: Upgrade to 3.8.1 with tooltip updates

werdnanoslen:
# Summary

This updates us to the latest uswds version, 3.8.1, which had two changes that required PRs here:
- [Make tooltip content hoverable](https://github.com/uswds/uswds/pull/5919)
- [Make tooltip dismissible with escape key](https://github.com/uswds/uswds/pull/5909)

## How To Test

1. Run new tests
2. Open [tooltip stories](http://localhost:9009/?path=/docs/components-tooltip--docs) and confirm (1) that you can hover on the tooltip body and (2) dismiss it by keyboard only with the Escape (and no other) key. Previously, we had set it so that any key would dismiss, which was neither helpful in general nor accessible by keyboard navigation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
   },
   "homepage": "https://github.com/trussworks/react-uswds#readme",
   "peerDependencies": {
-    "@uswds/uswds": "^3.7.1",
+    "@uswds/uswds": "^3.8.1",
     "focus-trap-react": "^10.2.3",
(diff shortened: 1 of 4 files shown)


werdnanoslen:
Merged into https://github.com/trussworks/react-uswds/pull/2848

PR #3011, opened 2024-08-23
fix: Update TextInput ref forwarding

ocoutts-usds:
# Summary

Currently this component forwards refs in a way that does not work in new versions of react and I'm not sure if it ever worked in any version of react. Leaving the old way in case somebody is using it.

### DIFF
--- a/src/components/forms/TextInput/TextInput.tsx
+++ b/src/components/forms/TextInput/TextInput.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { forwardRef } from 'react'
 import classnames from 'classnames'
 import { ValidationStatus } from '../../../types/validationStatus'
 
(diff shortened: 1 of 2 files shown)


src/components/forms/TextInput/TextInput.tsx
```diff
@@ -28,43 +28,51 @@ export type OptionalTextInputProps = CustomTextInputProps &
...
+        name={name}
+        type={type}
+        ref={inputRef || ref}
+        {...inputProps}
+      />
+    )
+  }
+)
 
+TextInput.displayName = 'TextInput'
```
brandonlenz, line 77: A bit weird, but the existing pattern for making sure the components have a display name is done like so:
- https://github.com/trussworks/react-uswds/blob/4513e51c04caa6c6f83ae57281b8419d874ab894/src/components/forms/DatePicker/Day.tsx#L27
- https://github.com/trussworks/react-uswds/blob/4513e51c04caa6c6f83ae57281b8419d874ab894/src/components/forms/DatePicker/Day.tsx#L113

I don't remember if there was a reason we did it that way, or if your approach is better, so I won't block on it

brandonlenz (approved):

kimallen:
Could you add some context in the Summary for what bug this addresses?

brandonlenz:
> Currently this component forwards refs in a way that does not work in new versions of react and I'm not sure if it ever worked in any version of react. Leaving the old way in case somebody is using it.

Yea this might have been a relic from much older versions of react, since I think this was likely one of the first components built for ReactUSWDS

src/components/forms/TextInput/TextInput.tsx
ocoutts-usds: I am happy to change but I find it a bit strange - this approach makes it so that the display name of the component (shows up in dev tools, error messages) will be `DayForwardRef` not `Day` which I find somewhat unideal.

brandonlenz: I agree with you. Let's leave it as is. Can always update easily if we find any problems or want to unify one way or the other

brandonlenz:
@all-contributors please add [ocoutts-usds](https://github.com/ocoutts-usds) for bug, code

brandonlenz:
https://github.com/all-contributors please add ocoutts-usds for bug, code

brandonlenz:
@all-contributors please add @ocoutts-usds for bug,code

PR #3096, opened 2025-02-28
fix: tooltip ID cleanup

mcmcgrath13:
# Summary

## Related Issues or PRs

Cleanup based on PR feedback to #3094 

Use hooks properly (per react's rules of hooks) in generating and using the ID for the tooltip component.

## How To Test

Pure refactor/bug fix.

### Screenshots (optional)

No react hooks warnings when loaded
image

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -38,8 +38,10 @@ const TRIANGLE_SIZE = 5
 const DEFAULT_POSITION = 'top'
 
 // useId was introduced in React 18 - polyfill for older versions
-const useGenId = (): string => {
-  return React.useId?.() ?? `${Math.floor(Math.random() * 900000) + 100000}`
+const useId = (): string => {
+  const id1 = React.useId?.()
+  const [id2] = React.useState(`${Math.floor(Math.random() * 900000) + 100000}`)
+  return id1 ?? id2
 }
 
 export function Tooltip(props: DefaultTooltipProps): JSX.Element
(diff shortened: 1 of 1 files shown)


brandonlenz:
@mdmower-csnw Does this PR match your expectation?

mdmower-csnw:
> @mdmower-csnw Does this PR match your expectation?

@brandonlenz - This fixes the issues I mentioned in comment https://github.com/trussworks/react-uswds/pull/3094#discussion_r1974973101, so from that perspective it's ok (the pull request description could use a little work, though).

As far as what good this PR and #3094 do for users... I'm less clear. The ID is still generated, not definable as requested in #2562. There's a better chance that the ID will be the same each time a test suite runs (for example `tooltip-:r3:`), but there's plenty of ways it can change, too. Re-rendering the `Tooltip` component during client-side rendering generates a new ID each time. I'm less experienced with server-side rendering, but I wonder if adding a new use of `Tooltip` that gets pre-rendered before the `Tooltip` of interest would also cause the ID to change? If the only goal here is to make the ID a bit more deterministic so that the ID is _likely_ to be the same when the component is rendered for tests, then I guess this is ok.

mcmcgrath13:
The main benefit is that React's `useId` is designed to play nicely with server side rendering. Hydration handles `useId` so that it is deterministic between server and client. It's also a much more targeted thing to mock during testing than random number generation

mcmcgrath13:
@brandonlenz any remaining concerns?

ajfarkas (approved):
lgtm

PR #3125, opened 2025-03-27
chore: move release-please files to root

ajfarkas:
# Summary

`release-please` is still throwing errors. It's unclear how paths work, but I think I can get around this by placing the config and manifest files on the root level.

This also updates the name of the Semantic PR job to match what's in the Branch Protection Rule (in Settings).  
I was seeing `Expected — Waiting for status to be reported` on `lint` and `action-semantic-pull-request`. I tracked down that these expectations come from the Branch Protection rules, and that if we don't see actions with those names, merging is blocked. 

## Related Issues or PRs

extension of #3124 et al

## How To Test

I still think this has to be merged to fully test. Preferably someone with more knowledge about GitHub actions will review this.


ajfarkas:
Note that I would only expect the Semantic PR check to update for the next PR

lea-m-truss (approved):
Let's give it a shot!

Issue #3159, opened 2025-05-16
[fix] v10.0 causes error for import statements: "Cannot find module '@trussworks/react-uswds' or its corresponding type declarations."


rachel-bernard:
ReactUSWDS Version & USWDS Version: 10.0

On installing the latest update via `npm i @trussworks/react-uswds@latest`, all places in my code that referenced the package are throwing errors.

Also seeing many build errors that say: "Unable to resolve path to module '@trussworks/react-uswds' " or "Unable to resolve path to module '@trussworks/react-uswds/lib/index.css'"

Reverting to v9.1.0 fixes the issue.

**To Reproduce**

Install the latest version of react-uswds in your project and attempt to import a module into any typescript code (such as `import { Table } from "@trussworks/react-uswds";`

**Expected behavior**

**Screenshots**
![Image](https://github.com/user-attachments/assets/b9c58043-a1e8-4767-a487-1bc11f139484)

![Image](https://github.com/user-attachments/assets/d2c5de71-e9af-430e-9710-8ed9a2194dee)

**Additional context**

**Device and Browser Information (please complete the following information if describing a UI bug):**

brandonlenz:
Not sure if the root cause is the same between what we're seeing, but some initial searching (https://github.com/nrwl/nx/issues/19149) indicates it could be related to the Vite migration (#2716), which appears to have set up our `main` and `module` properties in package.json

brandonlenz:
Shoulda looked for this first, but looks like the artifact published to NPM is mostly empty. 

Our github action no longer builds the code to `lib`, so it never gets included in the `pack` command.

I'll likely have to come back to this after the weekend. Not sure if this is from the yarn migration or whether any of our scripts changed

rachel-bernard:
Thank you @brandonlenz for all the time and attention you have given to this issue so far 👏 👏

brandonlenz:
> Thank you [@brandonlenz](https://github.com/brandonlenz) for all the time and attention you have given to this issue so far 👏 👏

This one was entirely my fault trying to hastily release 10.0.0

Please give 10.0.1 a try

PR #3160, opened 2025-05-16
fix(release): Fix import issues post release

brandonlenz:
# Summary

There seems to be issues for consumers of the library dating back to [8.0.0](https://github.com/trussworks/react-uswds/releases/tag/8.0.0) (we were able to replicated it as early as this release)

## Related Issues or PRs

Attempts to fix: #3159

## How To Test

Honestly, not sure. This seems to have been an issue for a little while (since version 8.0.0)

I didn't see this _locally_ pointing to the package on my machine before publish, but not sure that means much.

Might need to release and re-test until we can figure it out

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,7 +118,6 @@
     "happo-plugin-storybook": "^4.4.3",
     "happo.io": "^12.1.0",
     "jsdom": "^24.0.0",
-    "pinst": "^3.0.0",
     "prettier": "^3.2.5",
     "react": "^19.0.0",
     "react-dom": "^19.0.0",
--- a/package.json
+++ b/package.json
@@ -12,8 +12,7 @@
     "component library"
   ],
   "type": "module",
-  "main": "./lib/index.cjs",
-  "module": "./lib/index.js",
+  "main": "./lib/index.d.ts",
   "types": "./lib/index.d.ts",
   "exports": {
     ".": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -12,8 +12,7 @@
     "component library"
   ],
   "type": "module",
-  "main": "./lib/index.cjs",
-  "module": "./lib/index.js",
+  "main": "./lib/index.d.ts",
```
brandonlenz, line 15: So @ronaktruss has pointed out that I probably am wrong about this 😅

package.json
```diff
@@ -12,8 +12,7 @@
     "component library"
   ],
   "type": "module",
-  "main": "./lib/index.cjs",
-  "module": "./lib/index.js",
```
brandonlenz, line 16: Lots of info is pointing to this key being a problem, but there is some documentation about it here:

https://esbuild.github.io/api/#main-fields

brandonlenz:
@jpandersen87 have you been successfully using React USWDS > 8.0.0 since the Vite migration?

I suspect the project we're seeing issues on might need to update project configs.

We're now also seeing:

```
✘ [ERROR] Missing "./lib/index.css" specifier in "@trussworks/react-uswds" package [plugin vite:dep-scan]

    src/App.tsx:39:7:
      39 │ import '@trussworks/react-uswds/lib/index.css';
```

brandonlenz:
Looks like the .tgz generated by our package-release job is empty. Does not have a `lib` directory.

That's likely the issue. Need to verify which `yarn` scripts are needed to run on Yarn v3 to generate a release

PR #3189, opened 2025-06-07
validate DatePicker on blur

elijah-wright:
# Summary
this PR fixes the skipped tests in `DatePicker.test.tsx` by adjusting `DatePicker` to validate on blur rather than with a `useEffect` hook, which matches with USWDS. it also sets the `aria-invalid` attribute and uses `reportValidity` to send `VALIDATION_MESSAGE` to the user. this PR also changes some of the tests that were dependent on the `useEffect` behavior

## Related Issues or PRs

none

## How To Test

`yarn test`


elijah-wright:
yea, I just wanted more feedback first

PR #3202, opened 2025-06-24
feat: combobox custom options issue#3201

### LINKED ISSUE #3201: [feat] Allow ComboBox to return JSX in its drop down options

[ComboBox](https://designsystem.digital.gov/components/combo-box/) 

Currently the ComboBox takes options as a prop in the form of a ComboBoxOption type that consists of 2 strings, a value and a label. 

`export interface ComboBoxOption {
  value: string
  label: string
}
`
This means that all the options in the drop down can only be simple strings. 
This is frustrating when you need the option to be more sophisticated (more than one piece of info per selection)

Id like to be able to pass in JSX as a label for a drop down option in the ComboBox.
An easy way to do this would be to allow the ComboBoxOption interface to have an additional, optional render function field that returns JSX. 
`export interface ComboBoxOption {
  value: string
  label: string
  render?: () => JSX.Element
}
`
Then in the returned list tag, check for the render function, if it was passed, invoke it to return the JSX as the option. If it wasn't, just use option.label instead. 

`{option.render ? option.render(): option.label}`

Steven7926:
# Summary

## Related Issues or PRs

Closes #3201 

## How To Test

1. Use the ComboBox and in the options passed to it, include a render function that returns the JSX to use for the option.
2. Ex: ` const fruitList = Object.entries(fruits).map(([key, value]) => ({
    value: key,
    label: value,
    render: () => {
      return (
        <div className="padding-2 border border-base-lighter radius-md bg-white">
          <div className="font-sans-md text-bold">
            {value}
          </div>
          <div className="font-sans-sm text-base">
            Category: Fruit
          </div>
        </div>
      )
    }
  }))`
3. Pass the options as the options prop to the ComboBox
4. Click on the ComboBox and notice the options are rendered using the custom JSX you passed.

### DIFF
--- a/src/components/forms/ComboBox/ComboBox.stories.tsx
+++ b/src/components/forms/ComboBox/ComboBox.stories.tsx
@@ -46,6 +46,54 @@ export const DefaultComboBoxWithPropOptions = (): JSX.Element => {
   )
 }
 
+export const WithCustomOptions = (): JSX.Element => {
+  const veggieList = Object.entries(veggies).map(([key, value]) => ({
+    value: key,
+    label: value,
+    render: () => {
+      return (
+        <div className="padding-2 border border-base-lighter radius-md bg-white">
+          <div className="font-sans-md text-bold">
+            {value}
+          </div>
+          <div className="font-sans-sm text-base">
+            Category: Veggies 
+          </div>
+        </div>
+      )
+    }
+  }))
+
+  const fruitList = Object.entries(fruits).map(([key, value]) => ({
+    value: key,
+    label: value,
+    render: () => {
+      return (
+        <div className="padding-2 border border-base-lighter radius-md bg-white">
+          <div className="font-sans-md text-bold">
+            {value}
+          </div>
+          <div className="font-sans-sm text-base">
+            Category: Fruit
+          </div>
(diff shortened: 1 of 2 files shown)


brandonlenz:
When a maintainer has time to review and consider these changes, some things to evaluate:

1. At some point [`<option>` tags could only contain strings](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/option). Looks like that's no longer the case, but only for modern browsers
2. [The USWDS implementation](https://designsystem.digital.gov/components/combo-box/) only uses strings, so this is a case of balancing flexibility with potential footguns for consumers
3. Consider the impacts on filtering. One of the core features of the combobox is to filter options by label against default or custom regex. It will be important for consumers to follow the USWDS guidance, if they're now able to render options that don't match their labels:
   > **Use option strings familiar to users**. The combo box filters by matching strings. Include option text that includes familiar strings or spellings (for example, if using the combobox with a state list, include the postal abbreviation in the option text: District of Columbia (DC)).

I bet we'll want to document somewhere the details about filtering so that consumers don't get confused if their custom rendered content doesn't filter neatly if not tied closely to the `label` prop that they pass in

PR #3236, opened 2025-08-13
fix: update FilePreview to handle multiple renders in strict mode

### LINKED ISSUE #2433: [fix] FilePreview component breaks FileReader use in onChange events

@trussworks/react-uswds: 4.2.1
@uswds/uswds: 3.1.0

**Describe the bug**

When a function that uses `FileReader` is called from an `onChange` event in the `FileInput` component, it throws an error because `FileReader` is already being used on any added files in the `FilePreview` component.

Ideally, users should be able to disable calling `FileReader` inside the `FilePreview` component so that it can still be used in change events.

**To Reproduce**

1. On Chrome, create a component using `FileInput` with an `onChange` event using `FileReader`
2. Add a file
3. See error `Uncaught DOMException: Failed to execute 'readAsDataURL' on 'FileReader': The object is already busy reading Blobs.`

**Expected behavior**

Using `FileReader` within the `FileInput` component should be optional

**Additional context**

My specific use case for this is to be able to validate text file input in `onChange` events, and I'd be happy to put in a PR that allows for disabling calling the `FileReader` which seems to only be used for generating image previews. Thanks for taking a look!

https://github.com/trussworks/react-uswds/blob/c31423439cf6a30117fd327ee6cc2b1aacdb9124/src/components/forms/FileInput/FilePreview.tsx#L23-L34

 - Browser Chrome
 - Version 113.0.5672.126
 - Also reproduced on Firefox

doug-s-nava:
# Summary

A bug exists in the FilePreview component that prevents the FileInput component from correctly rendering when a file is uploaded in React Strict Mode. Because Strict Mode requires components to render twice, the mount logic in FilePreview will run twice, causing an error to be throw in relation to the FileReader object being accessed multiple times:
```
Error: An attempt was made to use an object that is not, or is no longer, usable
```

Note that this fix comes directly from the comment from @TomNUSDS (hi Tom!) here https://github.com/trussworks/react-uswds/issues/2433#issuecomment-2105286363

## Related Issues or PRs

Closes #2433 

### Reproduce bug
1. checkout the main branch
2. add react strict mode to storybook configuration. In .storybook/main.ts, `framework` config should look like
```
  framework: {
    name: '@storybook/react-vite',
    options: {
      strictMode: true,
    },
  },
```
3. start storybook with `npm run storybook`
4. navigate to http://localhost:9009/?path=/docs/components-file-input--docs
5. upload a file using the file input
6. _VERIFY_: error message appears (see screenshot)

### Verify fix

1. check out this branch
2. repeat steps 2 - 5 from above
3. _VERIFY_: file uploads successfully, no error message

### Screenshots (optional)
Screenshot 2025-08-13 at 8 34 53 AM


brandonlenz:
@all-contributors please add @TomNUSDS for bug, code

PR #3245, opened 2025-08-25
chore: Add simple-git-hooks and nano-staged to replace husky

### LINKED ISSUE #3155: Urgent: Find and implement alternative for Husky for precommit checks
Husky's requirement to use `pinst` in recent (well, several years now) make it unsuitable for a component library that many consumer use.

Context:
About 3 years ago, we tried and failed to migrate to the newer major version of Husky that requires dependencies, as it introduced problems for consumers of the repo. We reverted and hoped that the many issue reports would result in a better implementation from Husky. Years later, the implementation remains roughly the same, and React USWDS has been updated to use the newer version as documented, which will cause problems for consumers on newer versions of NPM.

https://github.com/typicode/husky/issues/1491#issuecomment-2556404543
https://gist.github.com/djcsdy/3ca078e23fdac4c50e077c84e8284a95

Acceptance criteria:
- [ ] Husky is completely removed as a dependency
- [ ] A suitable replacement is decided on and implemented, running the same checks as currently implemented via husky
- [ ] Optional: Ideally, the solution is JS/TS based, as a founding tenant of this library was to keep the stack entirely JS-based. e.g. precommit is excellent, but husky was chosen over precommit because it is JS based.

abbyoung:
# Summary

This PR adds [simple-git-hooks](https://github.com/toplenboren/simple-git-hooks) and [nano-staged](https://github.com/usmanyunusov/nano-staged) to replace functionality lost when we had to [remove husky.](https://github.com/trussworks/react-uswds/pull/3157)

## Related Issues or PRs

Closes #3155 

## How To Test
Will pair with @brandonlenz to verify working as expected.

* Check out branch and run `yarn install`
* Run `yarn simple-git-hooks` to set up pre-commit


package.json
```diff
@@ -54,7 +54,25 @@
     "prepack": "yarn lint && yarn test && yarn build",
     "happo": "happo run --config ./.happo.js",
     "happo-ci": "HAPPO_CONFIG_FILE='./.happo.js' npx -p happo.io happo-ci-github-actions --config ./.happo.js",
-    "contributors:add": "all-contributors add"
+    "contributors:add": "all-contributors add",
+    "postinstall": "simple-git-hooks",
+    "precommit:checks": "yarn test && yarn test:serverside --silent"
```
brandonlenz, line 59: I have a pretty heavy bias against pre commit or pre push checks that take a long time, especially if CI will do the longer running checks anyways.

In this case, I think we can just do the nano-staged formatting and linting automatically and leave it to CI to run the tests

brandonlenz (approved):
I think this is great. 2 suggestions (which might require me to re-approve, if so, just ping me!)

PR #3262, opened 2025-09-17
feat: Update Accordion expansion behavior

mdmower-csnw:
- When `multiselectable` is not enabled, only expand the last item with `expanded: true` on mount.
- When new items are added to the accordion:
  - If `multiselectable` is enabled, maintain existing expansions and respect `expanded: true` value of new items.
  - If `multiselectable` is not enabled, either maintain the existing expansion if no new items have `expanded: true` or collapse the existing expansion and expand the last new item with `expanded: true`.
- Add and update tests for the updated behavior.

### DIFF
--- a/src/components/Accordion/Accordion.tsx
+++ b/src/components/Accordion/Accordion.tsx
@@ -1,4 +1,4 @@
-import React, { useState, type JSX } from 'react'
+import React, { useEffect, useState, type JSX } from 'react'
 import classnames from 'classnames'
 
 import { HeadingLevel } from '../../types/headingLevel'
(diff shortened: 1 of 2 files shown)


mdmower-csnw:
> I'll follow up later today, hopefully on the simplification. If you already have something for that, feel free to push it up to your PR though.

Sounds good, thanks. Our component is sufficiently different that I don't have changes prepared for immediate post; I'd need to work on a changeset specific to this project. Happy to do so or to let you take the reigns.

mdmower-csnw:
@brandonlenz - following up on the useEffect usage and the key issue:
- Separate PR posted for the key issue: https://github.com/trussworks/react-uswds/pull/3317
- In [e144eec](https://github.com/trussworks/react-uswds/pull/3262/commits/e144eec1ed440f74c5ddc82fee4921a82f0e1009), I replace the `useEffect` hook with a "update state during render" strategy which is recommended as a viable alternative in React documentation: [Adjusting some state when a prop changes](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes) from "You Might Not Need an Effect".

src/components/Accordion/Accordion.tsx
```diff
@@ -62,16 +62,44 @@ export const AccordionItem = ({
...
+) {
+  const lastExpandedItem = items.findLast(
+    (item) => item.expanded || savedExpansions.get(item.id)
+  )
+  return items.reduce((map, item) => {
+    map.set(
+      item.id,
+      !multiselectable
+        ? !!lastExpandedItem && item.id === lastExpandedItem.id
+        : (savedExpansions.get(item.id) ?? !!item.expanded)
```
brandonlenz: nit: I believe we can simplify this thanks to Typescript (even though there was a similar simplification possible in the code before this PR)

```suggestion
        : (savedExpansions.get(item.id) ?? item.expanded)
```

src/components/Accordion/Accordion.tsx
```diff
@@ -62,16 +62,44 @@ export const AccordionItem = ({
...
   )
 
+  // Update saved expansions with new items as the appear
+  const [prevItems, setPrevItems] = useState(items)
+  if (items !== prevItems) {
+    setPrevItems(items)
+    setSavedExpansions((prevExpansions) =>
+      buildExpansions(items, multiselectable, prevExpansions)
+    )
+  }
```
brandonlenz: Nice!

src/components/Accordion/Accordion.tsx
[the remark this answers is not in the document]
mdmower-csnw: Applied in ef58754

Issue #3276, opened 2025-09-30
[fix] Tooltip title not being pass to button


cagosto:
**ReactUSWDS Version & USWDS Version:**

**Describe the bug**

The title prop is not being pass down to the tool tip button.  In your types you define title as a prop that can be passed as seen here 

```
type TooltipProps<T> = {
  label: ReactNode
  title?: string
  position?: 'top' | 'bottom' | 'left' | 'right' | undefined
  wrapperclasses?: string
  className?: string
  children: ReactNode
} &
```

however in reviewing the code i see that title is alway set with an empty string.  As seen here 

```
return (
      // the span that wraps the element will have the tooltip class
      <span data-testid="tooltipWrapper" className={wrapperClasses}>
        <button
          {...remainingProps}
          data-testid="triggerElement"
          ref={triggerElementRef}
          aria-describedby={tooltipID}
          tabIndex={0}
          type="button"
          className={triggerClasses}
          title=""
          onMouseEnter={showTooltip}
          onMouseOver={showTooltip}
          onFocus={showTooltip}
          onMouseLeave={hideTooltip}
          onBlur={hideTooltip}
          onKeyDown={hideTooltip}>
          {children}
        </button>
        <span
          data-testid="tooltipBody"
          title={title ?? (typeof label === 'string' ? label : undefined)}
          id={tooltipID}
          ref={tooltipBodyRef}
          className={tooltipBodyClasses}
          role="tooltip"
          aria-hidden={!isVisible}
          style={positionStyles}>
          {label}
        </span>
      </span>
    )
```

**To Reproduce**

1. Create a tool tip 
2. pass title prop on tool tip something like this 
```
<div className="margin-4">
    <Tooltip label="Default" title="test">Default</Tooltip>
  </div>
```
3. inspect element to see the title on the button 
4. it will show a blank title value `title=""`

**Expected behavior**

The title prop should be pass to the button element and readable by screen readers.

**Device and Browser Information (please complete the following information if describing a UI bug):**

 

All browser all versions

mdmower-csnw:
[...]

PR #3313, opened 2025-11-10
fix: Do not set title on Tooltip trigger element

brandonlenz:
# Summary

## Related Issues or PRs

closes https://github.com/trussworks/react-uswds/issues/3276

## How To Test

1. `yarn install`

2. `yarn storybook`

3. Note that `title` attribute does not exist on trigger elements, matching the USWDS implementation (https://github.com/trussworks/react-uswds/issues/3276#issuecomment-3512522117). Additionally, the USWDS implementation does not set the `title` on the tooltip span either, but we do. I did **_not_** change that, as that would be a breaking change if there are consumers currently using that.

also 

1. `yarn test`

### Screenshots (optional)

Button does not have `title`:
image


brandonlenz:
> Warnings
> ⚠️ 	This PR does not include changes to storybook, even though it affects component code.
> 
> Generated by 🚫 [dangerJS](https://danger.systems/js) against [5790382](https://github.com/trussworks/react-uswds/commit/57903829f057098f99425d22a72fef058261306b)

Not a problem in this case

abbyoung (approved):
LGTM!

PR #3321, opened 2025-11-11
fix: Address eslint react-hook violations in CharacterCount

mdmower-csnw:
# Summary

- Fix eslint react-hooks violation "Calling setState synchronously within an effect can trigger cascading renders" by replacing `useEffect` with state updates during render, as described in [Adjusting some state when a prop changes](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes).
- Set `initialCount` once and don't keep reevaluating it in each render.
- Minor cleanup in blur and change handlers (`value` cannot be undefined for text area and text input.

## How To Test

1. Verify no regressions in storybook CharacterCount stories.
1. The hook lint errors can be seen by temporarily enabling react-hooks linting:
   ```diff
   diff --git a/eslint.config.cjs b/eslint.config.cjs
   index fa6b8bc..a13ab1c 100644
   --- a/eslint.config.cjs
   +++ b/eslint.config.cjs
   @@ -31,6 +31,7 @@ module.exports = defineConfig([
            'plugin:import/warnings',
            'plugin:import/typescript',
            'plugin:react/recommended',
   +        'plugin:react-hooks/recommended',
            'plugin:jsx-a11y/recommended',
            'plugin:security/recommended-legacy',
            'plugin:storybook/recommended',
   ```

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,5 +1,7 @@
 # Summary
 
+<!-- (Optional) Additional details describing the proposed changes -->
+
 ## Related Issues or PRs
 
 <!-- Link existing Github issue(s), e.g. closes #123 -->
@@ -8,6 +10,14 @@
 
 <!-- Describe how a reviewer could test or verify your changes. -->
 
-<!-- Does this change fix an issue or bug in an application you work on? Make sure you've tested this branch in your application to verify it works before merging & releasing. -->
+<!-- Does this change fix an issue or bug in an application you work on? -->
+<!-- Make sure you've tested this branch in your application to verify it works before merging & releasing. -->
 
-### Screenshots (optional)
+## Screenshots (optional)
+
+## Author & Maintainer checklist
+
+- Is this is a [breaking change](https://github.com/trussworks/react-uswds/blob/main/docs/breaking-changes.md)?
+  - [ ] Yes, and I accounted for the breaking changes according to the linked documentation
+  - [ ] No
+- Once merged, add the PR and Issue author(s) as a contributor(s) via the [all-contributors bot](https://allcontributors.org/docs/en/bot/usage#all-contributors-add)
--- a/.github/workflows/semantic_pr.yml
+++ b/.github/workflows/semantic_pr.yml
@@ -8,17 +8,40 @@ on:
       - synchronize
     branches:
       - main
-      - sr-update-semantic-pr-config
 
 jobs:
   lint:
-    name: action-semantic-pull-request
+    name: Validate PR title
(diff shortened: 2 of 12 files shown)


src/components/forms/CharacterCount/CharacterCount.tsx
```diff
@@ -65,7 +65,7 @@ export const CharacterCount = ({
   getMessage = defaultMessage,
   ...remainingProps
 }: TextInputCharacterCountProps | TextareaCharacterCountProps): JSX.Element => {
-  const initialCount = getCharacterCount(value || defaultValue)
+  const [initialCount] = useState(getCharacterCount(value || defaultValue))
```
brandonlenz: I think this is fine, so not blocking, but out of curiosity: Why not use a ref here?

Something like

```suggestion
  const initialCount = useRef(null)
  if (intialCount.current === null) {
    intialCount.current = getCharacterCount(value || defaultValue)
  }
```

Admittedly, it looks way grosser, but is supposedly more semantically correct. [Found some relevant discussion/examples here](https://dev.to/raicuparta/compute-values-on-component-mount-with-react-hooks-state-vs-ref-4epk)

brandonlenz (approved):

src/components/forms/CharacterCount/CharacterCount.tsx
mdmower-csnw: @brandonlenz - Argh, neither of our approaches is correct here. I'm glad you left a comment.

My approach failed at the one improvement it was supposed to achieve: only evaluate `getCharacterCount(value || defaultValue)` once for the purpose of setting `initialCount` during the initial render. Use that value for the next few `useState()` lines and then forget about it. Because I passed `getCharacterCount(value || defaultValue)` as a value to `useState()` rather than through an initializer function, it _still_ gets evaluated every render. My change only achieved adding tiny overhead to each render cycle.

Your approach does better at only evaluating `getCharacterCount(value || defaultValue)` once like intended (kudos for spotting that and not assigning the value as the `useRef()` argument). Unfortunately, it violates another principle: ref values are not supposed to be accessed during renders. Enabling react-hooks recommended eslint ruleset notes this:

> Error: Cannot access refs during render
> 
> React refs are values that are not needed for rendering. Refs should only be accessed outside of render, such as in event handlers or effects. Accessing a ref value (the `current` property) during render can cause your component not to update as expected (https://react.dev/reference/react/useRef)

In 9d33668, I switch to using initializer functions for this `setState` and also for the message `setState`.

Issue #2706, opened 2024-01-17
[docs] Set pattern for storybook controls vs stories


werdnanoslen:
**Does this documentation relate to a specific USWDS feature?**
No, relates to our general storybook use

**Is your documentation related to a problem or inconsistency? Please describe.**
Inconsistency in using and creating stories

**Describe the documentation you'd like**
Let's set a pattern for using storybook controls vs stories. What component features should/can be set by [storybook controls
](https://storybook.js.org/docs/essentials/controls), and when do we split into separate stories. Once we have a solid pattern, let's roll that out to all components for a consistent experience.

**Additional context**
* A good example may be the [text-input component](https://trussworks.github.io/react-uswds/?path=/story/components-text-input--default-text-input), where many of these stories' differences could instead be set by controls, allowing for a better experience in experimenting with how such an input may render
* There've been some significant improvements in controls, which would require an update to our Storybook package

PR #3489, opened 2026-06-03
docs(storybook): Update Storybook Stories to CSF3

### LINKED ISSUE #3490: [feat]: Update to Storybook CSF 3

Storybook stories were written over time as [CSF](https://storybook.js.org/docs/api/csf) evolved. That's why, depending on how long ago they were written, they look different. 

We should unify them all, especially because there is no garauntee that our oldest storybook stories will remain compatible forever.

This will produce a huge diff, but will be easily testable by Happo. Ideally will not require any source code changes, just storybook refactors.

We could continue to do nothing, but this feels like a good way to improve the devX by adhering to the tooling's documentation, and make sure we don't get slapped with broken builds in a future storybook upgrade

brandonlenz:
# Summary

Started this _a while_ ago, only just getting around to finishing up the last few stories and pushing it.

Updates all our storybook examples to use [the more modern storybook syntax](https://storybook.js.org/docs/writing-stories#component-story-format).

While I have mixed feelings about this because "just write React" is something I believe in for clarity of the code, and Storybook's CSF3 tries to get clever, which ultimately just means learning CSF on top of React and USWDS.

That said, CSF Next is looking less verbose, and eventually we have to assume our oldest storybook stories will lose backwards compatibility at some point. 

## Related Issues or PRs

relates to: https://github.com/trussworks/react-uswds/issues/2706
closes: https://github.com/trussworks/react-uswds/issues/3490

## How To Test

This PR will be a success if Happo notices no differences. That's likely the best test. 

Notes:

* ~~Components/Card is showing an errant diff in safari for an image~~ fixed
* [A typo on one of the LanguageSelector stories was fixed, resulting in a diff for that as well](https://github.com/trussworks/react-uswds/pull/3489#discussion_r3350965791)

  - [x] No


src/components/forms/CharacterCount/CharacterCount.tsx
```diff
@@ -46,7 +46,9 @@ type BaseCharacterCountProps = {
 }
 
 export type TextInputCharacterCountProps = BaseCharacterCountProps &
-  TextInputProps
+  // Since CharacterCount provides a default 'type', make it optional instead of required
+  Omit<TextInputProps, 'type'> &
+  Partial<Pick<TextInputProps, 'type'>>
```
brandonlenz, line 51: Flagging a source code change that I _did_ have to make

brandonlenz: This only makes our interface _less strict_, and is not a breaking change because it simply introduces a previously implied default value that strict type checking would not allow you to use.

brandonlenz: Ah right, `type = "text"` is set as an implementation detail. It gets a little tangled in the implementation:

https://github.com/trussworks/react-uswds/blob/5cd3803fe1361a04484879fc95ad8c94481c8643/src/components/forms/CharacterCount/CharacterCount.tsx#L151-L153

I didn't want to mess with that in this already-way-too-big PR.

The prop was already inferred as optional for the purposes of writing the component signature, however, storybooks CSF3 syntax couldn't figure out that the prop was optional unless I explicitly updated the signature as I did in this PR.

So there's no need to update tests, IMO, since they are already written the way we expect them to be

src/components/LanguageSelector/LanguageSelector.stories.tsx
```diff
@@ -22,78 +22,75 @@ Source: https://designsystem.digital.gov/components/language-selector/
...
-}
+    ]
 
-export const LanguagesDisplayMoreThanTwoLanguagesPropSandbox =
-  MoreThanTwoLanguagesTemplate.bind({})
-LanguagesDisplayPropSandbox.args = {
-  displayLang: `en`,
+    return <LanguageSelector langs={languagesDisplayProp} displayLang={lang} />
+  },
 }
```
brandonlenz, line 92: I think this was previously a "bug"/typo, resulting in this storybook story starting in Arabic instead of `en`, so that's the reason for one happo diff.

I believe this line was meant to be `LanguagesDisplayMoreThanTwoLanguagesPropSandbox.args = ...` since  `LanguagesDisplayPropSandbox.args...` is already set near that story above.

This PR updates this story to use `en` as the arg for this story, hence the happo diff.

brandonlenz: In case folks don't have access to Happo: 

image

kleinschmidtj (approved):
Very nice!  Scanned the new code and looks good to me!

Issue #2599, opened 2023-09-29
[feat] Expose refs to every ReactUSWDS component


brandonlenz:
**Is your feature request related to a problem? Please describe.**

Consumers may want to assign and use refs for any number of our components, but only a select few forward refs.

**Describe the solution you'd like**

As a convention, every ReactUSWDS component should forward a ref, so that consumers can interact with them more flexibly given whatever use cases their apps present

**Describe alternatives you've considered**

Our current mixmatched approach is an alternative, but is

brandonlenz:
Something to note:

Components like TextInput and TextArea where the goal is to simply pass a ref to the underyling element wouldn't need to use `forwardRef` in React 19+

Since (at the time of writing) we support 16+, we still need to use forwardRef.

I'm saying this because if/when this gets picked up, our support of React <19 will determine the neatest path forward. It very well might be in the near future that forwardRef becomes unnecessary for many of our basic components (including some like TextInput where it has already been added).

For now, I'm planning to add it to components where it is missing on a case by case basis, as needed. e.g. at the moment I'm working on CharacterCount which leverages TextInput's forwarded ref, but text area will need the same treatment, so I'm adding that.

PR #3492, opened 2026-06-08
feat!: Refactor CharacterCount component for composability & replace legacy TextInput and Textarea `inputRef`s with forwarded `ref`

### LINKED ISSUE #3016: [refactor]: Deprecate and remove TextInput's inputRef property
**Describe the refactor**

Now that https://github.com/trussworks/react-uswds/pull/3011 has been introduced, let's remove the old way of dealing with refs of the TextInput via the inputRef property. 

This is a breaking change, so ideally deprecate in one PR, release, remove in another PR, release again (will be a major version bump)

### LINKED ISSUE #3510: [fix] CharacterCount does not wire up validation styling to label and formgroup when the input's value has exceeded the configured maximum
When a field exceeds its `maxLength`, the count message flips to its invalid state, as does the input, but nothing else does. This is visible in the storybook example (screenshots below)

### Steps to reproduce
1. Render `CharacterCount` with a `maxLength`.
2. Type past the limit.

### Expected
The input, label, and form group reflect the error state (`usa-input--error`, `usa-label--error`, `usa-form-group--error`) and the field reports as invalid, matching the USWDS character count example.

Image

e.g.
Image

### Actual
Only the count message and the input itself updates. The label, and form group keep their default styling.

Image

brandonlenz:
# Summary

Note, this PR started with the intent to remove the legacy refs from TextInput (and Textarea). Unfortunately, that has cascading impacts on CharacterCount, so the changes are bundled for a smoother roll out. That will make the scope of the PR larger than I'd ordinarily prefer.

* Updates the CharacterCount component to be more modular and composable, similar to the #3497
  * This means it is still easily possible to produce the same markup as before, but consumers are able to apply the character count to any custom input (or value) they'd like, instead of being constrained to specific TextInput or Textarea implementations
  * Some of the unit tests were using our oldest patterns with fireEvent. This updates them to use userEvent and covers some more edge cases.
  * HTML CustomValidity logic (triggers on HTML form submit) now allows a custom message which should help with i18n implementations
  * Fixes validation styling not being applied to label and form group on error (https://github.com/trussworks/react-uswds/issues/3510) & the happo diffs reflect that:
    image
    image
* Removes the deprecated `inputRef` prop from TextInput
* Replaces the `inputRef` prop in Textarea with a forwarded ref

> [!Note]
> There are other components that need to be modernized with forwardRef that are still using `LegacyInputRef` props. See #https://github.com/trussworks/react-uswds/issues/2599
> 
> In particular, a good fast follow to this might be a PR that tackles Radio, Checkbox, Select, and RangeInput with similar ref forwarding. These should hopefully not have the same cascading effects

### TextInput

The deprecated `inputRef` prop has been removed. Use the forwarded `ref` instead.

```diff
- <TextInput id="foo" name="foo" type="text" inputRef={ref} />
+ <TextInput id="foo" name="foo" type="text" ref={ref} />
```

### Textarea

`Textarea` now forwards a ref like `TextInput` does. The `inputRef` prop has been removed in favor of `ref`.

```diff
- <Textarea id="foo" name="foo" inputRef={ref} />
+ <Textarea id="foo" name="foo" ref={ref} />
```

### CharacterCount

`CharacterCount` no longer renders an input or textarea for consumers. It is now a presentational component that renders the count messages for any field you own (you could possibly get even more creative with it, but we recommend sticking close to USWDS recommendations). 
Consumers must now render the field themselves, hold its value in state (as a controlled input), and a new `useCharacterCount` hook provides the derived status to drive the messages and error styling.

#### What changed

- `CharacterCount` no longer renders a `TextInput`/`Textarea`.
- The new `useCharacterCount` hook drives the related state. And usage of the component is now opinionated about requiring the use of a controlled input.
- Error styling is now applied by consumers using the CharacterCountStatus return value, e.g. `status.isOverLimit`.
- The `CharacterCount` component's props interface now expects CharacterCountStatus, effectively enforcing hook usage to encourage consumers to set up input control and validation
- Removed type exports: `TextInputCharacterCountProps` and `TextareaCharacterCountProps`. 
- New exports: the `useCharacterCount` hook as well as `CharacterCountProps`, `CharacterCountStatus`, and `UseCharacterCountOptions`.

#### Before (text input)

```tsx
<CharacterCount id="message" name="message" maxLength={500} />
```

#### After (text input)

```tsx
const [value, setValue] = useState('')
const inputRef = useRef<HTMLInputElement>(null)
const status = useCharacterCount({ inputValue: value, maxLength: 500, inputRef })

return (
  <FormGroup error={status.isOverLimit}>
    <Label htmlFor="message" error={status.isOverLimit}>
      Message
    </Label>
    <TextInput
      id="message"
      name="message"
      type="text"
      ref={inputRef}
      value={value}
      onChange={(e) => setValue(e.target.value)}
      validationStatus={status.isOverLimit ? 'error' : undefined}
      aria-describedby="message-info"
    />
    <CharacterCount id="message-info" status={status} />
  </FormGroup>
)
```

#### After (textarea)

Same pattern, swap the field. `Textarea` takes `error` rather than `validationStatus`:

```tsx
const [value, setValue] = useState('')
const textareaRef = useRef<HTMLTextAreaElement>(null)
const status = useCharacterCount({
  inputValue: value,
  maxLength: 500,
  inputRef: textareaRef,
})

return (
  <FormGroup error={status.isOverLimit}>
    <Label htmlFor="message" error={status.isOverLimit}>
      Message
    </Label>
    <Textarea
      id="message"
      name="message"
      ref={textareaRef}
      value={value}
      onChange={(e) => setValue(e.target.value)}
      error={status.isOverLimit}
      aria-describedby="message-info"
    />
    <CharacterCount id="message-info" status={status} />
  </FormGroup>
)
```

#### Notes

- `id` on `CharacterCount` is the id of its message element. Consumers should point their field's `aria-describedby` to that same id, as recommended by USWDS.
- Custom counting and messaging move to the hook: `useCharacterCount({ inputValue, maxLength, getCharacterCount, getMessage })`.
- Pass `inputRef` to the hook (not the component) if you want the field's native validity set when over the limit. Override the message with `customValidityMessage` if needed (defaults to "The content is too long.").
- The hook returns `status` of shape `{ length, maxLength, message, isOverLimit }`.
- Error styling on the input, label, and form group is opt-in. Drive each from `status.isOverLimit` as shown above.

## Related Issues or PRs

closes: https://github.com/trussworks/react-uswds/issues/3016
closes: https://github.com/trussworks/react-uswds/issues/3510

## How To Test

* Unit tests
* Storybook

  - [ ] No


src/components/forms/CharacterCount/CharacterCount.stories.tsx
```diff
@@ -14,6 +16,14 @@ const meta: Meta<typeof CharacterCount> = {
         component: `
 ### USWDS 3.0 Character count component
 
+To use this component, you will need to use the useCharacterCount hook. 
+This hook will provide the CharacterCount component with the necessary information to properly render.
```
brandonlenz, line 20: Reviewers, let me know if you'd like me to include default implementations (similar to `<GovBanner/>` which provides all the individual components as well as a standard implementation for federal .gov websites) for `<TextInputWithCharacterCount/>` and `<TextareaWithCharacterCount/>`. These could be similar to the old code and might be nice conveniences for folks who want the default implementation, but they would be very rigid (and would include the `<FormGroup>` tag so that the proper validation styling could be included, which we did not previously include as part of the component. 

So, still very breaking, but could be a nice add because the React boilerplate is now otherwise fairly lengthy to implement this properly, though I do think its the right choice.

PR #3496, opened 2026-06-09
fix(alert): make headingLevel optional without heading

### LINKED ISSUE #3244: [feat] Make Alert headingLevel prop optional

https://designsystem.digital.gov/components/alert/

I don't know when it was introduced in, but it is present in 2.11.0

It is possible to not provide a heading to the Alert component, see screenshots from USWDS below. The implemented component accordingly has `heading` as an [optional prop](https://github.com/trussworks/react-uswds/blob/main/src/components/Alert/Alert.tsx#L10) that [conditionally renders a heading](https://github.com/trussworks/react-uswds/blob/main/src/components/Alert/Alert.tsx#L10) depending on whether it is present. However, the `headingLevel` prop [is required](https://github.com/trussworks/react-uswds/blob/main/src/components/Alert/Alert.tsx#L11) regardless of whether the heading prop is provided.

This leads to situations where I need to decide and pass in an arbitrary `headingLevel` prop, despite the Alert not having a heading.

Image

It should not be required to pass `headingLevel` when `heading` it is not present.

For stronger type enforcement, a typescript union type could be used to enforce that if heading is present then headingLevel must also be (but probably not vice versa, for backwards compatibility?). Another option would be to just make the `headingLevel` prop optional (presumably throwing an error if it's not present?).

**Additional context**

YforC:
# Summary

Make `Alert` require `headingLevel` only when a `heading` is provided. Headingless alerts already render with the component's default heading level internally, so consumers should not need to pass a heading level for alerts that do not render a heading.

The stories and tests now cover headingless alerts without `headingLevel`, matching the intended public API.

## Related Issues or PRs

Closes #3244

## How To Test

- `corepack yarn vitest --run src/components/Alert/Alert.test.tsx`
- `corepack yarn tsc --noEmit`

## Screenshots (optional)

N/A

  - [x] No

### DIFF
--- a/src/components/Alert/Alert.stories.tsx
+++ b/src/components/Alert/Alert.stories.tsx
@@ -65,16 +65,16 @@ export const Info: Story = {
 export const Slim: Story = {
   render: () => (
     <>
-      <Alert type="success" headingLevel="h4" slim>
+      <Alert type="success" slim>
         {testText}
       </Alert>
-      <Alert type="warning" headingLevel="h4" slim>
+      <Alert type="warning" slim>
         {testText}
       </Alert>
-      <Alert type="error" headingLevel="h4" slim>
+      <Alert type="error" slim>
         {testText}
       </Alert>
-      <Alert type="info" headingLevel="h4" slim>
+      <Alert type="info" slim>
         {testText}
       </Alert>
     </>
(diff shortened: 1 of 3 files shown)


brandonlenz:
@YforC thanks for this! Just as a heads up, I've got a major (breaking) refactor of the Alert component.

I started with almost the exact changes that you're proposing here (except using `never` for headingLevel in the event that `heading` was not defined), but decided (since the TS contract would become breaking anyways) that a refactor to align with the goals of more flexibly composability would be good timing.

Let me know what your thoughts are on https://github.com/trussworks/react-uswds/pull/3497

PR #3498, opened 2026-06-09
feat!: Upgrade Node.js dependencies (Actions and Supported versions)

### LINKED ISSUE #3458: [feat]: Update deprecated Actions runners
# Summary

Currently, actions are running on deprecated node versions. These should be updated before the expiration date 

## Example

From https://github.com/trussworks/react-uswds/actions/runs/24145703338:

Image

brandonlenz:
# Summary

- **Breaking change**: Drops support for Node 20 (EOL)
- Updates the development version to Node 26
- Updates actions to use v5 Node.js actions since v4 actions will be EOL by September.

TODO: Once approved, in order to merge, update repository settings to 
- [x] require Node 26 tests to pass
- [x] remove the requirement for Node 20 tests to pass

See also: https://nodejs.org/en/about/previous-releases

## Related Issues or PRs

closes: https://github.com/trussworks/react-uswds/issues/3458

## How To Test

- Install node 26 locally, run tests and storybook locally
- Verify CI checks using v5 runners pass

## Screenshots (optional)

image

image

  - [ ] No


brandonlenz:
TODO: Once approved, in order to merge, update repository settings to 
- require Node 26 tests to pass
- remove the requirement for Node 20 tests to pass

brandonlenz:
@jsclarridge can I tag you in for review on this one?

jsclarridge:
> @jsclarridge can I tag you in for review on this one?

As an incremental update to address the GitHub Actions versions that will be EOL by September, this looks good, although I think the use of `actions/cache@v4` in `.github/workflows/publish-next.yml` also needs to be bumped to at least v5.

brandonlenz:
> As an incremental update to address the GitHub Actions versions that will be EOL by September, this looks good, although I think the use of `actions/cache@v4` in `.github/workflows/publish-next.yml` also needs to be bumped to at least v5.

On this repo we've not historically pinned by sha, but I'm thinking that we probably should, so I can make that change while addressing this comment too

.github/dependabot.yml
```diff
@@ -60,3 +60,17 @@ updates:
...
+      separator: '-'
+    schedule:
+      interval: "weekly"
+    labels:
+      - 'type: dependencies'
+      - 'type: automerge'
+    groups:
+      github-actions:
+        patterns:
+          - '*'
```
brandonlenz: I actually didn't know Dependabot handled sha-style actions dependency management, but it does: https://docs.github.com/en/code-security/reference/supply-chain-security/supported-ecosystems-and-repositories#github-actions

> Dependabot updates the version documentation of GitHub Actions when the comment is on the same line, such as `actions/checkout@<commit> #<tag or link>` or `actions/checkout@<tag> #<tag or link>`.

jsclarridge: I suggest leaving the Dependabot config as-is for now and create a separate follow-up issue to possibly make this change later. If we want Dependabot to manage the GitHub Actions updates, I'm not sure if we want them to be automerged. And we should probably add a cooldown period so that Dependabot would wait a certain amount of time before opening a PR to update the actions.

jsclarridge:
> > As an incremental update to address the GitHub Actions versions that will be EOL by September, this looks good, although I think the use of `actions/cache@v4` in `.github/workflows/publish-next.yml` also needs to be bumped to at least v5.
> 
> On this repo we've not historically pinned by sha, but I'm thinking that we probably should, so I can make that change while addressing this comment too

The version pin changes look good to me, but I left a comment/suggestion about the Dependabot config update.

.github/dependabot.yml
brandonlenz: Fair enough, I'll make the change!

For the cooldown, I opted for a [weekly cadence](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#schedule-) with `schedule.interval: weekly`.

Or am I misunderstanding what you mean?

brandonlenz: https://github.com/trussworks/react-uswds/pull/3498/commits/4eb0c25ad72ff696083b3c410da09c4791834c72

brandonlenz:
Oh fun, the npm endpoint that Yarn 3 relies on is being retired. It now returns a 410 instead of the json that Dangerfile.ts expects to parse:

```
yarn npm audit
➤ YN0035: This endpoint is being retired. Use the bulk advisory endpoint instead. See the following docs for more info: https://api-docs.npmjs.com/#tag/Audit
➤ YN0035:   Response Code: 410 (Gone)
➤ YN0035:   Request Method: POST
➤ YN0035:   Request URL: https://registry.yarnpkg.com/-/npm/v1/security/audits/quick

➤ Errors happened when preparing the environment required to run this command.
➤ This might be caused by packages being missing from the lockfile, in which case running "yarn install" might help.
```

I'm going to bandaid it here with some comments pointing to the npm migration PR which will take care of it for good.

.github/dependabot.yml
jsclarridge: I was referring to the [cooldown](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-) option, which can be used in addition to `schedule.interval`. According to the documentation, there's a default cooldown of 3 days even if `cooldown` isn't set, but it may still be a good idea to set it to explicitly, and we may want a slightly longer cooldown, e.g., 7 days.

jsclarridge (approved):

PR #3499, opened 2026-06-09
fix(a11y): Set Pagination's last page aria-label to indicate that it is the last page

### LINKED ISSUE #3093: [fix] Add the ability for screen reader users to know that a page is the last page when navigating the Pagination component
```
    "@trussworks/react-uswds": "^9.1.0",
    "@uswds/uswds": "3.11.0",
```

**Describe the bug**
USWDS's [accessibility guidance](https://designsystem.digital.gov/components/pagination/#accessibility-guidance) on the Pagination component recommends the following:
```
Voice “last page” on the last page’s link in a bounded set. Use aria-label="last page, page [#]" to voice that the last element in the navigation is the last page in the paginated section. Don’t include this on the last navigation element in an unbounded set, since there is no last page in an unbounded set.
```

**To Reproduce**

1. Implement the pagination component
2. Notice that the last page link does not have an aria-label that tells the user it's the last page
3. Link to implementation: https://trussworks.github.io/react-uswds/?path=/docs/components-pagination--docs

**Expected behavior**
The last page link needs to have an aria-label that tells the user it's the last page.

**Screenshots**

Image

**Additional context**

- Mac OS
- Chrome 132.0.6834.160

brandonlenz:
# Summary

Sets the aria-label of the last page to `Last page, page ${the_page_number}`

Side note: This component could use a lot of TLC to make it usable with a translation framework. I did not help with that in this PR, but will create an issue for the backlog.

## Related Issues or PRs

closes: https://github.com/trussworks/react-uswds/issues/3093

## How To Test

- Unit tests confirm this behavior
- Can also verify in storybook on this branch

## Screenshots (optional)

image

  - [x] No

### DIFF
--- a/src/components/Pagination/Pagination.test.tsx
+++ b/src/components/Pagination/Pagination.test.tsx
@@ -4,16 +4,15 @@ import React from 'react'
 import { Pagination } from './Pagination'
 
 describe('Pagination component', () => {
-  const testPages = 24
-  const testThreePages = 3
-  const testSevenPages = 7
+  const totalPages = 24
   const testPathname = '/test-pathname'
 
   it('renders pagination for a list of pages', () => {
+    const currentPageNumber = 10
     render(
       <Pagination
-        totalPages={testPages}
-        currentPage={10}
+        totalPages={totalPages}
+        currentPage={currentPageNumber}
         pathname={testPathname}
       />
     )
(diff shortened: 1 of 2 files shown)


src/components/Pagination/Pagination.test.tsx
brandonlenz, line 1: Some minor test refactoring can be found in this file

src/components/Pagination/Pagination.tsx
```diff
@@ -16,20 +16,23 @@ export type PaginationProps = {
...
-  page,
-  isCurrent,
-  pathname,
-  onClickPageNumber,
-}: {
+type PaginationPageProps = {
   pathname: string
   page: number
   isCurrent?: boolean
+  isLastPage: boolean
```
ronaktruss: A little outside the scope of this ticket, but shouldn't these be identical? (Shouldn't isCurrent also be required/ both of them be optional?)

ronaktruss (approved):
LGTM!

src/components/Pagination/Pagination.tsx
brandonlenz: ooo probably. I'll take a look before merging. Since this type is internal only, it wouldn't be breaking.

For things that are always either true or false, I do prefer the typed value to reflect that, and isCurrent certainly could.

brandonlenz: Yea, the prop was already always passed in with a defined value. Good easy change. Nice catch!

https://github.com/trussworks/react-uswds/pull/3499/commits/779c5f90c1f9a96360197e860db7d3fdcbffcdd6

Issue #3324, opened 2025-11-13
[fix] CI to publish to GitHub Next Packages fails due to authentication issue


brandonlenz:
## ReactUSWDS Version & USWDS Version

N/A

## Describe the bug

For quite some time, our packages have not been getting published to GitHub Packages:

https://github.com/orgs/trussworks/packages/npm/package/react-uswds

The workflow responsible for this is here: https://github.com/trussworks/react-uswds/actions/workflows/publish-next.yml

The failed jobs indicate attempts to publish that fail due to `➤ YN0033: No authentication configured for request`

The successful jobs just skip the publish step due to `YN0000: Registry already knows about version <X.X.X>; skipping.`

So this workflow is not publishing at all anymore.

## Expected behavior

Merges to main should publish to GitHub packages successfully.

## Additional context

Without having found confirmation, it looks like the `GITHUB_TOKEN` we were using to is no longer valid for publishing. Instead, [GitHub says](https://docs.github.com/en/packages/learn-github-packages/publishing-a-package#publishing-a-package):

> GitHub Packages only supports authentication using a **personal access token (classic)**.

At one point, this was working, so I can only assume this was a GitHub security change.

## Device and Browser Information (please complete the following information if describing a UI bug)

N/A

PR #3500, opened 2026-06-12
feat(package-manager)!: Use npm instead of yarn

### LINKED ISSUE #3156: Move from `yarn` to `npm`
Context:

At one point in time, `yarn` was a net improvement over `npm`, so it was a good choice for this repository.

`npm` has made strides over the years where performance concerns are no longer a driver to remain on `yarn`

`yarn` versions and corresponding documentation feels disorganized

`npm` is the standard tool from which all others take inspiration, and therefore it is most guaranteed to retain "backwards" compatibility for consumers using other tools. Since the main focus of React USWDS is compatibility for adoption, `npm` is a logical choice.

`yarn` limitations have contributed to issues such as the Husky blockers for releases (https://github.com/trussworks/react-uswds/issues/3155)

Acceptance criteria:

- [ ] React USWDS uses `npm` instead of `yarn`
- [ ] All yarn-specific scripts are replaced with `npm` equivalents
- [ ] Some dangerfile scripting depends on yarn specific output. It should be updated to use the npm equivalent

brandonlenz:
# Summary

I recommend starting with the[ ADR (on this branch)](https://github.com/trussworks/react-uswds/blob/bl-use-npm-instead-of-yarn/docs/adr/0002-use-npm-instead-of-yarn.md)

I know this looks scary:
image

The overwhelming majority of that is `yarn.lock` being deleted and `package-lock.json` being added.

## Related Issues or PRs

closes: #3156 

## How To Test

One of the ways I tested was comparing the output of `yarn pack` and `npm pack` which helped identify the slight difference in the tarball naming, and the need to manually include `CHANGELOG.md` in the package output to maintain parity.

  - [x] Yes, but also no. This _shouldn't_ break things for anyone, but we've had surprising package manager conflicts in the past if consumer tooling didn't also match what we were using (yarn). So out of an abundance of caution, I'm treating this as a breaking change.
     - In the past, when someone using `npm` ran into issues because of a side effect of us using `yarn`, it was this library's problem. Going forward, it will be the other way around (but also hopefully be better for consumers of the library on average)
  - [ ] No


.github/workflows/publish-next.yml
brandonlenz, line 1: This action only runs when code lands on main. Additionally, it's been broken for some time. There's an open issue for that here: #3324 

Just for visibility, I won't invest any effort in making it work or triaging a failure in that pipeline after this PR until picking that issues up.

package.json
```diff
@@ -93,7 +99,7 @@
     "@babel/core": "^7.23.9",
     "@eslint/compat": "^2.0.0",
     "@eslint/eslintrc": "^3.3.0",
-    "@eslint/js": "^10.0.1",
+    "@eslint/js": "^9.24.0",
```
brandonlenz, line 102: Dependapot previously bumped `@eslint/js` on its own causing a peer dependency violation.

Yarn treats those as warnings, so we didn't notice it. npm is more strict. 

In this PR, we drop the version back and update the dependabot config to include this dependency in the eslint group to make sure that they all get bumped together as intended.

eslint.config.cjs
brandonlenz, line 1: Excuse the diff. Did a prettier format of this file.

brandonlenz: Actual changes for linting dangerfile can be found at the bottom

docs/adr/0002-use-npm-instead-of-yarn.md
```diff
@@ -0,0 +1,78 @@
...
+- Date: 2026-06-12
+
+## Context and Problem Statement
+
+ReactUSWDS has used yarn since the project's initial setup in 2019.
+At that time, yarn held real advantages over npm v6: faster installs, a deterministic offline cache, and workspaces.
+npm has since closed those gaps, and install performance is now comparable for a project of this size.
+
+Meanwhile, using a non-default package manager carries recurring costs for a library that is published to and overwhelmingly consumed from the npm registry.
+Which package manager should this project use going forward?
```
ronaktruss, line 15: Nit: The problem statement is hard to find when skimming. Can you add another empty line so the `Which package manager should this project use going forward?` gets displayed on its own line?

.github/workflows/publish-next.yml
```diff
@@ -15,35 +15,35 @@ jobs:
...
-
-      # Setup .npmrc file to publish to GitHub Packages
-      - name: Set up node
-        uses: actions/setup-node@v4
-        with:
-          registry-url: 'https://npm.pkg.github.com'
-          node-version-file: '.node-version'
+        run: npm ci
+
+      # npm has no equivalent of yarn's --tolerate-republish, so skip
```
ronaktruss, line 29: Is there a way to test this? This looks good based on reading it though.

ronaktruss (approved):
Overall looks good to me! There are some newer conflicts. Nice catch on the tar file name changes and other minor fixes added in here.

With such a large number of changes I haven't tested everything with my typical level of thoroughness but having read through the changes everything seemed fine.

.github/workflows/publish-next.yml
brandonlenz: I think we'll have to resolve https://github.com/trussworks/react-uswds/issues/3324 in order to test, because I suspect this job will still fail even with these changes.

I definitely had to take a real close look on these changes, because I believe the intention was for `--tolerate-republish` to actually publish an artifact with the same version for every non-release commit to main to the Github registry. 

But I don't think that's how that works. [The github registry's versioned artifacts are immutable. `--tolerate-republish` just fails to publish to github.](https://github.com/yarnpkg/berry/discussions/4770#discussioncomment-3416639)

So really, this _could_ be refactored into an action that only runs when a release PR merges, and it would have the same effect, and likely be more efficient.

But I'm preserving parity here as much as possible and leaving all of that up to https://github.com/trussworks/react-uswds/issues/3324 for more singular focus

docs/adr/0002-use-npm-instead-of-yarn.md
brandonlenz: Good call, will do!

brandonlenz: https://github.com/trussworks/react-uswds/pull/3500/commits/1d991a99e9f8f49639d4fc098415dfcd34fc4e2a

brandonlenz:
Current CI errors are from npm ci failing where yarn would have warned on peer dependency mismatches (new after merging updates from main). Looking in to it

brandonlenz:
> Current CI errors are from npm ci failing where yarn would have warned on peer dependency mismatches (new after merging updates from main). Looking in to it

Ah yea, that's going to be expected until this merges. Weekly, dependabot will try to update eslint in a way that yarn is more tolerant of, but npm blocks. 

I prefer npm's approach where if the packages declare that they are not intended to work together (via their peer ranges), the build fails. Louder, but preserves stability.

https://github.com/trussworks/react-uswds/pull/3500/commits/d0f1ea9810323b471982420340c0cbadd787cc1b

Issue #2947, opened 2024-06-12
[feat] The TimePicker component should support 24 hour time formats


dakota002:
**Is your feature request related to a problem? Please describe.**

In some fields, like astronomy, a 24 hour time format is more common/useful. The TimePicker component currently only supports the 12 hour format.

**Describe the solution you'd like**

An optional property on the component, something like `format`, with options 12 and 24, defaulting to 12.

PR #3506, opened 2026-06-16
feat: Add 24h view for time picker

ronaktruss:
# Summary

Adds optional property to enable military time

## Related Issues or PRs

#2947

## How To Test

Look at time picker in storybook and review the 24h option.

## Screenshots (optional)
image

  - [x] No

### DIFF
--- a/src/components/forms/TimePicker/TimePicker.tsx
+++ b/src/components/forms/TimePicker/TimePicker.tsx
@@ -29,7 +29,7 @@ type BaseTimePickerProps = {
   minTime?: string
   maxTime?: string
   step?: number
-  /** Display times in 12-hour (e.g. "1:00pm") or 24-hour military format (e.g. "13:00"). Defaults to "12h". */
+  /** Display times in 12-hour (e.g. "1:00pm") or 24-hour format (e.g. "13:00"). Defaults to "12h". */
   format?: TimePickerFormat
   /** Recommended text: "Select a time from the dropdown. Type into the input to filter options." */
   hint?: string
--- a/src/components/forms/TimePicker/TimePicker.tsx
+++ b/src/components/forms/TimePicker/TimePicker.tsx
@@ -10,9 +10,14 @@ import {
   DEFAULT_MIN_TIME,
   DEFAULT_MIN_TIME_MINUTES,
   DEFAULT_STEP,
+  DEFAULT_TIME_FORMAT,
   MIN_STEP,
   TIME_PICKER_CUSTOM_FILTER,
+  TIME_PICKER_CUSTOM_FILTER_24H,
+  TimePickerFormat,
 } from './constants'
+
+export type { TimePickerFormat }
 
 type BaseTimePickerProps = {
   id: string
@@ -24,6 +29,8 @@ type BaseTimePickerProps = {
   minTime?: string
   maxTime?: string
   step?: number
+  /** Display times in 12-hour (e.g. "1:00pm") or 24-hour military format (e.g. "13:00"). Defaults to "12h". */
+  format?: TimePickerFormat
   /** Recommended text: "Select a time from the dropdown. Type into the input to filter options." */
   hint?: string
   className?: string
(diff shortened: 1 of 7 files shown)


src/components/forms/TimePicker/constants.ts
```diff
@@ -16,3 +19,11 @@ export const TIME_PICKER_CUSTOM_FILTER: CustomizableFilter = {
     minuteQueryFilter: '[\\d]+:([0-9]{0,2})',
   },
 }
+
+export const TIME_PICKER_CUSTOM_FILTER_24H: CustomizableFilter = {
+  filter: '0?{{ hourQueryFilter }}:{{ minuteQueryFilter }}.*',
+  extras: {
+    hourQueryFilter: '([0-9]{1,2})',
```
ronaktruss, line 25: We could have more specific regex here, but made it as similar to the other TIME_PICKER_CUSTOM_FILTER so that they behave as similar as possible.

brandonlenz: I think that's the right move. I haven't double checked, but I'm fairly certain a lot of these supporting constants and logic are lifted almost 1:1 from USWDS, so keeping it "similar" is exactly the right call.

src/components/forms/TimePicker/constants.ts
```diff
@@ -1,5 +1,8 @@
 import { CustomizableFilter } from '../ComboBox/ComboBox'
 
+export type TimePickerFormat = '12h' | '24h'
```
brandonlenz: Not blocking, but I would move this into TimePicker.tsx, personally. 

We generally try to keep types (for prop signatures at least) adjacent to the components that use them.

src/components/forms/TimePicker/TimePicker.tsx
```diff
@@ -24,6 +29,8 @@ type BaseTimePickerProps = {
   minTime?: string
   maxTime?: string
   step?: number
+  /** Display times in 12-hour (e.g. "1:00pm") or 24-hour format (e.g. "13:00"). Defaults to "12h". */
+  format?: TimePickerFormat
```
brandonlenz: Not blocking, but you could skip the comment if you wanted to, IMO, since TimePickerFormat is nicely typed.

Especially if it is documented in storybook (which you did, thank you!) and if the TimePickerFormat is defined close by

src/components/forms/TimePicker/utils.ts
```diff
@@ -63,10 +65,12 @@ export const getTimeOptions = (
...
   ) {
     const { minute, hour24, hour12, ampm } = getTimeContext(minutes)
+    const value = `${padZeros(hour24, 2)}:${padZeros(minute, 2)}`
 
     timeOptions.push({
-      value: `${padZeros(hour24, 2)}:${padZeros(minute, 2)}`,
-      label: `${hour12}:${padZeros(minute, 2)}${ampm}`,
+      value,
+      label:
+        format === '24h' ? value : `${hour12}:${padZeros(minute, 2)}${ampm}`,
```
brandonlenz: nit: Since you pulled `value` out, I recommend doing the same for `label` here, for consistency


```suggestion
    const value = `${padZeros(hour24, 2)}:${padZeros(minute, 2)}`
    const label = format === '24h' ? value : `${hour12}:${padZeros(minute, 2)}${ampm}`

timeOptions.push({
      value,
      label,
```

Or leaving them _both_ inline. 

Not blocking

brandonlenz (approved):
Thanks for the storybook example! Made it easy to test.

This was a nice, surgical change. Thanks!

src/components/forms/TimePicker/TimePicker.tsx
```diff
@@ -10,10 +10,15 @@ import {
...
   DEFAULT_MIN_TIME_MINUTES,
   DEFAULT_STEP,
+  DEFAULT_TIME_FORMAT,
   MIN_STEP,
   TIME_PICKER_CUSTOM_FILTER,
+  TIME_PICKER_CUSTOM_FILTER_24H,
+  TimePickerFormat,
 } from './constants'
 
+export type { TimePickerFormat }
```
brandonlenz: Oh yea, I forgot to say, thanks for exporting this in index.tsx. 

Another reason to move it to this file (rather than putting it in constants) is this line here. It seems odd to import it, to then just export it from here when we could export it inline where we define it in this file

brandonlenz:
@all-contributors please add @ronaktruss for code

PR #3528, opened 2026-07-02
fix: Close language selector dropdown on selection or outside click

abbyoung:
# Summary
❗Note to reviewers: Code in this PR was written using Codex.

Fixes a bug in the language selector dropdown menu that's used when more than two languages are present.

* Language selector dropdown menu closes when a user clicks outside the menu
* Language selector dropdown menu closes when a user clicks a menu item
* Adds tests

## Related Issues or PRs

Closes #3170

## How To Test

1. Run storybook for this repo `yarn storybook`
2. Review the story `MoreThanTwoLanguagesAsALink` and confirm behavior
    * Click `Language` link > click outside the menu. It should close.
    * Click `Language` link > click a language. You will be redirected (existing behavior).
    * In the new page, click the `Language` link again > click a language > Confirm the menu closes as expected.
4. Review the story `MoreThan Two LanguagesAsAButton` and confirm behavior
     * Click `Language` link > click outside the menu. It should close.
     * Click `Language` link > click a language > menu should close.
5. Review the story `Custom Class` and confirm behavior.
     * Click `Language` link > click outside the menu. It should close.
     * Click `Language` link > click a language > menu should close.
     
## Author & Maintainer checklist

- Is this is a [breaking change](https://github.com/trussworks/react-uswds/blob/main/docs/breaking-changes.md)?
  - [ ] Yes, and I accounted for the breaking changes according to the linked documentation
  - [x] No
- Once merged, add the PR and Issue author(s) as a contributor(s) via the [all-contributors bot](https://allcontributors.org/docs/en/bot/usage#all-contributors-add)

PR #3534, opened 2026-07-08
feat: Close Language Selector on pressing escape and on blur

brandonlenz:
# Summary

Was reviewing https://github.com/trussworks/react-uswds/pull/3528, which works and looks great. It did cause me to deep dive on the USWDS implementation, which I noticed has additional close conditions we were not capturing:
* Close on pressing escape
* Close onBlur

So I added those.

There was also weird behavior in Safari that I just _happened_ to stumble on. Safari does not focus the element on click, and our click listener to close the modal was pinned to the container div (which spans horizontally across the screen). So in Safari, if you were to click next to the language selector, it would not close. The [USWDS implementation](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/iframe.html?id=components-language-selector--three-or-more-languages&args=&viewMode=story) works in Safari just fine, so I updated the click event listener to match their implementation more closely and look for clicks outside of the button itself.

While updating the tests, also moved us from fireEvent to userEvent to take care of a little tech debt there.

Finally, the language-options dropdown previously had a static `id`, meaning that if (for whatever reason) someone included 2 language selectors on a single page, they'd run into some potential issues not being able to set unique ids. Now, if an id is passed to the LanguageSelector, the dropdown will use the id as a prefix.

## Related Issues or PRs

https://github.com/trussworks/react-uswds/pull/3528

## How To Test

Added unit tests

Tested via storybook. Importantly, in Safari

  - [x] No (arguably the `id` change _could_ be in rare circumstances where someone was passing an id and relying on the hardcoded dropdown options `id`, but I'm going to make a judgment call and call this non-breaking still.

### DIFF
--- a/src/components/LanguageSelector/LanguageSelector.test.tsx
+++ b/src/components/LanguageSelector/LanguageSelector.test.tsx
@@ -1,10 +1,7 @@
 import React from 'react'
 import { fireEvent, render, waitFor } from '@testing-library/react'
 import { userEvent } from '@testing-library/user-event'
-import {
-  LanguageSelector,
-  LanguageDefinition,
-} from '../LanguageSelector/LanguageSelector'
+import { LanguageSelector, LanguageDefinition } from './LanguageSelector'
 
 const voidLink = '#test'
 const languages: LanguageDefinition[] = [
@@ -373,6 +370,37 @@ describe('LanguageSelector component', () => {
         'false'
       )
       expect(getByTestId(languages[0].attr)).not.toBeVisible()
+    })
+
+    it('links the button to the menu with the default id', () => {
+      const { getByTestId } = render(
+        <LanguageSelector langs={languages} label="Languages" />
+      )
+
+      expect(getByTestId('languageSelectorButton')).toHaveAttribute(
+        'aria-controls',
+        'language-options'
+      )
+      expect(document.getElementById('language-options')).toBeInTheDocument()
+    })
+
+    it('derives a unique menu id from a given id', () => {
+      const { getByTestId } = render(
+        <LanguageSelector
+          langs={languages}
+          label="Languages"
+          id="my-custom-id"
+        />
+      )
+
+      expect(getByTestId('languageSelector')).toHaveAttribute(
+        'id',
+        'my-custom-id'
+      )
+      expect(getByTestId('languageSelectorButton')).toHaveAttribute(
+        'aria-controls',
+        'my-custom-id-language-options'
+      )
     })
 
     it('closes an open selector when another selector is opened', async () => {
--- a/src/components/LanguageSelector/LanguageSelector.test.tsx
+++ b/src/components/LanguageSelector/LanguageSelector.test.tsx
@@ -1,5 +1,6 @@
 import React from 'react'
 import { fireEvent, render, waitFor } from '@testing-library/react'
+import { userEvent } from '@testing-library/user-event'
 import {
   LanguageSelector,
   LanguageDefinition,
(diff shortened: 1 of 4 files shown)


src/components/LanguageSelector/LanguageSelector.test.tsx
```diff
@@ -227,9 +235,134 @@ describe('LanguageSelector component', () => {
...
+      await user.click(getByTestId('languageSelectorButton'))
+
+      expect(getByTestId('languageSelectorButton')).toHaveAttribute(
+        'aria-expanded',
+        'false'
+      )
+      expect(getByTestId(languages[0].attr)).not.toBeVisible()
+    })
+
+    it('closes the list when an outside click handler stops propagation', async () => {
```
brandonlenz, line 345: This test validates and confirms the `{ capture: true }` addition to the click event listener

src/components/LanguageSelector/LanguageSelectorDropdown.tsx
```diff
@@ -47,56 +46,72 @@ const generateMenuItems = (
...
     }
   }, [isOpen])
 
-  const classes = classnames(
-    'usa-language-container',
-    {
-      [`usa-language--small`]: small !== undefined,
-    },
-    className
-  )
```
brandonlenz, line 83: This was previously repeated twice. Now just passing the classes in from LanguageSelector

src/components/LanguageSelector/LanguageSelectorDropdown.tsx
```diff
@@ -47,56 +46,72 @@ const generateMenuItems = (
...
   return (
+    // eslint-disable-next-line jsx-a11y/no-static-element-interactions
     <div
-      className={classes}
+      id={id}
+      className={className}
       data-testid="languageSelector"
       ref={containerRef}
+      onBlur={closeOnFocusOut}
+      onKeyDown={closeOnEscape}
```
brandonlenz, line 108: For the additional ways that the USWDS LanguageSelector closes

src/components/LanguageSelector/LanguageSelectorDropdown.tsx
```diff
@@ -47,56 +46,72 @@ const generateMenuItems = (
...
+  const closeOnEscape = (event: React.KeyboardEvent<HTMLDivElement>): void => {
+    if (isOpen && event.key === 'Escape') {
+      buttonRef.current?.focus()
+      setIsOpen(false)
+    }
+  }
+
   const displayLabel = langs.find((langDef) => langDef.attr === displayLang)
-  const menuID = 'language-options'
+  const menuID = id ? `${id}-language-options` : 'language-options'
```
ronaktruss, line 95: Nit: Is this actually needed? Is this logic worth having?

src/components/LanguageSelector/LanguageSelectorDropdown.tsx
```diff
@@ -47,56 +46,72 @@ const generateMenuItems = (
...
     setIsOpen(false)
   }
 
   return (
+    // eslint-disable-next-line jsx-a11y/no-static-element-interactions
     <div
-      className={classes}
+      id={id}
+      className={className}
       data-testid="languageSelector"
```
ronaktruss, line 105: Should this also be dynamic similar to the menu id?

src/components/LanguageSelector/LanguageSelector.test.tsx
```diff
@@ -227,9 +235,134 @@ describe('LanguageSelector component', () => {
...
+        <LanguageSelector langs={languages} label="Languages" />
+      )
+
+      await user.click(getByTestId('languageSelectorButton'))
       expect(getByTestId(languages[0].attr)).toBeVisible()
 
+      // fireEvent, not userEvent: dispatch a bare click with no focus
+      // movement, as in Safari, so the focusout handler cannot close the
+      // menu and mask a regression in the click listener
+      fireEvent.click(getByTestId('languageSelector'))
```
ronaktruss, line 262: Nice!

ronaktruss (approved):
These changes look good. Nice catch on the previous closeOnOutsideClick not working when clicking next to the language selector. It was also an issue on Firefox.

src/components/LanguageSelector/LanguageSelectorDropdown.tsx
brandonlenz: I think so.

If, for some reason, you were to put more than one language selector on a page (maybe one in the header, one in the footer) you'd run in to accessibility tooling issues by having duplicate hard-coded ids.

This lets consumers of the library pass in an id to avoid that problem

src/components/LanguageSelector/LanguageSelectorDropdown.tsx
brandonlenz: Gave this some thought.

We certainly _can_, but I lean no, purely because these are just used for our own tests and aren't design for consumers to use/alter. This would expand the PR a little because then I'd go down a rabbit hole of prefixing the data-testid's of the child elements as well, which is a whole thing.

If someone asks for it, or we otherwise develop a need for it, sure.

PR #3571, opened 2026-08-10
fix(ci): wait for image decoding before Happo snapshots Card stories

### LINKED ISSUE #3509: fix(ci): Happo frequently finds errant diff on the Card stories in Safari
**Describe the bug**

The images on this story are consistently flagged as diffs in Happo. We should look into our options for making happo less sensitive regarding these images:

Image

**To Reproduce**

1. Go to https://github.com/trussworks/react-uswds/pull/3508 (or most other recent PRs)
2. Click on the Happo CI step to view the diff
3. See what I posted in the screenshot above

**Expected behavior**

Should not flag false positives so frequently (ideally never, but understandable that the tool will occasionally make changes or browsers will make changes that will produce false positives occasionally)

**Additional context**

Start here, probably: https://docs.happo.io/docs/spurious-diffs

Munan059:
Happo intermittently flags the Card story images as diffs in Safari and iOS Safari (see #3509). The Card stories render base64 JPEG data URIs, whose `load` event fires synchronously, so Happo's built-in image wait is satisfied before Safari has actually finished decoding the JPEG pixels. That race produces flaky diffs.

This uses Happo's documented `beforeScreenshot` Storybook hook together with `HTMLImageElement.decode()` to deterministically wait for every `` on the page to be decoded before the snapshot is taken.

Fixes #3509


.happo.js
```diff
@@ -16,9 +16,11 @@ export default {
     }),
     safari: new RemoteBrowserTarget('safari', {
       viewport: '1024x768',
+      preRenderDelay: 500,
```
brandonlenz: @Munan059 can you please link to the documentation or [source code](https://github.com/happo/happo/blob/main/src/config/RemoteBrowserTarget.ts) for this setting?

Munan059:
Sorry — last time's "preRenderDelay" was an invalid option, now removed.

This fix: the Card stories use base64 JPEGs, and Happo's image wait fires before Safari finishes decoding, causing flaky diffs (#3509). Switched to "beforeScreenshot" + "HTMLImageElement.decode()" to wait for decoding before the snapshot. See "Card.stories.tsx".

References:
- RemoteBrowserTarget.ts (only viewport/chunks/maxHeight are used): https://github.com/happo/happo/blob/main/src/config/RemoteBrowserTarget.ts
- beforeScreenshot hook: https://docs.happo.io/docs/storybook
- HTMLImageElement.decode(): https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/decode

PR #3574, opened 2026-08-10
fix: treat exact CharacterCount limit as valid

### LINKED ISSUE #3573: [fix] CharacterCount marks an exact-length initial value as invalid

ReactUSWDS 12.0.0; USWDS 3.13.0.

**Describe the bug**

`CharacterCount` initializes its validity with `initialCount < maxLength`, but subsequent changes use `length <= maxLength`. An initial `value` or `defaultValue` whose character count equals `maxLength` therefore receives error styling and becomes natively invalid on blur, even though the same value is valid after an input event.

**To Reproduce**

1. Render `<CharacterCount id="message" name="message" maxLength={5} defaultValue="abcde" />`.
2. Observe that the input and status message receive error classes.
3. Blur the input and observe that native validity reports an error.
4. Trigger an input event with the same five-character value and observe that the error clears.

**Expected behavior**

A value is over the limit only when its character count is greater than `maxLength`. A value exactly equal to the limit should be valid on initial render and after subsequent input. This matches the installed USWDS 3.13.0 predicate.

**Additional context**

PR #3492 also removes this inconsistency as part of a breaking CharacterCount redesign planned for v13. This issue tracks the isolated, non-breaking v12 correction.

rswerve:
# Summary

- Treat an initial CharacterCount value exactly at `maxLength` as valid, matching the component's post-edit behavior and USWDS.
- Add regression coverage for initial values at and over the limit.
- Add an `AtCharacterLimit` Storybook state.

This is a non-breaking correction. The CharacterCount API and structural redesign in #3492 remains deferred to the v13 breaking-change work.

## Related Issues or PRs

Closes #3573.

Related future breaking work: #3492.

## How To Test

1. Run `npm test -- src/components/forms/CharacterCount/CharacterCount.test.tsx`.
2. Open the CharacterCount / At Character Limit story.
3. Confirm the five-character initial value has no invalid styling at a five-character limit.

The full local gate also passed: coverage, server-side rendering and bundle tests, lint, format check, production dependency audit, and the Storybook production build.

## Screenshots (optional)

Not included; the new Storybook state exposes the corrected boundary for review. Happo reports five added snapshots and 1,554 unchanged snapshots. The comparison requires snapshot acceptance before merge but is not treated as a code defect.

  - [x] No

### DIFF
--- a/src/components/forms/CharacterCount/CharacterCount.stories.tsx
+++ b/src/components/forms/CharacterCount/CharacterCount.stories.tsx
@@ -42,10 +42,16 @@ export const TextInput: Story = {
           name="with-hint-input"
           aria-describedby="with-hint-input-info with-hint-input-hint"
           maxLength={args.maxLength}
+          defaultValue={args.defaultValue}
         />
       </FormGroup>
     </Form>
   ),
+}
+
+export const AtCharacterLimit: Story = {
+  ...TextInput,
+  args: { maxLength: 5, defaultValue: 'abcde' },
 }
 
 export const Textarea: Story = {
--- a/src/components/forms/CharacterCount/CharacterCount.test.tsx
+++ b/src/components/forms/CharacterCount/CharacterCount.test.tsx
@@ -248,6 +248,46 @@ describe('CharacterCount component', () => {
       expect(getByText('20 characters allowed')).toBeInTheDocument()
     })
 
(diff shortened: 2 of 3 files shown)


brandonlenz (approved):
I don't think the storybook stories are necessary, but we also don't strictly mirror USWDS examples for our stories, so not blocking.

Issue #3585, opened 2026-08-10
Triage exhaustive-deps findings and enable the lint rule


rswerve:
## Problem

An audit found 23 `react-hooks/exhaustive-deps` findings across 12 shipped source files. Enabling the rule at warning severity immediately would still break the repository's pre-commit hook, which runs ESLint with `--max-warnings=0`, even when a contributor did not introduce the warning.

## Proposed change

- Triage each finding against the component's intended behavior.
- Fix verified dependency defects with focused regression tests.
- Suppress intentional one-time effects with a source comment that explains the invariant.
- Enable `react-hooks/exhaustive-deps` as an error only after source is clean.
- Extend the ESLint configuration regression test to prove missing dependencies fail lint.

## Acceptance criteria

- Existing source reports no `react-hooks/exhaustive-deps` findings.
- Missing hook dependencies fail lint and the pre-commit hook.
- Behavioral fixes include regression tests.
- The repository's full local gate passes.

This is expected to be non-breaking, but each behavioral correction must be evaluated independently.

Issue #3584, opened 2026-08-10
Enable React Hooks lint guardrails


rswerve:
## Problem

`eslint-plugin-react-hooks` is registered, but `react-hooks/rules-of-hooks` is not enabled. An audit found no violations in shipped component source and 31 false positives in Storybook CSF `render` functions.

## Proposed change

- Enable `react-hooks/rules-of-hooks` as an error for source code.
- Exempt `*.stories.tsx` files because Storybook's lowercase CSF `render` callback is a component boundary that the rule cannot infer.
- Add an ESLint API regression test that proves the source rule severity and story exemption through the live flat config.

## Acceptance criteria

- Hook rule violations in ordinary source fail lint.
- Existing Storybook CSF render functions do not fail lint solely because their callback is named `render`.
- The repository's full local gate passes.

This is a tooling-only, non-breaking change.

PR #3588, opened 2026-08-10
chore: enforce React hook rules

rswerve:
## Summary

- enable `react-hooks/rules-of-hooks` as an ESLint error for shipped source
- exempt Storybook CSF render callbacks, whose lowercase `render` name produces false positives
- add a regression test that exercises both paths through the repository's real ESLint configuration
- leave `react-hooks/exhaustive-deps` for #3585, where its 23 existing findings can be triaged before the rule blocks commits

## Related Issues

https://github.com/trussworks/react-uswds/issues/3584

Follow-up: https://github.com/trussworks/react-uswds/issues/3585

## How To Test

Run:

```sh
npm run test
npm run test:coverage
npm run lint
npm run format:check
npm run build
npm run test:serverside
npm run build-storybook
npm audit
```

The ESLint configuration test must report a source hook-order violation as an error while allowing the same hook shape inside a Storybook `render` callback.

## Screenshots

Not applicable.

## Breaking Changes

- [x] This change is not breaking.

### DIFF
--- a/eslint.config.cjs
+++ b/eslint.config.cjs
@@ -76,6 +76,7 @@ module.exports = defineConfig([
     rules: {
       'no-only-tests/no-only-tests': 'error',
       'react/prop-types': 'off',
+      'react-hooks/rules-of-hooks': 'error',
       '@typescript-eslint/no-unused-expressions': [
         'error',
         { allowShortCircuit: true },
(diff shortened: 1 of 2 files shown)


brandonlenz (approved):
Happy to have this rule enforced. 

I think the test is a bit overkill and could be removed, though its a nice verification that the flag does trip and is properly configured. I'd be satisfied with calling that a good ephemeral smoke test and removing it since we don't have tests for other linting rules, etc.

brandonlenz:
while we're at it:

@all-contributors please add @rswerve for bug, maintenance

PR #3595, opened 2026-08-11
fix: keep FilePreview synchronized

rswerve:
# Summary

- Re-read a replacement file when FilePreview receives a new File with the same name.
- Restore the loading and generic-preview state while the replacement is read.
- Create FileReader instances inside the effect and abort in-flight reads during cleanup, so render remains free of browser API construction.
- Add regression coverage for replacement previews, loading state, cleanup, and StrictMode.

The isolated review identified two missing assertions on the original diff. The amended tests cover both; the amended tree passed the full local gate and an additional targeted review.

## Related Issues or PRs

https://github.com/trussworks/react-uswds/issues/3585

## How To Test

1. Run `npx vitest --run src/components/forms/FileInput/FilePreview.test.tsx`.
2. Select a file, then replace it with a different file that has the same name; confirm the preview enters its loading state and then displays the replacement file.
3. Run `npm test`, `npm run lint`, and `npm run format:check`.

## Screenshots (optional)

Not applicable; the visible change occurs only while a replacement file is loading.

  - [x] No

### DIFF
--- a/src/components/forms/FileInput/FileInput.stories.tsx
+++ b/src/components/forms/FileInput/FileInput.stories.tsx
@@ -19,6 +19,8 @@ const meta: Meta<typeof FileInput> = {
         component: `
 ### USWDS 3.0 FileInput component
 Source: https://designsystem.digital.gov/components/file-input
+
+File previews stay synchronized when a selected file is replaced by another file with the same name.
 `,
       },
     },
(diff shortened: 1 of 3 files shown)


src/components/forms/FileInput/FileInput.stories.tsx
```diff
@@ -19,6 +19,8 @@ const meta: Meta<typeof FileInput> = {
         component: `
 ### USWDS 3.0 FileInput component
 Source: https://designsystem.digital.gov/components/file-input
+
+File previews stay synchronized when a selected file is replaced by another file with the same name.
```
brandonlenz, line 23: We don't enumerate every functional element of a component in the stories description, but I've noticed that agents really like to put fluff here. 

I'd revert this change, not necessary.

PR #3634, opened 2026-09-15
feat!: upgrade to USWDS 3.14.0

mdmower-csnw:
# Summary

Update `@uswds/uswds` to [3.14](https://github.com/uswds/uswds/releases/tag/v3.14.0) and make changes to maintain parity. Each table row from USWDS release notes is covered below.

### Features
- [x] `usa-accordion`, `uswds-core` (uswds/uswds#6789):
  - Added optional `iconPosition` prop to `Accordion` which applies `usa-accordion--icon-start` or `usa-accordion--icon-end` when set.
  - Update `Accordion` storybook entries to default the first accordion item to expanded, matching USWDS storybook.
  - Add "Icon Start" and "Icon End" storybook entries to demonstrate effect of `iconPosition` prop.
  - Add/update tests.
- [x] `usa-breadcrumb` (uswds/uswds#6722):
  - Add `truncate` option to `variant` prop. Wrap is now the default.
  - No longer apply inert class `usa-breadcrumb--wrap`; only apply `usa-breadcrumb--truncate` when variant is `truncate`.
  - Update `BreadcrumbBar` storybook entries to use longer `current` breadcrumb, matching USWDS storybook.
  - Replace "Wrapping Breadcrumb" storybook entry with "Truncated Breadcrumb".
  - Add/update tests.
- [x] `usa-range` (uswds/uswds#6673, uswds/uswds#6811):
  - Add optional `hint` prop to `RangeInput`. By default, hint "Move the slider to change the value" appears above the input.
  - Change label in storybook entries to "Range slider", matching USWDS.
  - Add "No Hint" storybook entry.
  - Add "Disabled" and "Aria Disabled" storybook entries for parity with USWDS.
  - Add/update tests.
- [x] `usa-range` (uswds/uswds#6659):
  - No changes required, CSS only.
- [x] `usa-date-picker` (uswds/uswds#6593):
  - Set `aria-current` to `"date"` for today's date.
  - Add/update tests.
- [x] `usa-file-input` (uswds/uswds#6669):
  - No changes required, CSS only.

### Bug fixes
- [x] `usa-modal`, `uswds-core` (uswds/uswds#6703, uswds/uswds#6786):
  - When modal opens and initial focus is determined...
    1. continue respecting the element with `[data-focus]` if it exists, then fall back to...
    2. the first non-disabled button in the footer if it exists, then fall back to...
    3. any non-disabled button in the modal if it exists, and finally fall back to...
    4. the modal itself.
  - Update type annotations, fixing incorrect type assertion on the queried elements.
  - Apply `aria-hidden`/`aria-modal` attributes to modal.
  - Update tests.
- [x] `usa-memorable-date` (uswds/uswds#6725):
  - Support adding screen-reader hints to date inputs with prop `srHint` on `DateInput`
  - Update "Date Of Birth Example" storybook to more closely mimic USWDS.
  - Add tests.
- [x] `usa-file-input` (uswds/uswds#6660):
   - Remove IE11 and legacy Edge user agent tests as a basis for drag text hints. These browsers are not supported by USWDS 3.
    - Switch from "hide drag text" to "show drag text" state based on whether `window.matchMedia()` matches `(hover: hover) and (pointer: fine)`. React to media match changes.
    - Support setting "choose text" for either the fine or not-fine cases.
    - Use "Custom Text" storybook entry to demonstrate use of `chooseTextWithoutDrag`.
    - Replace IE11/Edge choose text tests with new "pointer capability" tests.
- [x] `usa-character-count` (uswds/uswds#6595):
  - This change had related changes that are missing from the v3.14 release notes: uswds/uswds#6767 and uswds/uswds#6664 .
  - Skip announcement on initial render.
  - When character limit is exceeded, prefix announcement with "Character limit exceeded." The new mixed punctuation in strings like `Character limit exceeded. 1 character over limit` matches USWDS.
  - Delay announcements by 1200ms when validity is unchanged or when the count becomes invalid, avoiding unnecessary announcements if subsequent changes are made quick enough to cancel earlier ones. Quickly update announcement status when character count becomes valid.
  - Use 'assertive' announcements when invalid and 'polite' when valid.
  - Update tests.
- [x] `usa-character-count` (uswds/uswds#6385):
  - Does not affect this project.
- [x] `usa-footer` (uswds/uswds#6674):
  - Does not affect this project.
- [x] `usa-banner`, `uswds-core` (uswds/uswds#6714)
  - While this technically does not affect this project, its premise does. Several components are broken in shadow DOMs and light DOMs. A separate issue should be created to improve compatibility.
- [x] `usa-core` (uswds/uswds#6713):
  - Does not affect this project. Focus trap configured with Escape support.
- [x] `usa-core` (uswds/uswds#6594):
  - Does not affect this project.
- [x] `usa-table` (uswds/uswds#6661):
  - No changes required, CSS only.
- [x] `usa-table` (uswds/uswds#6633):
  - No changes required, CSS only.
- [x] `usa-time-picker` (uswds/uswds#6634):
  - No changes required, CSS only.
- [x] `usa-input`, `usa-textarea`, `usa-range`, `usa-combo-box`, `usa-input-prefix-suffix`, `usa-select` (uswds/uswds#6736):
  - No changes required, CSS only.
- [x] `usa-in-page-navigation` (uswds/uswds#6688):
  - Does not affect this project.
- [x] `uswds-core` (uswds/uswds#6651):
  - No changes required, CSS only.

### Guidance changes
- [x] Alert
  - Direct users to review USWDS 3.14 release notes.

### Markup changes
- [x] Memorable date (`DateInput` in this project)
  - This mostly only affects the "Date Of Birth" storybook example. No real breaking changes need to be described, but new support for screen reader only hints could be mentioned.
- [x] File input
  - Differences in drag text based on whether hover and fine pointer are available are discussed in breaking changes.
- [x] Accordion icon alignment
  - Breaking changes mention the changed default icon position new theme setting `$theme-accordion-icon-position`.

## Breaking changes

- **USWDS 3.14.0 required.** Peer dependency is now `@uswds/uswds ^3.14.0`. USWDS 3.14 also changes the appearance of the accordion, breadcrumb, range slider, borderless table row headers, and file input error border. See the [USWDS 3.14.0 release notes](https://github.com/uswds/uswds/releases/tag/v3.14.0).
- **Accordion:** USWDS now places the expand/collapse icon at the start (left) by default. Use the new `iconPosition="end"` prop, or set `$theme-accordion-icon-position: "end"` in your USWDS theme, to keep the previous placement.
- **BreadcrumbBar:** Breadcrumbs wrap by default instead of truncating. `variant="wrap"` no longer has an effect and `usa-breadcrumb--wrap` is no longer applied. Use `variant="truncate"` to restore single-line truncation.
- **RangeSlider:** Component `RangeInput` has been renamed to `RangeSlider`, matching USWDS naming. A new `hint` prop is available to render a usage hint above the slider; it is referenced by `aria-describedby`. Recommended text from USWDS: "Move the slider to change the value".
- **FileInput:** On devices without hover and a fine pointer, the drag instruction is omitted and the text reads "Choose from folder". Customize it with the new `chooseTextWithoutDrag` prop. IE11 and legacy Edge user agent detection was removed. Tests asserting the exact instruction text may need updating.
- **CharacterCount:** The screen reader status is empty on initial render and updates after 1200ms (was 1000ms). Over-limit announcements are prefixed with "Character limit exceeded." and use `aria-live="assertive"`.
- **Modal:** Initial focus goes to the `[data-focus]` element, then the first enabled button in `ModalFooter`, then the first enabled button in the modal, then the modal itself. The dialog has `aria-hidden="true"` while closed and `aria-modal="true"` while open. Tests that query the dialog role while it is closed need `hidden: true`.

## Related Issues or PRs

- #3613

## How To Test

Confirm components behave the same as described in [USWDS release notes for version 3.14.0](https://github.com/uswds/uswds/releases/tag/v3.14.0).

## Happo differences

- **Components/Accordion**
  - This changeset sets the first accordion item to expanded, matching USWDS storybook.
  - The new default expansion icon is on the left now (configurable by theme setting: `$theme-accordion-icon-position`).
  - "Icon Start" and "Icon End" entries are new.
- **Components/Date input**
  - Hint for "Date Of Birth" updated to match USWDS storybook entry "Memorable Date".
- **Components/Modal**
  - Changes are a result of new auto-focus behavior for modals.
- **Components/File input**
  - Changes are a result of new "choose file" messaging based on whether fine pointer exists.
- **Components/Breadcrumb**
  - Current breadcrumb title matches new string from USWDS.
  - Default behavior is to wrap; entry "Wrapping Breadcrumb" became "Truncated Breadcrumb" with opposite behavior.
- **Components/Range slider**
  - "Disabled" and "Aria Disabled" entries added for parity with USWDS storybook.
  - Default hint was added in this changeset. New entry "No Hint" shows slider without hint.

  - [ ] No


src/components/forms/CharacterCount/CharacterCount.tsx
brandonlenz, line 1: Noting this is going to cause https://github.com/trussworks/react-uswds/pull/3492 to get further stale. I'm fine with that as its more important to maintain parity than to introduce breaking refactors. 

Hopefully we can get both in at some point, because I do think a more flexible version of this component would be helpful to have

src/components/forms/RangeSlider/RangeSlider.stories.tsx
```diff
@@ -20,9 +20,7 @@ Source: https://designsystem.digital.gov/components/range-slider
   args: { id: 'range-slider', name: 'range' },
   render: (args) => (
     <>
-      <Label htmlFor="range-slider" hint=" (drag to adjust or use arrow keys)">
-        Slider input
-      </Label>
+      <Label htmlFor="range-slider">Range slider</Label>
```
brandonlenz, line 23: Want to take this a step further and go ahead and rename the component itself from RangeInput to RangeSlider?

Also would make a fine follow-up PR. DateInput needs the same treatment (MemorableDate), which I wouldn't expect of this PR since nothing else about that one's storybook naming changed.

src/components/forms/RangeInput/RangeInput.tsx
```diff
@@ -9,6 +9,8 @@ export type RangeInputProps = {
   max?: number
   textPreposition?: string
   textUnit?: string
+  /** Set to `null` to remove default hint */
+  hint?: React.ReactNode
```
brandonlenz: I'll be thinking about this a bit. the workaround with `null` feels a _bit_ out of place.

I lean more towards removing the default and explicitly passing it through in our examples. I think that would better fit with how the rest of the library works, in addition to leaning in to encouraging consumers to pass in their own content (which many do for i18n purposes)

src/components/Accordion/Accordion.tsx
```diff
@@ -86,6 +87,7 @@ export const Accordion = ({
   items,
   className,
   multiselectable = false,
+  iconPosition,
```
brandonlenz, line 90: I was going to write

> `iconPosition` is optional, but doesn't have a default value meaning most implementations will not properly set either class until they manually add the property.
>
> IMO, the property should either be required, or default to "start"

But looking at the [USWDS implementation](https://designsystem.digital.gov/components/accordion/), this approach matches. Without the class, icon is properly positioned to the left by default. Setting `start` doesn't meaningfully change that (which is what feels off), just makes the choice explicit.

So that said, I think this is the correct way to make this change after all 👍🏻

brandonlenz (commented):
Thanks for this thorough set of changes to bring us to parity with USWDS 3.14.0!

Really only the "set the hint to null for a specific side effect" thing needs some attention here, IMO. Otherwise, this looks great.  remarked on a few couple things that don't necessarily need to be something done in this PR.

src/components/forms/RangeSlider/RangeSlider.stories.tsx
mdmower-csnw: Sure, done in [baa994f](https://github.com/trussworks/react-uswds/pull/3634/commits/baa994f35a152e50a6b2ee68bd025bfb4ef0b0f2). I opted to leave `RangeInput` as a passthrough with a JSDoc `@deprecated` tag. I also considered using `withDeprecationWarning`, but that function doesn't appear to be working (and there's a good chance I'm responsible: #3315). Let me know if you'd prefer the `RangeInput` export be removed since this changeset will require a major version change anyways.

src/components/forms/RangeInput/RangeInput.tsx
mdmower-csnw: I agree with you. I was trying to mirror USWDS, but this definitely felt backwards. I removed the default value in 
7ce13a3 and added a JSDoc with recommended text.

src/components/Accordion/Accordion.tsx
mdmower-csnw: > Setting `start` doesn't meaningfully change that

In fact, it could meaningfully change things for the worse if a user sets theme variable `$theme-accordion-icon-position: "end"` and does not set `iconPosition`. The component default would override the theme setting.

